PR #3, opened 2022-04-09
Bug 574363 - Move jdt.junit* bundles to Java 11

iloveeclipse:
Originally created as
https://git.eclipse.org/r/c/jdt/eclipse.jdt.ui/+/182392


iloveeclipse:
???
```
SNAPSHOT:compile (default-compile) on project org.eclipse.jdt.junit.runtime: Compilation failure: Compilation failure: 
11:48:22  [ERROR] /home/jenkins/agent/workspace/eclipse.jdt.ui-github_PR-3/org.eclipse.jdt.junit.runtime/src/org/eclipse/jdt/internal/junit/runner/ITestReference.java: 
11:48:22  [ERROR] 	/*******************************************************************************
11:48:22  [ERROR] 	^
11:48:22  [ERROR] Type java.lang.Object is indirectly referenced from required .class files but cannot be resolved since the declaring package java.lang exported from module java.base conflicts with a package accessible from module <unnamed>
11:48:22  [ERROR] 1 problem (1 error)
```

org.eclipse.jdt.junit5.runtime/build.properties
```diff
@@ -18,3 +18,4 @@ bin.includes = about.html,\
                plugin.properties,\
                .,\
                META-INF/
+jre.compilation.profile = JavaSE-1.8
```
vogella, line 21: Why change BREE but add a 1.8 compilation profile? If that is intended, please document via commit message.

ktatavarthi:
@iloveeclipse I donot have much idea regarding `jre.compilation.profile = JavaSE-1.8` , but as there is no pom change and bree changes to Java 11, will Maven-Tycho build use the above attribute to use --release option?

org.eclipse.jdt.junit5.runtime/build.properties
iloveeclipse: > Why change BREE but add a 1.8 compilation profile? If that is intended, please document via commit message.

See https://bugs.eclipse.org/bugs/show_bug.cgi?id=574363

vogella: Please document via the commit message.

ktatavarthi:
> > will Maven-Tycho build use the above attribute to use --release option?
> 
> I have no idea what Tycho/Maven will do. If no one knows if compiling on java 11 with target 8 is possible, we can either drop this idea entirely and keep the bundle on 8 or move to 11 but that would mean 8 will be not supported for test.

I have experimented locally and the below code added to the pom files of the three junit runtime plugins makes sure that the tycho maven build uses the --release option 

```
  <build>
	<plugins>
		<plugin>
		 	<groupId>org.eclipse.tycho</groupId>
			<artifactId>tycho-compiler-plugin</artifactId>
			<configuration>
				<release>8</release>
			</configuration>
		</plugin>
	</plugins>
  </build>
```

ktatavarthi:
> > > will Maven-Tycho build use the above attribute to use --release option?
> > 
> > 
> > I have no idea what Tycho/Maven will do. If no one knows if compiling on java 11 with target 8 is possible, we can either drop this idea entirely and keep the bundle on 8 or move to 11 but that would mean 8 will be not supported for test.
> 
> I have experimented locally and the below code added to the pom files of the three junit runtime plugins makes sure that the tycho maven build uses the --release option
> 
> ```
>   <build>
> 	<plugins>
> 		<plugin>
> 		 	<groupId>org.eclipse.tycho</groupId>
> 			<artifactId>tycho-compiler-plugin</artifactId>
> 			<configuration>
> 				<release>8</release>
> 			</configuration>
> 		</plugin>
> 	</plugins>
>   </build>
> ```

The above change would be in addition to all the changes that are already done in this PR.

iloveeclipse:
> The above change would be in addition to all the changes that are already done in this PR.

Cool, thanks, I've rebased now.

jjohnstn (approved):
LGTM

iloveeclipse:
Thanks Jeff.

PR #13, opened 2022-04-13
Fixes #12 : Bug 577317 - [17] Provide quick fixes to add permitted

### LINKED ISSUE #12: Bug 577317 - [17] Provide quick fixes to add permitted class/interface for sealed class/interface
Adding the description from https://bugs.eclipse.org/bugs/show_bug.cgi?id=577317#c0 by @SarikaSinha 

I create a "sealed" class C1 without any permits.
This shows compilation error "Sealed class or interface lacks the permits clause and no class or interface from the same compilation unit declares test1 as its direct superclass or superinterface"
But I don't see any corresponding quick fix, it will be good to see 2 quickfixes. One to create class in the same compilation unit and other one to create a child class in the same package.

ktatavarthi:
class/interface for sealed class/interface

Signed-off-by: Kalyan Prasad Tatavarthi <kalyan_prasad@in.ibm.com>


SarikaSinha:
image

If I use quick fix to add class in same package, the "permits" is not coloured properly

ktatavarthi:
> image
> 
> If I use quick fix to add class in same package, the "permits" is not coloured properly. After going back to the file and adding a "enter" applies the colouring.

I am looking into this.

ktatavarthi:
> For interface, we have more challenges. We already had 2 quickfixes to create implementation class and to create new interface. These 2 should not appear if sealed is used as this again leads to errors in both the files.
> 
> Also, we can explore keeping these quickfixes for class and interface even without errors to add new permits. We can discuss that in a separate issue.

As currently there are no quick fixes available, I think these quick fixes are valid. There is no way to avoid error using sealed in the class creation wizard as there is no easy solution for this other than quick fixes. If we restrict creation of permits class or interface with final and non-sealed modifiers and not allow sealed modifier, then we restrict our Quick Fix to not create sub types which can also be sealed.

SarikaSinha:
Created https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/25 as follow up.

PR #16, opened 2022-04-17
remove unused import

carstenartur:
Unfortunately eclipse is unable to detect unused import if the import
references the class itself. The "organize imports" is able to find it
thought.


jjohnstn:
Looks good, merging.

PR #19, opened 2022-04-19
Bug 578871 - use ISchedulableOperation

jukzi:
to prevent UI freeze during UNDO

needs
https://github.com/eclipse-platform/eclipse.platform.runtime/pull/30

### DIFF
--- a/org.eclipse.ltk.core.refactoring/src/org/eclipse/ltk/internal/core/refactoring/UndoableOperation2ChangeAdapter.java
+++ b/org.eclipse.ltk.core.refactoring/src/org/eclipse/ltk/internal/core/refactoring/UndoableOperation2ChangeAdapter.java
@@ -29,7 +29,10 @@ import org.eclipse.core.runtime.IStatus;
 import org.eclipse.core.runtime.NullProgressMonitor;
 import org.eclipse.core.runtime.Status;
 import org.eclipse.core.runtime.SubProgressMonitor;
+import org.eclipse.core.runtime.jobs.ISchedulableOperation;
+import org.eclipse.core.runtime.jobs.ISchedulingRule;
 
+import org.eclipse.core.resources.IWorkspace;
 import org.eclipse.core.resources.IWorkspaceRunnable;
 import org.eclipse.core.resources.ResourcesPlugin;
 
@@ -40,7 +43,7 @@ import org.eclipse.ltk.core.refactoring.RefactoringCore;
 import org.eclipse.ltk.core.refactoring.RefactoringStatus;
 
 
-public class UndoableOperation2ChangeAdapter implements IUndoableOperation, IAdvancedUndoableOperation  {
+public class UndoableOperation2ChangeAdapter implements IUndoableOperation, IAdvancedUndoableOperation, ISchedulableOperation  {
 
 	private String fLabel;
 	private String fDescription;
@@ -348,7 +351,7 @@ public class UndoableOperation2ChangeAdapter implements IUndoableOperation, IAdv
 				monitor.done();
 			}
 		};
-		ResourcesPlugin.getWorkspace().run(runnable, pm);
+		ResourcesPlugin.getWorkspace().run(runnable, getSchedulingRule(), IWorkspace.AVOID_UPDATE, pm);
 		return result;
 	}
 
@@ -389,5 +392,10 @@ public class UndoableOperation2ChangeAdapter implements IUndoableOperation, IAdv
 		}
 		fActiveChange.dispose();
 		fActiveChange= null;
+	}
+
+	@Override
+	public ISchedulingRule getSchedulingRule() {
+		return ResourcesPlugin.getWorkspace().getRoot();
 	}
 }
--- a/org.eclipse.ltk.core.refactoring/META-INF/MANIFEST.MF
+++ b/org.eclipse.ltk.core.refactoring/META-INF/MANIFEST.MF
@@ -3,7 +3,7 @@ Automatic-Module-Name: org.eclipse.ltk.core.refactoring
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.ltk.core.refactoring; singleton:=true
-Bundle-Version: 3.12.100.qualifier
+Bundle-Version: 3.12.200.qualifier
 Bundle-Activator: org.eclipse.ltk.internal.core.refactoring.RefactoringCorePlugin
 Bundle-ActivationPolicy: lazy
 Bundle-Vendor: %providerName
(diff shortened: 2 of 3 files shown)


jukzi:
please rebuild.

org.eclipse.ltk.core.refactoring/src/org/eclipse/ltk/internal/core/refactoring/UndoableOperation2ChangeAdapter.java
```diff
@@ -40,7 +43,7 @@
 import org.eclipse.ltk.core.refactoring.RefactoringStatus;
 
 
-public class UndoableOperation2ChangeAdapter implements IUndoableOperation, IAdvancedUndoableOperation  {
+public class UndoableOperation2ChangeAdapter implements IUndoableOperation, IAdvancedUndoableOperation, ISchedulableOperation  {
```
gayanper, line 46: I would like to know why we really needed the `ISchedulableOperation` interface here ?

jukzi: please see Bug 578871. Otherwise it will freeze the UI - as long as any other Thread (for example autobuild) has any lock in the workspace.

gayanper: Ok thanks @jukzi i didn't refer to the platform ui PR which is still open. The change make sense then. But shall we wait till we have the ui changes as well ? Otherwise the change LGTM

jukzi: can you please trigger rebuild?

org.eclipse.ltk.core.refactoring/src/org/eclipse/ltk/internal/core/refactoring/UndoableOperation2ChangeAdapter.java
```diff
@@ -348,7 +351,7 @@ private ExecuteResult executeChange(final IValidationCheckResultQuery query, IPr
 				monitor.done();
 			}
 		};
-		ResourcesPlugin.getWorkspace().run(runnable, pm);
+		ResourcesPlugin.getWorkspace().run(runnable, getSchedulingRule(), IWorkspace.AVOID_UPDATE, pm);
```
iloveeclipse, line 354: It would be good to update commit message and explain why the change is needed, how this works and why this is safe to use extra lock in undo adapter.

I personally haven't yet tested this, but it would not surprise me if that would lead to similar deadlocks like the one seen in https://github.com/eclipse-platform/eclipse.platform.ui/issues/15

If any of the custom operations that are wrapped by this adapter do some non trivial locking, that might be dangerous. I also thinking about non-workspace rules that might be used - couldn't they lead to errors (rules cannot be nested?). 

So this need to be evaluated and if the change is safe, explained in commit why.

iloveeclipse (changes requested):

org.eclipse.ltk.core.refactoring/src/org/eclipse/ltk/internal/core/refactoring/UndoableOperation2ChangeAdapter.java
iloveeclipse: Hint: updating commit message will automatically trigger rebuild.

org.eclipse.ltk.core.refactoring/src/org/eclipse/ltk/internal/core/refactoring/UndoableOperation2ChangeAdapter.java
jukzi: is there a way to update commit message within github? i dont find the button

iloveeclipse: It is called "amend commit" and "force push" from Eclipse

jukzi:
I think I covered all requested changes but the review is still -1, what do you think is missing?

org.eclipse.ltk.core.refactoring/src/org/eclipse/ltk/internal/core/refactoring/UndoableOperation2ChangeAdapter.java
iloveeclipse: > So this need to be evaluated and if the change is safe, explained in commit why.

OK, new commit message is better but still doesn't explain **why** it is safe, just says it **is** safe.

The explanation that would simplify life of the reviewers could be:

The original call to `ResourcesPlugin.getWorkspace().run(runnable, pm);` is equivalent to the call of   `Workspace.run(IWorkspaceRunnable, IProgressMonitor)`, which is implemented internally as `run((ICoreRunnable) action, defaultRoot, IWorkspace.AVOID_UPDATE, monitor);` - means, changing the original code to proposed one is safe and doesn't change any behavior because the executed code it is identical at the end.

iloveeclipse (changes requested):
Jörg: I see there are two commits now. 

Ideally that should be one (no need here for multiple commits), with *appropriate* commit message. I fear, github own "squash and merge" will generate some stupid commit message, so ideally you could locally squash your two commits into one (using proper commit message) & push again.

jukzi:
I have no influence on what you will put in the commit message.:
When you press squash then you get an msg editor that you have to edit. The message header will be the first line in the commit message and the multilinetext below ist line 3+ in the commit message. It defaults to a combination of all commit messages plus some additional lines if any. Feel free to change it to anything you understand.

jarthana:
> I have no influence on what you will put in the commit message.

That's what we (I believe @iloveeclipse too) are trying to avoid, I suppose. Besides, I think it also makes it easier to review if the changes are in one commit, although I have no personal preference.

iloveeclipse:
@jukzi : I've tried to undo method rename wit all patches applied, but none operations I get in `org.eclipse.ui.operations.OperationHistoryActionHandler.run()` are UndoableOperation2ChangeAdapter.

Can you provide **exact** steps to reproduce the original issue described in https://bugs.eclipse.org/bugs/show_bug.cgi?id=578871?

jukzi:
> @jukzi : I've tried to undo method rename

In eclipse ide in a java editor select a java method name, press ALT+SHIFT+R to rename, edit name, press ENTER. 
Undo with CTRL+Z

iloveeclipse:
> In eclipse ide in a java editor select a java method name, press ALT+SHIFT+R to rename, edit name, press ENTER.
Undo with CTRL+Z

I'm doing exact this and still get no UndoableOperation2ChangeAdapter

Debugging that I see this https://github.com/eclipse-platform/eclipse.platform.ui/pull/18 was not complete, see my debugger screenshot that explains why:
image

jukzi:
> Debugging that I see this [eclipse-platform/eclipse.platform.ui#18](https://github.com/eclipse-platform/eclipse.platform.ui/pull/18) was not complete,

Then we can let TriggeredOperations forward the ISchedulableOperation.

iloveeclipse:
> Then we can let TriggeredOperations forward the ISchedulableOperation.

Would you provide a PR?


Reminder for myself: draft commit message here (assuming no other changes needed)
```
Bug 578871 - UndoableOperation2ChangeAdapter use ISchedulableOperation

Give the caller of the UndoableOperation2ChangeAdapter.execute a hint
that this operation acquires workspace root as ISchedulingRule.
Undo operation freezes UI if any other thread (for example
autobuild job) holds the workspace lock during undo.

With the given hint UI (org.eclipse.ui.operations.OperationHistoryActionHandler) 
can prevent the freeze in RefactoringExecutionHelper.perform() while waiting 
for the workspace rule.

This change does not introduce any additional locking, since original code
already implicitly acquired workspace lock by using workspace root as a rule.
```

jukzi:
> Would you provide a PR?

https://github.com/eclipse-platform/eclipse.platform.ui/pull/37

iloveeclipse (approved):

PR #20, opened 2022-04-20
Bug 99622 - [rename] Rename method misses ambiguously overridden method

jjohnstn:
- add null check to prevent NPE in TreeSet comparator


gayanper (approved):
LGTM

PR #22, opened 2022-04-20
Add number of changed files to cleanup preview

carstenartur:
Currently you get a long list of changes in the preview and it is not easy to see what impact a change in settings on a rerun has got. This patch adds the number of files (currently in a separate line). This is a poc to see how it looks like. Maybe it would be better to have a more detailed list about the amount of each cleanup in a list.


jjohnstn:
Hi Carsten, I think that the number of files should be incorporated in the title message:

The following changes to 28 files is necessary to perform the refactoring

You need to specify "files" because there can be many more changes and it is not clear from the current version of this patch that is files and not changes.  I also think a user may want to know the number of potential changes more than files.  For example,

The following 63 changes to 28 files is necessary to perform the refactoring.

The message would not sync with user unselecting various clean-ups. Not sure, but a user may want the number of files and changes if filtering is selected (e.g. down to 13 changes to 3 files if user selects "enhanced for loop").

vogella:
Great to see you again @carstenartur I was a bit concerned that something happened to you, I also tried to contact you via email but got no reply. Welcome back!

carstenartur:
> Hi Carsten, I think that the number of files should be incorporated in the title message:
> 
> 
> 
> The following changes to 28 files is necessary to perform the refactoring
> 
> 
> 
> You need to specify "files" because there can be many more changes and it is not clear from the current version of this patch that is files and not changes.  I also think a user may want to know the number of potential changes more than files.  For example,
> 
> 
> 
> The following 63 changes to 28 files is necessary to perform the refactoring.
> 
> 
> 
> The message would not sync with user unselecting various clean-ups. Not sure, but a user may want the number of files and changes if filtering is selected (e.g. down to 13 changes to 3 files if user selects "enhanced for loop").

Yes, I thought similar. But then I wondered if it wouldn't be better to present a small list of sums of amounts of hits for each cleanup choosen. For such a list I need to find out how to compute that from the result of the cleanup run. Maybe I need to change the code processing each cleanup in an earlier stage first to be able to provide the requested information (number of lines changed, number of hits for each cleanup, total number of characters added/removed,...)?

Because of this triggering possibly requirement for further changes I thought first asking is a good idea. Do you think I should at first put the single number of affected files into the sentence like you suggested it and maybe as a second step try to provide more information? I'm not sure that I catched the point in your last sentence with beeing out of sync. However, counting number of hits of a cleanup is sometimes not so useful anyway. The reason is that there are cleanups taking care a changing use of a Variable that might be used n times. But I'm open to try that if I find a way to compute the information.

jjohnstn:
> 
I don't think number of lines is necessary.  I just thought number of changes might be useful but I don't think number of lines is necessary.

Adding the number of files to title as a start is reasonable.  I think the number of changes can be determined from the result set (i.e. number of operations). 

The point about sync is that when you filter, you only see the cleanup you specify, but I would not expect the message to change to limit to the files/changes specific to the filtered cleanup.

org.eclipse.ltk.ui.refactoring/src/org/eclipse/ltk/internal/ui/refactoring/PreviewWizardPage.java
```diff
@@ -561,6 +561,9 @@ protected void setTreeViewerInput() {
 		PreviewNode input= null;
 		if (fTreeViewerInputChange != null) {
 			input= AbstractChangeNode.createNode(null, fTreeViewerInputChange);
+			int filenumber= fTreeViewerInputChange.getFilenumber();
+			String fulldescription= MessageFormat.format(RefactoringUIMessages.PreviewWizardPage_description, String.valueOf(filenumber),filenumber==0?"":"s",filenumber==0?"is":"are"); //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$ //$NON-NLS-4$
```
jjohnstn, line 565: I believe you mean to test against fileNumber == 1 instead of 0.  So that you get "change for 1 file is" vs
"changes for 2 files are"

jjohnstn (changes requested):

org.eclipse.ltk.ui.refactoring/src/org/eclipse/ltk/internal/ui/refactoring/PreviewWizardPage.java
carstenartur: Indeed, but maybe it is a bad idea to solve it this way anyway. It is not possible to translate such sentence in babel, the Eclipse translation project, right?

jjohnstn:
Hi Carsten.  I messed up your PR doing a push --force but I have corrected it after rebasing on master.

org.eclipse.ltk.ui.refactoring/src/org/eclipse/ltk/internal/ui/refactoring/PreviewWizardPage.java
jjohnstn: Hmm, thinking about it, the correct way is to have two messages - one plural and one non-plural.  That way, translation can occur properly.

carstenartur: Changed it to have 
RefactoringUIMessages.PreviewWizardPage_description_s for 1
RefactoringUIMessages.PreviewWizardPage_description_m for multiple files changed

jjohnstn:
Hi Carsten, I couldn't merge this for 4.24 M3.  I didn't know how to get around the freeze period check above and tried asking on our slack channel but no help.  On gerrit, I simply ignore it but there is no option on this page for me to do so.  They set the freeze period for Monday even though 5:00 is the deadline.  I tried rerunning the verification but the freeze period check was too fast.  Another github problem for me.  Anyway, this will have to wait until 4.25 M1.

PR #26, opened 2022-04-24
add integer and long to systemproperty cleanup

carstenartur:
based on pr-24
Hi Jeff, I *think* I can only edit (push) the pull requests when you work on it using the "fetch github pull request" functionality of egit and push to my fork what should be possible for maintainers or use any other supported way for github as I do not have a push permission on original eclipse-jdt/eclipse.jdt.ui .
The "Allow edits and access to secrets by maintainers" checkbox is checked for this pull request.
Regarding the content of this pull request
1) I added another "boxed" change in the tests
2) I did not yet remove the unrelated change regarding stringbuffer and stringbuilder. Maybe you have added it intentionally?
3) I added a second case System.getProperty("foo","false") for boolean. Maybe we can add similar cases for integer and long if the default is exactly what would have read anyway


jjohnstn:
Thanks Carsten.  I changed the stringbuffer and stringbuilder as it was causing a warning due to the link being a typo.  It can be removed if you want and I'll fix it in a separate issue.  The only issue with the patch is that the new parseBoolean support for a second argument fails to remove the NLS comment after the String argument is removed.  We can make this a separate issue if you want or else you can tackle it.

carstenartur:
> Thanks Carsten.  I changed the stringbuffer and stringbuilder as it was causing a warning due to the link being a typo.  It can be removed if you want and I'll fix it in a separate issue.  The only issue with the patch is that the new parseBoolean support for a second argument fails to remove the NLS comment after the String argument is removed.  We can make this a separate issue if you want or else you can tackle it.

I applied further changes to support default values for long and integer but there are now at least three bugs:
1) missing removement of the second nls comment as you mentioned
2) in case of a default value as string literal after cleanup is applied the number is used instead of string literal
3) long value should use <number>l instead of <number> I guess

jjohnstn:
Hi Carsten.  When I try the latest patch set, I am finding that NLS comments are removed for:

		System.out.println(Long.parseLong(System.getProperty("Jeff", "15"))); //$NON-NLS-1$ //$NON-NLS-2$

without any other changes to the line.

but for:

		System.out.println(Boolean.parseBoolean(System.getProperty("Jeff", "false"))); //$NON-NLS-1$ //$NON-NLS-2$

it makes the change and does not remove the NLS.  I am wondering if there is some reverse logic needed in your last commit which I see adds NLS removal.

jjohnstn:
I found a problem with replaceAndRemoveNLS().  You are using the toString() method to get a string for the visited node and for the replacement string.  When you try and find the visited.toString() in the buffer, this fails if there is any white-space as the debug string removes all white-space.  Similarly, the replacement string will have no white-space so that is why parameters are not spaced.  I added space before replacement and it appears fine.  The final issue was to add an UNUSED object to note when a parameter was discarded and we still need to remove NLS.  I updated the tests and repushed.  Once verification is complete, I will merge the patches.

jjohnstn (approved):

jjohnstn:
Hi Carsten, are you willing to add a N&N entry?

carstenartur:
> I found a problem with replaceAndRemoveNLS().  You are using the toString() method to get a string for the visited node and for the replacement string.  When you try and find the visited.toString() in the buffer, this fails if there is any white-space as the debug string removes all white-space.  Similarly, the replacement string will have no white-space so that is why parameters are not spaced.  I added space before replacement and it appears fine.  The final issue was to add an UNUSED object to note when a parameter was discarded and we still need to remove NLS.  I updated the tests and repushed.  Once verification is complete, I will merge the patches.

Great, thanks. Wasn't the formatting of the toString() call using the code fixed regarding indentation in https://git.eclipse.org/r/c/jdt/eclipse.jdt.core/+/181069 ?
Nah, I really don't remember completely. Maybe at one point it should be checked again.

carstenartur:
> Hi Carsten, are you willing to add a N&N entry?

Sure, I can try to do it this Weekend.

carstenartur:
> Hi Carsten, are you willing to add a N&N entry?

https://github.com/eclipse-platform/www.eclipse.org-eclipse-news/pull/19

PR #55, opened 2022-05-16
While loop to foreach loop conversion

carstenartur:
convert while loop to foreach where possible


carstenartur:
Hi @jjohnstn,
if you feel like doing something different you are welcome to comment on this cleanup especially on the @Ignored junit tests. Some corner cases definitely should be supported.

jjohnstn:
Hi Carsten, I'll take a look.  I had started working on the feature, but will try working on your patch as yours is mostly complete.

jjohnstn:
Hi Carsten, I fixed the Iterable check and enabled the test that should not change because it is not a true Iterable.  I'll keep working on fixing the ignored tests until all are enabled.

jjohnstn:
@carstenartur I have added a number of checks and all of your tests are now enabled without ignoring.  If you can think of any other test scenarios, let me know, otherwise, I think this is ready for 4.25 M1.  I don't think the VisitorTest is enabled at the moment.

org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/fix/helper/WhileToForEach.java
```diff
@@ -0,0 +1,275 @@
...
+			Set<CompilationUnitRewriteOperation> operations, Set<ASTNode> nodesprocessed) {
+		ReferenceHolder<ASTNode, WhileLoopToChangeHit> dataholder= new ReferenceHolder<>();
+		Map<ASTNode, WhileLoopToChangeHit> operationsMap= new LinkedHashMap<>();
+		WhileLoopToChangeHit emptyHit= new WhileLoopToChangeHit();
+		HelperVisitor.callVariableDeclarationStatementVisitor(Iterator.class, compilationUnit, dataholder, nodesprocessed, (init_iterator, holder_a) -> {
+			List<Object> computeVarName= computeVarName(init_iterator);
+			if (computeVarName != null) {
+				HelperVisitor.callWhileStatementVisitor(init_iterator.getParent(), dataholder, nodesprocessed, (whilestatement, holder) -> {
+					String name= computeNextVarname(whilestatement);
+					if (computeVarName.get(0).equals(name)) {
```
carstenartur, line 73: Is this the usual way to check we are refering to the right variable? Should we somehow consider multiple variable cases?

org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/fix/helper/WhileToForEach.java
```diff
@@ -0,0 +1,275 @@
...
+		result.setName(name);
+
+		Expression expression= hit.loopvardeclaration.getExpression();
+		SimpleName variable= ASTNodes.as(expression, SimpleName.class);
+		String looptargettype;
+		looptargettype= variable.resolveTypeBinding().getErasure().getQualifiedName();
+		VariableDeclarationStatement typedAncestor= ASTNodes.getTypedAncestor(hit.loopvardeclaration, VariableDeclarationStatement.class);
+		Type type;
+		if (hit.nextwithoutvariabledeclation) {
+//			ITypeBinding o=hit.loopvardeclaration.getExpression().resolveTypeBinding();
```
carstenartur, line 223: Maybe some more cleanup is needed.

org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/fix/helper/WhileToForEach.java
```diff
@@ -0,0 +1,275 @@
...
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Carsten Hammer
+ *******************************************************************************/
```
carstenartur, line 13: The Copyright header is just copied from my sandbox project. Please tell me how to adjust or do it yourself. As minimal solution you should add yourself in all files changed or new.

carstenartur (commented):
Many thanks, Jeff - really impressive work. 
There are some cases to investigate at
https://github.com/carstenartur/eclipse.jdt.ui/commit/cc0fbc3c999cbbdd26c42b0003007c9a497f701a
They can either cherry picked or I can add them to the pull request.
You should tell me if you want me to do any cleanup or if you want to do what is necessary before merging. Both way fine with me.

org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/fix/helper/WhileToForEach.java
jjohnstn: Feel free to push into the PR.  I will start working on the new test scenarios.  I think the complex one with if statements should be ignored.  The iterator() call should be in the same block as the while loop that uses it.

jjohnstn: Hi Carsten, I decided to remove the extra groups.  I was trying to solve a bug that I know believe is in the rewrite code, but I ended up just leaving them out as I think it is not needed and adds complexity.  I added support for checking that the loop var is used and adhering to the sub-preference which is necessary because the end-loop will show a warning/error and the user needs to decide whether it is ok.  I added a couple of new nested tests.  I also added a check where the code just calls the it.next() method without doing anything with it.  This type of call can't be converted to the loop var name without causing an error (e.g. it.next(); -> string;)  I did some formatting changes and converting some variables to camel-case.

FWIW: the failure occurs when I have a nested while and after the inner while, there is only comments.  It could be any while but this is how I found it.  I tried a variety of attempts to counter it (e.g. converting inner loop first), but it appears it is an error somwhere in the rewrite code (probably doesn't copy or move the comment properly).

    void m(List<String> strings,List<String> strings2) {
        Collections.reverse(strings);
        Iterator it = strings.iterator();
        while (it.hasNext()) {
        	it.next();
            Iterator it2 = strings2.iterator();
            while (it2.hasNext()) {
                String s2 = (String) it2.next();
                System.out.println(s2);
            }
            // this extraneous comment causes a range error to occur
        }
    }

jjohnstn:
Hi Carsten, I added a validation step that fixes invalid situations.  There is only one ignored test left but I feel it is an unfair scenario.  A user could also modify the collection that is passed as argument but in another thread.  Another method could take a collection and do modifications and the user could pass the same collection to both, etc... IMO, the change should occur and the user may have to track a ConcurrentModificationException if it happens but it should not be on the cleanup to determine such scenarios.

carstenartur:
> Hi Carsten, I added a validation step that fixes invalid situations. There is only one ignored test left but I feel it is an unfair scenario. A user could also modify the collection that is passed as argument but in another thread. Another method could take a collection and do modifications and the user could pass the same collection to both, etc... IMO, the change should occur and the user may have to track a ConcurrentModificationException if it happens but it should not be on the cleanup to determine such scenarios.

Hi Jeff, 
I really like your work. Still the code does not grow so much though I wonder if we can integrate some boiler plate code into the visitor - but later more to this question.

First regarding the junit tests: tbh, that is exactly the situation I had in mind with different threads. I know it is a little bit unfair but similar code using instance variables for collections you can find all over eclipse causing issues with multithreading. In fact some collections allow this others do not. For the simple for loop to enhanced for loop cleanup we have had cases (still have - discovered by the cleanup!) where the original and still used code is broken but nobody can see it. It is the fail-fast approach of the enhanced for loop that suddenly  discovers these cases throwing concurrentmodificationexceptions. Without that you work on silently changed data opening the door to raceconditions.
It is of course up to you to decide how to go on. I definitely can already do a lot with the cleanup as it is now by considering corner cases myself.
I just would like to know if that is difficult to implement such "follow the instance variable use" and check for accesses that add or remove entries on the collection. As I said for the simple for loop to enhanced for loop cleanup it might be useful too.

Here some issues I see. Maybe I add some more junit tests, but not today.

	// iterate variables according to the priorities defined in FakedTrackingVariable.IteratorForReporting.Stage
	Iterator<FakedTrackingVariable> iterator = new FakedTrackingVariable.IteratorForReporting(this.trackingVariables, this, location != null);
	while (iterator.hasNext()) {
		FakedTrackingVariable trackingVar = iterator.next();

After cleanup

	// iterate variables according to the priorities defined in FakedTrackingVariable.IteratorForReporting.Stage
	for (FakedTrackingVariable trackingVar : this) {
		if (returnVar != null && trackingVar.isResourceBeingReturned(returnVar)) {
			continue;
		}

I think the "this" is an invalid substitution here.
Another testcase that looks wrong, maybe just overlook something 

		                                Iterator<ParameterizedTypeBinding> captureKeys = tmpBoundSet.captures.keySet().iterator();
						Set<ParameterizedTypeBinding> toRemove = new LinkedHashSet<ParameterizedTypeBinding>();
						while (captureKeys.hasNext()) {
							ParameterizedTypeBinding key = captureKeys.next();
							int len = key.arguments.length;
							for (int i = 0; i < len; i++) {
								if (TypeBinding.equalsEquals(key.arguments[i], variable)) {
									toRemove.add(key);
									break;
								}
							}
						}

After cleanup

		                                Set<ParameterizedTypeBinding> toRemove = new LinkedHashSet<ParameterizedTypeBinding>();
						for (ParameterizedTypeBinding key : tmpBoundSet.captures.keySet()) {
							int len = key.arguments.length;
							for (int i = 0; i < len; i++) {
								if (TypeBinding.equalsEquals(key.arguments[i], variable)) {
									toRemove.add(key);
									break;
								}
							}
						}
I think the inner loop should be changed too but maybe our code is not able to do it in one step.

Now back to the point regarding removing boiler plate code and moving it into the visitor. See such code:

				HelperVisitor.callWhileStatementVisitor(init_iterator.getParent(), dataholder, nodesprocessed, (whilestatement, holder) -> {
					String name= computeNextVarname(whilestatement);
					
The BiPredicate called is able to make use of two variables: "whilestatement" which is the found node itself and "holder" as generic space for all kinds of data. The declaration allows to handout a lot of different data:

ReferenceHolder<ASTNode, WhileLoopToChangeHit> dataholder= new ReferenceHolder<>();

I wonder at that point if we should not make one of the two values "WhileStatement"-specific in that way that it e.g. has a whilestatementholder type that can return the generic "holder" and at the same time provides some "WhileStatement"-specific methods e.g a  computeNextVarname() method removing this code from our class completely and offer it to be very easy reusable. Of course there is more code that can be moved to this type. Depending on the node type each specific holder could offer a bunch of methods performing some computation tasks for the node. I do not say that I can tell which methods are useful but I guess it doesn't matter to start with whatever we have and can add more methods as soon as we have more code.
When done right I can imagine that a lot of code can be moved out of the find() method and becomes reusable.
Maybe that all can be done at a later point in time as well but I'm not sure if we do need some preparation right now to not sacrify compatibility.
Of course for "HelperVisitor.callVariableDeclarationStatementVisitor" the BiPredicate would then have a spcific "VariableDeclarationStatementHolder", the "HelperVisitor.callMethodInvocationVisitor" would have a "MethodInvocationHolder" and so on. Eg. you do not have to compute the method name, you call MethodInvocationHolder.getFullyQualifiedName() - something like this.
What do you think?

jjohnstn:
Hi Carsten, I fixed the problem with FakedTrackingVariable.  The cleanup no longer allows indirect setting of iterator either by method or constructor.  Added some tests as well.  I also added support for setting the iterator via assignment as long as it is before the while statement.  Regarding adding some more functionality to holders, it would have to be generic.  Stuff like computeVarName() is specific to this particular cleanup only, unless I am missing something.  Generic stuff could be useful though.

carstenartur:
> Hi Carsten, I fixed the problem with FakedTrackingVariable. The cleanup no longer allows indirect setting of iterator either by method or constructor. Added some tests as well. I also added support for setting the iterator via assignment as long as it is before the while statement. Regarding adding some more functionality to holders, it would have to be generic. Stuff like computeVarName() is specific to this particular cleanup only, unless I am missing something. Generic stuff could be useful though.

Looks great already, feel free to merge when it is fine for you. Maybe additional functionality as eg. providing a way to give you the variables a while is running on can be investigated later as well. A breaking change is possible as long as there is no release, right?

jjohnstn:
> > Hi Carsten, I fixed the problem with FakedTrackingVariable. The cleanup no longer allows indirect setting of iterator either by method or constructor. Added some tests as well. I also added support for setting the iterator via assignment as long as it is before the while statement. Regarding adding some more functionality to holders, it would have to be generic. Stuff like computeVarName() is specific to this particular cleanup only, unless I am missing something. Generic stuff could be useful though.
> 
> Looks great already, feel free to merge when it is fine for you. Maybe additional functionality as eg. providing a way to give you the variables a while is running on can be investigated later as well. A breaking change is possible as long as there is no release, right?

Ok, thanks.  Will do a final look to see if there are any format tweaks needed.  Also need to add a N&N entry.

By breaking changes, I assume you mean tweaking your new public APIs; yes, you can do that before the final 4.25 release.  The new APIs are marked as a minor release bump compared to 4.24 and not compared to 4.25 M1, M2 etc...  Getting them solidified before 4.25 is out is a must.

carstenartur (commented):
Hi Jeff,
do you have any prefered way to activate the junit5-based suite in the build?
Or stick to plain junit5 tests?

jjohnstn:
No idea.  I think Alex said a while back that JUnit5 stuff is not possible due to the automated build process.  I would suggest doing as the Romans do for the time-being and conforming to the current JDT UI JUnit test level and adding to their suites so the test gets run as part of PR verification, nightly builds.  JUnit5 conversion is something that can be done as a whole for JDT in the future.

HannesWell:
Thank you for this enhanced clean-up!
I assume this is the fix of https://bugs.eclipse.org/bugs/show_bug.cgi?id=575078, isn't it?

carstenartur:
> Thank you for this enhanced clean-up!
> 
> I assume this is the fix of https://bugs.eclipse.org/bugs/show_bug.cgi?id=575078, isn't it?

I think you are right.
Looks like it is the same so can be closed if you think it is good enough. I did not think about it because this work was started unrelated to the mentioned issue at an earlier point in time

https://git.eclipse.org/r/c/jdt/eclipse.jdt.ui/+/167142

And then in a different github project.
I'm not sure how the duplication of issue tracking in jdt should be handled now as we have github *and* bugzilla. A reference doesn't hurt though.

HannesWell:
> And then in a different github project. I'm not sure how the duplication of issue tracking in jdt should be handled now as we have github _and_ bugzilla. A reference doesn't hurt though.

Yes this looks good 👍🏽 I'll just link the Bug to this issue and close it as fixed. It's not the first duplicated report.
I wasn't aware of your previous work and just noticed from the N&N entry.

I just applied the clean-up to the Equinox code base and found a few problems, but I'm about to report them in a separate issue.
Thanks again for your work.

carstenartur:
> > And then in a different github project. I'm not sure how the duplication of issue tracking in jdt should be handled now as we have github _and_ bugzilla. A reference doesn't hurt though.
> 
> Yes this looks good 👍🏽 I'll just link the Bug to this issue and close it as fixed. It's not the first duplicated report. I wasn't aware of your previous work and just noticed from the N&N entry.
> 
> I just applied the clean-up to the Equinox code base and found a few problems, but I'm about to report them in a separate issue. Thanks again for your work.

Kudos should go to Jeff, who is able to understand some of the Voodoo code of jdt where I‘m lost and implemented very important parts of the cleanup. Bugs are mine.
Thanks for extracting all these reproducers. This is really helpful. If you feel like doing even more you could implement the reproducers as pull request. However it should be straight-forward so no requirement.

HannesWell:
> Kudos should go to Jeff, who is able to understand some of the Voodoo code of jdt where I‘m lost and implemented very important parts of the cleanup. Bugs are mine. 

Every contribution helps so KUDOs to you both :) This is a clean up that can be applied to a lot of places in Equinox, P2, PDE and likely trough out the entire Platform.

> Thanks for extracting all these reproducers. This is really helpful. If you feel like doing even more you could implement the reproducers as pull request. However it should be straight-forward so no requirement.

Sorry for not creating the integration test PR, but I shared your assessment that it would be straight forward for you that are familiar with the tests, and that it would take me just more that is better invested in improving those parts of Eclipse I'm more familiar with. :)

org.eclipse.jdt.ui.tests/pom.xml
```diff
@@ -39,7 +39,8 @@
           <useUIThread>true</useUIThread>
           <includes>
             <include>org/eclipse/jdt/ui/tests/AutomatedSuite.class</include>
-            <!-- <include>org/eclipse/jdt/ui/tests/LeakTestSuite.class</include> -->
+            <include>org/eclipse/jdt/internal/common/JUnit5TestSuite.class</include>
```
ktatavarthi, line 42: There is no JUnit5TestSuite. If you plan to add it, please add it to this plugin.

carstenartur: > There is no JUnit5TestSuite. If you plan to add it, please add it to this plugin.

It got lost somehow in rebase.
Can you have a look at 

https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/129

Where I added it again why it is not executed?

PR #68, opened 2022-05-26
Ignore leading white-space when finding element at offset

jjohnstn:
- modify SelectionConverter.getElementAtOffset() to by-pass white
  space at beginning of ITextSelection so that enclosing element
  will be correct and not find the parent element
- fixes https://bugs.eclipse.org/bugs/show_bug.cgi?id=67077

## What it does

When calculating element at offset for an ITextSelection, ignore leading white-space so that selecting a set of lines for a member
does not end up returning the Type it is contained in.  This causes a problem for users that select an entire member and try to add
Javadoc comment using Source menu.

## How to test

See above and also: https://bugs.eclipse.org/bugs/show_bug.cgi?id=67077

## Author checklist


iloveeclipse:
@jjohnstn : please check https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/175. I believe it is caused by this PR.

PR #71, opened 2022-05-27
Character methods should not be used in some cases

carstenartur:
For full support of unicode for java identifiers some code needs changes. Character.toLowerCase and Character.toUpperCase have issues with unicode.

Eclipse has some issues to process java files written making use of full java spec.
While it is not recommended to make use of it in some cases it might make sense in the long run to allow some more freedom.
To start support at one day we need some preparation and change some coding to allow that in future.

see
https://bugs.eclipse.org/bugs/show_bug.cgi?id=26529
https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Character.html#supplementary
http://www.javawenti.com/?post=26640

### DIFF
--- a/org.eclipse.jdt.ui/core refactoring/org/eclipse/jdt/internal/corext/refactoring/code/IntroduceParameterRefactoring.java
+++ b/org.eclipse.jdt.ui/core refactoring/org/eclipse/jdt/internal/corext/refactoring/code/IntroduceParameterRefactoring.java
@@ -489,7 +489,7 @@ public class IntroduceParameterRefactoring extends Refactoring implements IDeleg
 			if (! Character.isUpperCase(firstAfterPrefix))
 				continue; //not uppercase after prefix
 			//found matching prefix
-			String proposal= Character.toLowerCase(firstAfterPrefix) + methodName.substring(prefix.length() + 1);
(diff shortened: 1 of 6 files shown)


org.eclipse.jdt.ui/core refactoring/org/eclipse/jdt/internal/corext/refactoring/code/IntroduceParameterRefactoring.java
```diff
@@ -489,7 +489,7 @@ private List<String> guessTempNamesFromMethodInvocation(MethodInvocation selecte
 			if (! Character.isUpperCase(firstAfterPrefix))
```
jjohnstn, line 489: Doesn't this need to be changed as well?

carstenartur: Yes, I think so - and there is more code to change. I just wanted to have a small one to get familar with the contribution process in jdt.ui now. Maybe that one is too much..

carstenartur: Should we ever try to adopt full java spec in jdt we can investigate again. To shorten the queue (summer holiday ended🤷‍♂️) I close this one.

PR #74, opened 2022-05-31
update jar-in-jar loader from java 6 to java 8

carstenartur:
For the feature export runnable jar eclipse makes use of a package that is built out of parts of the jdt.ui source code configured as java 11 compatible. It is implmented as ant build step ripping out the needed files of the sourcetree. As it is not in a separate project you always have to take care to not make use of java features possible with newer java standard than java 6.
 
https://help.eclipse.org/latest/index.jsp?topic=%2Forg.eclipse.jdt.doc.user%2Freference%2Fref-export-runnable-jar.htm

Currently it is built with source and target java 6 compatibility and is generating warnings doing that. 
I cannot find any specification for jdt or eclipse what target java runtime should be supported for 2022-06 but I guess the minimum is java 8.

## What it does
The patch bumps the version of the source and target version of java in the ant file to create the jar-in-jar loader from java 6 to java 8. Someone with knowledge about what is officially supported should comment on what is the right version to support. 
It does not rip out the code used here into a separate module what would be better. But even then it might be the right thing to update the java version for the code.

## How to test
exported java projects including third party libraries should work as long as the java version is greater or equal 8.


iloveeclipse (changes requested):
@carstenartur : I see in IDE that your PR is a merge PR now, containing lot of unrelated changes. I can't even find the actual change in the list of files :-) So please re-create your change only on top of master branch.

For the future, please don't push any merge commits or merge in github, always rebase your change in the IDE before.

The way I'm working is: 
- checkout "master" branch and pull from origin repo to master
- create branch on top of master "issue_42" (set "rebase" policy in the new branch dialog in Eclipse)
- hack & commit on that "issue_42" branch
- push  "issue_42" branch to my own "iloveeclipse" clone & create PR from that
- hack & amend commit on that branch if I need any changes
- checkout "master" and pull from origin to get latest commits
- checkout "issue_42", in history view right click "master" and say "Rebase head on"
- **force** push  "issue_42" branch to "iloveeclipse" repo (PR will be updated automatically)

For the current PR: I honestly also have no idea what requirements we have there, but I guess it is OK to assume Java 8 as minimal supported version.

PR #91, opened 2022-06-08
Inline module info question #59

### LINKED ISSUE #59: Inline module info question
Would it be possible to inline the options of module info in the Java project wizard page?

image

All the other options are set on the first page, only the module info is done via an additional popup which feels weird and IMHO is bad UX.

ktatavarthi:
Inline module info question #59. Fixes #59

Signed-off-by: Kalyan Prasad Tatavarthi <kalyan_prasad@in.ibm.com>

### DIFF
--- a/org.eclipse.jdt.ui/ui/org/eclipse/jdt/ui/wizards/NewJavaProjectWizardPageOne.java
+++ b/org.eclipse.jdt.ui/ui/org/eclipse/jdt/ui/wizards/NewJavaProjectWizardPageOne.java
@@ -906,7 +906,7 @@ public class NewJavaProjectWizardPageOne extends WizardPage {
 		public String getModuleNameOrProjectName() {
 			String moduleName = getModuleName();
 			if (moduleName == null || moduleName.length() == 0) {
-				moduleName= fNameGroup.getName().trim();
+				moduleName= fNameGroup.getName();
 			}
 			return  moduleName;
 		}
--- a/org.eclipse.jdt.ui/ui/org/eclipse/jdt/ui/wizards/NewJavaProjectWizardPageOne.java
+++ b/org.eclipse.jdt.ui/ui/org/eclipse/jdt/ui/wizards/NewJavaProjectWizardPageOne.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2018 IBM Corporation and others.
+ * Copyright (c) 2000, 2022 IBM Corporation and others.
  *
  * This program and the accompanying materials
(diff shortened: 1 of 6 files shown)


org.eclipse.jdt.ui/ui/org/eclipse/jdt/ui/wizards/NewJavaProjectWizardPageOne.java
```diff
@@ -1440,6 +1579,31 @@ public boolean getCreateModuleInfoFile() {
...
+	 * Returns the module name given.
+	 *
+	 * @return 'Module name:' text field value, removed of all spaces. If empty , returns
+	 * the project name, removed of all spaces.
+	 * @since 3.27
+	 */
+	public String getModuleName() {
+		String moduleName = fModuleGroup.getModuleName();
+		if (moduleName == null || moduleName.length() == 0) {
+			moduleName= fNameGroup.getName().replaceAll(WHITE_SPACE_STR, EMPTY_STR);
```
SarikaSinha, line 1592: Do we need to do this or just throw invalid error like we do in Create module-info dialog?

SarikaSinha: Looks good now.

org.eclipse.jdt.ui/ui/org/eclipse/jdt/ui/wizards/NewJavaProjectWizardPageOne.java
```diff
@@ -804,17 +811,31 @@ public IWorkingSet[] getSelectedWorkingSets() {
 		}
 	}
 
-	private final class ModuleGroup implements IDialogFieldListener{
+	private final class ModuleGroup extends Observable implements IDialogFieldListener{
```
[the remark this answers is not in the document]
SarikaSinha: Looks good now.

SarikaSinha:
Looks good.

SarikaSinha:
Not part of this change, but looking at the code noticed that should the source and compliance be same as derived now or should be retrieved from Java Project like this: 
javaProject.getOption(JavaCore.COMPILER_SOURCE, true)
javaProject.getOption(JavaCore.COMPILER_COMPLIANCE, true)

ktatavarthi:
> Not part of this change, but looking at the code noticed that should the source and compliance be same as derived now or should be retrieved from Java Project like this: javaProject.getOption(JavaCore.COMPILER_SOURCE, true) javaProject.getOption(JavaCore.COMPILER_COMPLIANCE, true)

We cannot use this because we are creating the java project here.

PR #110, opened 2022-06-19
Add number of changed files to PreviewWizardPage

carstenartur:
Add number of changed files to PreviewWizardPagel
bump pom file version
Use number of files within existing description
use two strings for single and multi case instead of one

## What it does
Instead of (currently) presenting this text after cleanups have been searching for hits:

The following changes are necessary to perform the refactoring.

you get this text:

The following change to 1 file is necessary to perform the refactoring.

or 

The following changes to {0} files are necessary to perform the refactoring.

That way you are able to see a difference even in case of a very huge number of hits when changeing the selected cleanups.
Otherwise you get a list presented but for hundreds of hits it is impossible to tell the number of hits increased or decreased.

## How to test
Use the context menu for cleanup on a java project and play with the selection of cleanups. You should now see the number of hits immediately in the preview presented.

## Author checklist


jjohnstn:
Hi Carsten, there is still a PR for this functionality from 4.24 that I mean to merge into 4.25 M1.  Does this replace it and if so, why is it a draft?

carstenartur:
> Hi Carsten, there is still a PR for this functionality from 4.24 that I mean to merge into 4.25 M1.  Does this replace it and if so, why is it a draft?

Both should be the same. I just wondered what the build error is about and if it would be better to keep the name of the public static variable the same instead of appending a "_s". If you want to merge the other pull request please do so!

PR #111, opened 2022-06-20
WIP context assist on type experiment

gayanper:
## Author checklist


gayanper:
@mickaelistria @noopur2507 Please suggest me few reviewers to get feedback and provide your feedback as well.

Adding for feedback:
@stephan-herrmann 
@iloveeclipse 
@mpalat

gayanper:
I will search and report a issue for this if one doesn't exist.

mickaelistria:
I think it's https://bugs.eclipse.org/bugs/show_bug.cgi?id=101420 . I hope to look at this proposal soon.

gayanper:
> I think it's https://bugs.eclipse.org/bugs/show_bug.cgi?id=101420 . I hope to look at this proposal soon.

@mickaelistria  Found an issue in Windows, even though the it is configured not to show empty list, the empty list is shown in windows, but not in macos.

gayanper:
> I think it's https://bugs.eclipse.org/bugs/show_bug.cgi?id=101420 . I hope to look at this proposal soon.

If you are running I-Builds of Eclipse, you can install this change and related [jface.text mods](https://github.com/eclipse-platform/eclipse.platform.text/pull/49) using this updatesite patch-site - https://gayanper.github.io/org.gap.eclipse.jdt.eap/p2/i-builds

humphreygao:
You type something in one line, and find that your input is totally wrong, so you decide to keep pressing backspace. When there are only leading whitespaces in one line, the content assist menu should not pop up, but it pops up every time you press backspace.

https://user-images.githubusercontent.com/52394598/175967866-62f6b9f6-9d8c-4e57-8d3a-c30540d1b2fa.mp4

Was it possible to disable content assist when you press backspace but there are only whitespaces(or nothing) in one line?

gayanper:
> 

I think i can see how to solve this issue. Thanks for testing and catching this scenario. Did you install the changes from the update site i shared which container jface patch as well ?

gayanper:
I think this behavior is by default in assist popup. But i think we can improve it. There is no point in keeping the popup after you erased your completion token.

humphreygao:
>  Did you install the changes from the update site i shared which container jface patch as well ?
yes

org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/text/java/ContentAssistProcessor.java
```diff
@@ -730,4 +748,46 @@ public ICompletionProposal getSelectedProposal() {
...
+			}
+		}
+		return false;
+	}
+
+	public final void setAutoActivateOnType(boolean autoActivateOnType) {
+		fAutoActivateOnType= autoActivateOnType;
+	}
+
+	public final boolean isAutoActivateOnType() {
```
mickaelistria, line 785: I don't see it used anywhere. Is it expected?

mickaelistria (commented):
I like the progress being made, however I believe that it shouldn't be the responsibility of the processor to decide whether to auto activate on typing. This is more a responsibility of the popup as it affects all processors in the same way.

org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/text/java/ContentAssistProcessor.java
gayanper: No this is just the getter that is added to comply with bean pattern, But i think we can remove these properties since it make sense to have it in the content assist it self rather than the processors to be beneficial in other places.

gayanper:
> I like the progress being made, however I believe that it shouldn't be the responsibility of the processor to decide whether to auto activate on typing. This is more a responsibility of the popup as it affects all processors in the same way.

@mickaelistria check the latest patches, I have moved auto activation into jface contentassist.

gayanper:
@mickaelistria and @jjohnstn please have a look at the changes, We can try to merge this for next release (4.26) M1.

gayanper:
@mickaelistria @noopur2507 should we merge this and try this out with M1 ?

mickaelistria:
@gayanper Sorry, I cannot decide nor act here, I'm not a committer on JDT.

Issue #109, opened 2022-06-19
Problems with "While loop to foreach loop conversion "


HannesWell:
I just applied the enhanced [Convert while loops to enhanced for loops](https://www.eclipse.org/eclipse/news/4.25/jdt.php#while-to-enhanced-for) clean-up introduced in https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/55 to the equinox code base (The unchanged result is at https://github.com/eclipse-equinox/equinox/pull/45) and encountered the following problems. The option "Only if loop variable used" was enabled and all other clean-ups were disabled.

- [x] Loop variable of `Map.entrySet()` iterator is not typed correctly, the second argument is missing:
```
	public static void before() {
		Map<String, Object> map = Map.of("Hello", new Object());
		Iterator<Entry<String, Object>> iterator = map.entrySet().iterator();
		while (iterator.hasNext()) {
			Entry<String, Object> entry = iterator.next();
			System.out.println(entry);
		}
	}

	public static void after() {
		Map<String, Object> map = Map.of("Hello", new Object());
		for (Entry<String> entry : map.entrySet()) {
			System.out.println(entry);
		}
	}
```
- [x] The import of java.util.Iterator is not always removed if not used anymore. I noticed it is removed most often but in the equinox-code base for [ModuleDatabase.java](https://github.com/eclipse-equinox/equinox/pull/45/files#diff-9c1e3f220b7f4b01c47e22ccf78778c76428f9575afa32a303523de97442f542) the clean-up fails to remove it.
- [x] The conversion of a counting-loop over a list fails if the current entry is assigned to two different variables (although I suspect this is a bug in the corresponding code that the second variable is assigned from the same list). If two separate lists use the same counter-variable the clean-up correctly does nothing:
```
	private void before1() {
		List<String> list1 = List.of("");
		for (int i = 0; i < list1.size(); i++) {
			String s1 = list1.get(i);
			String s2 = list1.get(i);
			System.out.println(s1 + "," + s2);
		}
	}

	private void after1() {
		List<String> list1 = List.of("");
		for (String s1 : list1) {
			System.out.println(s1 + "," + s2);
		}
	}

	private void before2() {
		List<String> list1 = List.of("");
		List<String> list2 = List.of("");
		for (int i = 0; i < list1.size(); i++) {
			String s1 = list1.get(i);
			String s2 = list2.get(i);
			System.out.println(s1 + "," + s2);
		}
	}

[...]

PR #114, opened 2022-06-21
Fix while loop to enhanced for loop conversion issue with EntrySet github issue 109

carstenartur:
https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/109

	public static void before() {
		Map<String, Object> map = Map.of("Hello", new Object());
		Iterator<Entry<String, Object>> iterator = map.entrySet().iterator();
		while (iterator.hasNext()) {
			Entry<String, Object> entry = iterator.next();
			System.out.println(entry);
		}
	}

	public static void after() {
		Map<String, Object> map = Map.of("Hello", new Object());
		for (Entry<String> entry : map.entrySet()) {
			System.out.println(entry);
		}
	}

In this case the Entry<String> is wrong and should be Entry<String,Object>

## What it does
The code now considers a more complex case Entry<String,Object>

## How to test
Run the junit test case or a code sample as given above where loop if over a map.entrySet()

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/fix/helper/WhileToForEach.java
+++ b/org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/fix/helper/WhileToForEach.java
@@ -425,16 +425,12 @@ public class WhileToForEach extends AbstractTool<WhileLoopToChangeHit> {
 			looptargettype= type2.resolveBinding().getErasure().getQualifiedName();
 			Type collectionType= ast.newSimpleType(addImport(looptargettype, cuRewrite, ast));
 			ParameterizedType genericType= ast.newParameterizedType(collectionType);
-			if(typeArguments.size()==1) {
-				object= typeArguments.get(0);
-				String fullyQualifiedName= object.getName().getFullyQualifiedName();
-				genericType.typeArguments().add(ast.newSimpleType(ast.newName(fullyQualifiedName)));
-			} else if(typeArguments.size()==2){
-				object= typeArguments.get(0);
+			object= typeArguments.get(0);
+			String fullyQualifiedName= object.getName().getFullyQualifiedName();
+			genericType.typeArguments().add(ast.newSimpleType(ast.newName(fullyQualifiedName)));
+			if(typeArguments.size()==2){
 				SimpleType object2= typeArguments.get(1);
-				String fullyQualifiedName= object.getName().getFullyQualifiedName();
 				String fullyQualifiedName2= object2.getName().getFullyQualifiedName();
-				genericType.typeArguments().add(ast.newSimpleType(ast.newName(fullyQualifiedName)));
 				genericType.typeArguments().add(ast.newSimpleType(ast.newName(fullyQualifiedName2)));
 			}
 			result.setType(genericType);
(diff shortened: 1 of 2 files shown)


carstenartur:
Hi @jjohnstn , Hi @HannesWell ,
for the first part of https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/109 this patch might be a solution.
Feel free to give it a try.
Best regards,
Carsten

jjohnstn:
Thanks Carsten.  It does solve the While loop issue.  I have enhanced the patch to handle nested parameterized types (e.g. Map<List<String>, Object>).  The other issues are for the old for iterator loop fix code.

HannesWell:
Great, thank you both!
I'll try it again as soon as it is available with the new I-Build.

HannesWell:
I just tested this fix and it worked well for maps with simple types and types that themself are parameterized with generic arguments.

But unfortunately it still fails with array types:

```
Map<Date[], Date[]> map = Map.of(new Date[0], new Date[1]);
Iterator<Entry<Date[], Date[]>> iterator = map.entrySet().iterator();
while (iterator.hasNext()) {
	Entry<Date[], Date[]> entry = iterator.next();
	System.out.println(entry);
}
```
Should be converted to
```
Map<Date[], Date[]> map = Map.of(new Date[0], new Date[1]);
for (Entry<Date[], Date[]> entry : map.entrySet()) {
	System.out.println(entry);
}
```
but is converted to
```
Map<Date[], Date[]> map = Map.of(new Date[0], new Date[1]);
for (Entry<> entry : map.entrySet()) {
	System.out.println(entry);
}
```

If generics and arrays are mixed
```
Map<List<Date[]>, Date[]> map = Map.of(List.of(), new Date[1]);
Iterator<Entry<List<Date[]>, Date[]>> iterator = map.entrySet().iterator();
while (iterator.hasNext()) {
	Entry<List<Date[]>, Date[]> entry = iterator.next();
	System.out.println(entry);
}
```
becomes
```
Map<List<Date[]>, Date[]> map = Map.of(List.of(), new Date[1]);
for (Entry<List<>> entry:map.entrySet()){
	System.out.println(entry);
}
```

Or the other ways round, Array of a generic type (although this is not warning free):
```
Map<List<String>[], Date[]> map = Map.of(new List[0], new Date[1]);
Iterator<Entry<List<String>[], Date[]>> iterator = map.entrySet().iterator();
while (iterator.hasNext()) {
	Entry<List<String>[], Date[]> entry = iterator.next();
	System.out.println(entry);
}
```
is falsely cleaned up to
```
Map<List<String>[], Date[]> map = Map.of(new List[0], new Date[1]);
for (Entry<> entry : map.entrySet()) {
	System.out.println(entry);
}
```

I suspect that in the just added addGenericTypes() method the case `typeArgument instanceof ArrayType` is missing (not sure if the class is named ArrayType, but I mean an array type).

jjohnstn:
Yes, the addGenericTypes() method needs support for ArrayType.  Working on it.

jjohnstn:
I have removed the addGenericTypes() method.  There is a more straight-forward way to set the type which I got from the older iterator-for-loop code.  It doesn't require building the type.

HannesWell:
> I have removed the addGenericTypes() method. There is a more straight-forward way to set the type which I got from the older iterator-for-loop code. It doesn't require building the type.

That worked well, I just re-run the clean up on Equinox. Thank you.

PR #115, opened 2022-06-21
Forloopbug

jjohnstn:
## What it does
Fixes part of issue 109

Don't convert a for loop to enhanced for loop that refers to a collection's get() method more than once in the loop body.

## How to test
Try the Convert to enhanced for loop cleanup on:

	public void before1() {
		List<String> list1 = List.of(""); //$NON-NLS-1$
		for (int i = 0; i < list1.size(); i++) {
			String s1 = list1.get(i);
			String s2 = list1.get(i);
			System.out.println(s1 + "," + s2); //$NON-NLS-1$
		}
	}

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/fix/ConvertForLoopOperation.java
+++ b/org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/fix/ConvertForLoopOperation.java
@@ -462,6 +462,7 @@ public class ConvertForLoopOperation extends ConvertLoopOperation {
 		Statement body= statement.getBody();
 		try {
 			body.accept(new GenericVisitor() {
+				private boolean fGetSeen= false;
 				@Override
 				protected boolean visitNode(ASTNode node) {
 					if (node instanceof ContinueStatement) {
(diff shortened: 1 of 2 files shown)


HannesWell:
Thank you for this fix. This way the case reported in #109 is fixed.

But it looks like the assumption is too strict, because cases like the following are not converted anymore:

```
List<String> stringList = List.of("Hello", "World");
for (int i = 0; i < stringList.size(); i++) {
	System.out.println(stringList.get(i));
	System.out.println(stringList.get(i));
}
```
Here I still would expect (which was also the outcome before):
```
List<String> stringList = List.of("Hello", "World");
for (String element : stringList) {
	System.out.println(element);
	System.out.println(element);
}
```

For arrays that case still works:
```
String[] stringArray = new String[] { "Hello", "World" };
for (int i = 0; i < stringArray.length; i++) {
	System.out.println(stringArray[i]);
	System.out.println(stringArray[i]);
}
```
is converted to 
```
String[] stringArray = new String[] { "Hello", "World" };
for (String element : stringArray) {
	System.out.println(element);
	System.out.println(element);
}
```

jjohnstn:
Hi Hannes, I have replaced the patch with one that properly converts the loop and doesn't block the cleanup.  I have added a test to CleanUpTest1d5.

HannesWell:
Great. Thank you.

Do you think it would be good to have another test case that ensures that

```
List<String> stringList = List.of("Hello", "World");
for (int i = 0; i < stringList.size(); i++) {
	System.out.println(stringList.get(i));
	System.out.println(stringList.get(i));
}
```
is converted to
```
List<String> stringList = List.of("Hello", "World");
for (String element : stringList) {
	System.out.println(element);
	System.out.println(element);
}
```
Or do you think the current cases are sufficient?

PR #128, opened 2022-06-29
Move trailing comments with initializer. Fixes #56

### LINKED ISSUE #56: Quickfix/cleanup removes //$NON-NLS-1$
**Cleanup**
```

-		File file = null;
-		file = new File(homeDir, "eclipse"); //$NON-NLS-1$
+		File file = new File(homeDir, "eclipse");
```

**Quickfix**

```
fBrowseFile.addSelectionListener(new SelectionAdapter() {
-			@Override
-			public void widgetSelected(SelectionEvent e) {
-				chooseFile(fArchiveCombo, new String[] {"*" + ZIP_EXTENSION, "*" + WAR_EXTENSION}); //$NON-NLS-1$ //$NON-NLS-2$
-			}
-		});
+		fBrowseFile.addSelectionListener(widgetSelectedAdapter(e -> chooseFile(fArchiveCombo, new String[] {"*" + ZIP_EXTENSION, "*" + WAR_EXTENSION})));
```

tsmaeder:
## What it does
This PR makes sure any line end comments are moved with the corresponding initializer when cleaning up an overriding assignment. The new code checks that both the initializer and the overriding assignment are alone on a line except for a trailing comment. Otherwise the clean-up is not applied. The cleanup then replaces the line end comment in the declaration with the one in the overriding assignment. 
This is a fix for #56

## How to test
Experiment with overriding assignments inside blocks, in particular in the presence of the line end comments. For example, see the related issue.

### DIFF
--- a/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/fix/OverriddenAssignmentCleanUp.java
+++ b/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/fix/OverriddenAssignmentCleanUp.java
@@ -23,11 +23,13 @@ import org.eclipse.core.runtime.CoreException;
 import org.eclipse.text.edits.TextEditGroup;
 
 import org.eclipse.jdt.core.ICompilationUnit;
+import org.eclipse.jdt.core.JavaModelException;
 import org.eclipse.jdt.core.dom.ASTNode;
 import org.eclipse.jdt.core.dom.ASTVisitor;
 import org.eclipse.jdt.core.dom.Assignment;
 import org.eclipse.jdt.core.dom.Block;
 import org.eclipse.jdt.core.dom.CompilationUnit;
+import org.eclipse.jdt.core.dom.Expression;
 import org.eclipse.jdt.core.dom.ExpressionStatement;
 import org.eclipse.jdt.core.dom.IBinding;
 import org.eclipse.jdt.core.dom.IVariableBinding;
@@ -36,6 +38,7 @@ import org.eclipse.jdt.core.dom.Statement;
 import org.eclipse.jdt.core.dom.VariableDeclarationFragment;
 import org.eclipse.jdt.core.dom.VariableDeclarationStatement;
 import org.eclipse.jdt.core.dom.rewrite.ASTRewrite;
+import org.eclipse.jdt.core.dom.rewrite.TargetSourceRangeComputer;
 
 import org.eclipse.jdt.internal.corext.dom.ASTNodes;
 import org.eclipse.jdt.internal.corext.dom.InterruptibleVisitor;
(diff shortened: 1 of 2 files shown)


ktatavarthi:
If this PR is ready, please go ahead and delivery it by EOD today, for M1.

org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/fix/OverriddenAssignmentCleanUp.java
```diff
@@ -159,42 +159,39 @@ public boolean visit(final VariableDeclarationStatement node) {
...
-					lineEnd= unit.getLength();
-				}
-				String textBefore= cu.getBuffer().getText(lineStart, node.getStartPosition()-lineStart);
-				if (!textBefore.isBlank()) {
-					return false;
+				int stmtStartLine= unit.getLineNumber(stmt.getStartPosition());
+				int stmtEndLine= unit.getLineNumber(stmt.getStartPosition() + stmt.getLength());
+				Statement prevStmt= ASTNodes.getPreviousSibling(stmt);
+				if (prevStmt == null) {
+					ASTNode parent= stmt.getParent();
```
tsmaeder, line 182: I'm not really which condition you're trying to catch here: we're in the "this is the first statement in the block" case, right? So if we go up the parent chain, we'll end up with a method declaration, for example, no? Or are we looking at other types of blocks like this?
```
if (foo == false) { String foo= bar; //NON-NLS-1
``

**Comment by tsmaeder** [comment]
As far as comments go, I don't see how we can consistently do the right thing. Consider this piece of code (from `Parser.java` in jdt.core)
```
	// MarkerAnnotation ::= AnnotationName
	// MarkerTypeAnnotation ::= TypeAnnotationName
	MarkerAnnotation markerAnnotation = null;

	int oldIndex = this.identifierPtr;

	TypeReference typeReference = getAnnotationType();
	// some really important comment on the initilizer here
        markerAnnotation = new MarkerAnnotation(typeReference, this.intStack[this.intPtr--]);
```
it gets transformed to 
```
	int oldIndex = this.identifierPtr;

	TypeReference typeReference = getAnnotationType();
	MarkerAnnotation markerAnnotation = new MarkerAnnotation(typeReference, this.intStack[this.intPtr--]);
```

So we lose the leading comments. In the general case, which one we should keep cannot be determined.

jjohnstn:
In your example, the reasonable solution would be to always preserve leading comments of lines that aren't removed.  So, in your example, the comment: // some really important comment on the initializer here
should be preserved.  It's validity does not change by making the line a declaration rather than an assignment.  If you moved the assignment up, then the comments for the initial declaration should be preserved.  There will be cases where the user will want something else, but that is up to the user to change manually or undo the cleanup.

jjohnstn:
For reference.  There is already precedence: When a statement is removed, it takes out its leading comments.  For example:

// Some comment
int t = 4;

// Some other comment
t = 5;

and if t is an unused local, removing t and all its references, removes // Some comment
and also // Some other comment

Similarly, when a statement is transformed (such as a while -> for loop or etc...), the leading comments of the statement that is transformed are preserved.  So, in your example above, the removed lines should remove the comments as well, and the transformed lines should keep the comment.  This is already true when you transform a declaration and add the final value.  This just has to be done for the case where we move the declaration down.

I am preparing a change for this and will modify this patch and its test case.

PR #147, opened 2022-07-07
Prevent moving initializer in error cases.

### LINKED ISSUE #136: OverridingAssignmentCleanup moves up initializer with undefined Variable.
Running the `OverridingAssignmentCleanup` on such a code block:
```
		char[] result = null;
		...
		int sigLength = sig.length();
		result = new char[sigLength];
```
will result in this code:
```
		char[] result = new char[sigLength];
		...
		int sigLength = sig.length();
```
which does not compile due to `sigLength` not being defined.

To reproduce, turn on the "Remove overridden assignment" cleanup and run it on the above code.

tsmaeder:
## What it does
Checks two things and prevents moving the initializer if true:

1. Whether a variable is reassigned in a statement that may have sub-
statements (like an if statement)
2. Whether the initializer uses variables that are not define at the
place the initializer would be move up to.

 Fixes #136

## How to test
* Have a looks at the new test cases introduced and play around with similar cases
* Apply cleanup to Eclipse source code.

### DIFF
--- a/org.eclipse.jdt.ui.tests/ui/org/eclipse/jdt/ui/tests/quickfix/CleanUpTest.java
+++ b/org.eclipse.jdt.ui.tests/ui/org/eclipse/jdt/ui/tests/quickfix/CleanUpTest.java
@@ -17524,6 +17524,86 @@ public class CleanUpTest extends CleanUpTestCase {
 	}
 
 	@Test
+	public void testDoNotMoveAssignment() throws Exception {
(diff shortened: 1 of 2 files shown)


stephan-herrmann:
Sorry, I don't feel qualified to review this, as I've never touched this code area.

Not sure if github provides an option to formally resign from a requested review?

tsmaeder:
>  requested review from

It's a request, not an order ;-)

jjohnstn:
Looks good.

Issue #148, opened 2022-07-08
Parameter name code mining not supported for record constructor invocation


rgrunber:
Originally discussed at https://github.com/redhat-developer/vscode-java/issues/2414

JDT does't support parameter name code minings for record constructor invocations. This is because the logic relies on accessing the constructor node's java model (`IMethod`), but the method is synthetic so `null`. In JDT-LS they try to use the record's field declaration to get the parameter names, but order is not guaranteed so that isn't correct.

It's possible to expose the ordered parameter names of the record's constructors using the method binding so we just need to expose it in jdt.core, and then write some code to construct the mining from a method binding.

**Result with proposed changes :**
![](https://user-images.githubusercontent.com/1417342/174167406-f0aad370-c608-49e4-902f-06f1cfd17f6d.png)

PR #149, opened 2022-07-08
Add parameter name code mining support for records.

rgrunber:
- Fixes https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/148
- Use IMethodBinding#getParameterNames() to support parameter name code
  mining for records

This change requires https://github.com/eclipse-jdt/eclipse.jdt.core/pull/201 to use the new API for IMethodBinding.

### DIFF
--- a/org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/codemining/ParameterNamesCodeMiningTest.java
+++ b/org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/codemining/ParameterNamesCodeMiningTest.java
@@ -78,7 +78,7 @@ public class ParameterNamesCodeMiningTest {
 			closeIntro(PlatformUI.getWorkbench());
 		}
 		fProject= JavaProjectHelper.createJavaProject(getClass().getName(), "bin");
-		JavaProjectHelper.addRTJar_16(fProject, true);
+		JavaProjectHelper.addRTJar_17(fProject, true);
 
 		Map<String, String> options= fProject.getOptions(false);
 		JavaProjectHelper.set17_CompilerOptions(options);
--- a/org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/codemining/ParameterNamesCodeMiningTest.java
+++ b/org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/codemining/ParameterNamesCodeMiningTest.java
@@ -18,6 +18,7 @@ import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
 
 import java.util.Arrays;
+import java.util.Map;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.LongSupplier;
@@ -77,7 +78,12 @@ public class ParameterNamesCodeMiningTest {
 			closeIntro(PlatformUI.getWorkbench());
 		}
 		fProject= JavaProjectHelper.createJavaProject(getClass().getName(), "bin");
-		JavaProjectHelper.addRTJar(fProject);
+		JavaProjectHelper.addRTJar_16(fProject, true);
+
+		Map<String, String> options= fProject.getOptions(false);
+		JavaProjectHelper.set17_CompilerOptions(options);
+		fProject.setOptions(options);
+
 		IPackageFragmentRoot root= JavaProjectHelper.addSourceContainer(fProject, "src");
 		fPackage= root.getPackageFragment("");
 		fParameterNameCodeMiningProvider= new JavaMethodParameterCodeMiningProvider();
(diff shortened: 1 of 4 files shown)


org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/javaeditor/codemining/JavaMethodParameterCodeMining.java
```diff
@@ -40,6 +41,17 @@ public JavaMethodParameterCodeMining(Expression parameterNode, int parameterInde
 		}
 	}
 
+	public JavaMethodParameterCodeMining(Expression parameterNode, int parameterIndex, IMethodBinding mbinding, boolean isVarargs, ICodeMiningProvider provider) {
```
tsmaeder, line 44: The constructor is the exact same code except for `binding.getParameterNames()`. Please extract the common code either to a method or to the call sites of the constructor.

org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/javaeditor/codemining/CalleeJavaMethodParameterVisitor.java
```diff
@@ -104,25 +104,34 @@ public boolean visit(MethodInvocation methodInvocation) {
...
 				IMethod method = resolveMethodBinding(methodBinding);
-				collectParameterNamesCodeMinings(method, arguments, methodBinding.isVarargs());
+				collectParameterNamesCodeMinings(method, arguments, methodBinding);
 			}
 		}
 		return super.visit(methodInvocation);
 	}
 
-	protected void collectParameterNamesCodeMinings(IMethod method, List<?> arguments, boolean isVarargs) {
+	protected void collectParameterNamesCodeMinings(IMethod method, List<?> arguments, IMethodBinding mbinding) {
```
tsmaeder, line 113: This methods has one part that uses only the `IMethod`, the other only the `IMethodBinding`. This should be a method overload with two different parameterTypes and the dispatch can be done in a third method.

tsmaeder, line 113: Also, we have a method binding != null at each call site. Why do we even bother with computing the `IMethod` which may be `!= null` if we already have an object that can serve just as well?

org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/javaeditor/codemining/CalleeJavaMethodParameterVisitor.java
```diff
@@ -154,6 +163,18 @@ private boolean skipParameterNameCodeMining(IMethod method, List<?> arguments, i
 		}
 	}
 
+	private boolean skipParameterNameCodeMining(IMethodBinding mbinding, List<?> arguments, int parameterIndex) {
```
tsmaeder, line 166: This method is a duplicate of the other `skipParameterNameCodeMining(...)` method, except for how we get the `parameterNames` array. Please remove the code duplication.

org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/javaeditor/codemining/JavaMethodParameterCodeMining.java
tsmaeder, line 44: Again, I think we have a method binding in ever call site of this constructor. Why bother with the `IMethod` at all?

tsmaeder (changes requested):
While the change seems pretty trivial, there is code duplication which can easily be avoided, as far as I understand.

tsmaeder:
@rgrunber how can I test this change?

org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/javaeditor/codemining/CalleeJavaMethodParameterVisitor.java
rgrunber: You're right. I think with this API addition from jdt.core, there's no need for `IMethod` at all. We need the method binding for `isVarArgs()` so if that needs to exist, we don't need to bother also getting the parameter names. I'll make the change and see if it things work as expected.

rgrunber: I've tried eliminating `IMethod#getParameterNames()` and just using the `IMethodBinding#getParameterNames()` but it doesn't seem to work in some cases. It breaks [`ParameterNamesCodeMiningTest`](https://github.com/eclipse-jdt/eclipse.jdt.ui/blob/master/org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/codemining/ParameterNamesCodeMiningTest.java). Looks to be empty for certain testcases, while `IMethod@getParameterNames()` does work. No idea why yet.

On the plus side, I didn't think this had testcases, so may as well add one for records to this PR.

tsmaeder:
@rgrunber should I have another look or should are you still working on this?

rgrunber:
@tsmaeder , let me see if I can figure out why bindings are empty in some other cases. I know there's some obscure cases like extracting parameter names from `.class` files, so maybe the java element gets that information but bindings don't :\ I know generally I've preferred using the java element for info and used bindings as a last resort.

rgrunber:
I haven't been able to determine the exact reason that using just the bindings doesn't work. As an example, the bindings do not contain the parameter names for a method when the library is compiled without debugging information (`-g`) and a source attachment exists. The java model on the other hand seems to consult the source attachments for the parameters and works as expected.

From that alone, I think it makes sense to continue to use `IMethod#getParameterNames()`, and only fall back to `IMethodBinding#getParameterNames()` when the former doesn't exist (eg. a synthetic method)

@tsmaeder , I think the PR is ready for review, given that going through just the bindings would regress some functionality.

tsmaeder:
@rgrunber that makes sense, given that the names are optional in https://docs.oracle.com/javase/specs/jvms/se18/html/jvms-4.html#jvms-4.7.24. So no debug info, no parameter name, I guess.

org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/codemining/ParameterNamesCodeMiningTest.java
```diff
@@ -77,7 +78,12 @@ public void setUp() throws Exception {
 			closeIntro(PlatformUI.getWorkbench());
 		}
 		fProject= JavaProjectHelper.createJavaProject(getClass().getName(), "bin");
-		JavaProjectHelper.addRTJar(fProject);
+		JavaProjectHelper.addRTJar_16(fProject, true);
+
+		Map<String, String> options= fProject.getOptions(false);
+		JavaProjectHelper.set17_CompilerOptions(options);
```
tsmaeder, line 84: rt.jar from 16, but compiler options for 17?

tsmaeder (approved):
Looks good to me, +/- the mismatched versions in the test.

org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/codemining/ParameterNamesCodeMiningTest.java
rgrunber: Looks like `addRTJar_17(..)` didn't exist so I probably just took the path of least resistance in the tests. I'll go ahead and create the method.

tsmaeder:
@rgrunber if you are able, I'll let you merge this in proper sequence.

PR #150, opened 2022-07-10
Detect/avoid illegal combination of --release and --add-exports

stephan-herrmann:
This is a follow-up from issue [eclipse.jdt.core#114](https://github.com/eclipse-jdt/eclipse.jdt.core/issues/114), where a new buildpath error has been implemented.

In this current PR we try to signal the problem right while configuring a project:
* **Build Path**: When `--release` is set, the Module Dependencies tab will disable the "Expose Package" button for all system modules - providing an info message why this is not possible.
* **Compiler Options**: When `--add-exports` exists on a system module, let selection of `--release` raise an error, thus disabling "Apply".

I chose this slightly asymmetric design, because
* I didn't find a suitable why to explain the problem on the Compiler Options page short of raising an error when the illegal combination is selected.
* On the Module Dependencies tab I preferred to disable that button, so the user is spared the effort of selecting the details for --add-exports only to learn that this is illegal from the outset. Selecting a system module seemed a suitable trigger for showing the info about why "Expose Package" is not possible.

To test simply create a project with JRE > 9 and compliance less than the JRE version, thus auto-enabling `--release`: "Expose Package" will be disabled for system modules. Conversely, when `--release` is deselected, and an export is added to a system module, after apply, re-enabling `--release` will no longer be possible.

Two caveats: 
* When starting from disabled `--release`, going to compiler options to select `--release` then switching to Build Path without Apply, "Expose Package" will still be enabled, because `--release` is not yet effective. A build path error will be reported when the project is built on final Apply.
* JDT/Core doesn't yet fully understand when exactly classpath validation needs to be triggered, because at this point not real .classpath changes are relevant but also changes to the compiler option `--release`. I'll file a follow-up in JDT/Core about this.

### DIFF
--- a/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/preferences/ComplianceConfigurationBlock.java
+++ b/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/preferences/ComplianceConfigurationBlock.java
@@ -21,6 +21,8 @@ import java.util.Hashtable;
 import java.util.List;
 import java.util.Map;
 
+import com.ibm.icu.text.MessageFormat;
+
 import org.osgi.service.prefs.Preferences;
 
 import org.eclipse.swt.SWT;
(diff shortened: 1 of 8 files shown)


stephan-herrmann:
> JDT/Core doesn't yet fully understand when exactly classpath validation needs to be triggered, because at this point not real .classpath changes are relevant but also changes to the compiler option `--release`. I'll file a follow-up in JDT/Core about this.

https://github.com/eclipse-jdt/eclipse.jdt.core/issues/202

PR #162, opened 2022-07-13
Add new buttons to Cleanup preference pages

### LINKED ISSUE #119: Add '(De)Select all' buttons to Clean-up Configuration dialog and ability to search for a specific clean-up
It would be handy if the Java  Code Style Clean-Up Preference dialog would have Buttons to `Deselect all` and `Select all` available clean ups. Especially the former avoids a lot of manual unchecking if one you want to have a clean configuration to then select only one or a few certain clean ups.

Furthermore it would super convenient if the Clean-Up preferences dialog would offer something like a search bar, where you can type in the name (label, id, etc.) and get the corresponding clean up elements displayed.
The Java -> Compiler Errors/Warnings dialog offers such search bar, but also has a different structure. Therefore I don't know how this should work for different tabs.

I assume that the dialog itself is the same for 
- `Preferences -> Java Code Style -> Clean Up -> Edit`
- `Preferences -> Java -> Editor -> Save Actions -> Configure (Additional actions)`
- `Right click on a project in the Explorer -> Source -> Clean up... -> Configure (Use custom profile)`

jjohnstn:
- add new Select All, Deselect All, Defaults, and Reset buttons
  for each cleanup preference page
- fixes #119

## What it does
Adds new buttons to the Cleanup preference pages to allow:  Select All, Deslect All, Defaults, and Reset to original values

## How to test
Do Source->Cleanup and configure a profile

## Author checklist


org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/preferences/formatter/ModifyDialogTabPage.java
```diff
@@ -862,8 +863,85 @@ public void controlResized(ControlEvent e) {
...
+		});
+
+		b= new Button(buttonHolder, SWT.PUSH);
+		b.setText(FormatterMessages.ModifyDialogTabPage_deselectAll_text);
+		gridData= new GridData(SWT.LEFT, SWT.TOP, false, true);
+		b.setLayoutData(gridData);
+		b.setEnabled(true);
+		b.setVisible(true);
+		SWTUtil.setButtonDimensionHint(b);
+		b.addSelectionListener(new SelectionAdapter(){
```
HannesWell, line 910: Using `SelectionListener.widgetSelectedAdapter()` would allow you to use a lambda and therefore would make the code more compact.
This could also be applied to the following two occurrences.

HannesWell (commented):
Thank you Jeff for working on this.

In my little testing this worked well.

I understand your rational to provide the buttons per page as you explained in https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/119#issuecomment-1183490076.
But for the scenario where one wants to start from scratch and wants to clear all selections it would be more convenient to have a Deselect All buttons that applies to all pages, maybe at the bottom? It is definitely better to the previous state but one still has to go through 9 tabs.
I think for `Defaults` and `Reset` it could be convenient as well if there were is variant applicable for all.
For example the launch-configuration dialog also provides `Revert` and `Apply` buttons that work for all tabs, even the not selected ones.

But actually the UI should not be overloaded with Buttons for all cases so I'm torn here and don't know the optimal solution.

org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/preferences/formatter/ModifyDialogTabPage.java
jjohnstn: Not a fan of lambdas, but sure, done.

noopur2507:
> @noopur2507 What do you think?

Commented on [#119](https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/119).

noopur2507 (commented):
Commented on [#119](https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/119.)

jjohnstn:
New patch based on @noopur2507 comments on issue.

cdietrich:
we at xtend have an "illegal" subclass of ModifyDialogTabPage.
The new abstract methods break this class.
i wonder if they could be modified to be non abstract to keep compatibility

jjohnstn:
Hi @cdietrich what would like to propose?  The methods could be made non-abstract protected with a default no-op implementation but then what do you do with the additional buttons? (noop??)

cdietrich:
yes, having a noop -  this is how i am currently gonna implement it, instead of copy & pasting all the old jdt code.
unfortunately i dont have the time to properly implement it currently.

jjohnstn:
@cdietrich I have posted a patch which I will try to get into M2.  Will you be making changes for M3 and if so, can I change the code back to abstract?  The PR is: https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/179

howlger:
Nice feature!
Minor glitch: _Select All_ selects also all radio buttons (e.g. in the _Code Style_ tab) and _Deselect All_ deselects all radio buttons which does not make sense. _Select/Deselect All_ should ignore radio buttons.

jjohnstn:
Hi @howlger.  This needs to be fixed.  I have created: https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/253

PR #168, opened 2022-07-18
Add JUnit5 support to New TestSuite Wizard

### LINKED ISSUE #142: [JUnit5] New JUnit Test Suite Project wizard does not give option to create JUnit 5 Suite
We get the options to Create JUnit 3 and JUnit 4 Test Suite.
Should we not got an option to create JUnit 5 Test Suite?

jjohnstn:
- fixes #142
- modify NewTestSuiteWizardPage, SuiteClassesContentProvider, and UpdateTestSuite classes

## What it does
Add JUnit 5 support to the NewTestSuiteWizard

## How to test
Invoke the NewTestSuiteWizard via the New -> Other -> Java -> JUnit -> JUnit Test Suite

Select JUnit 5 Test Suite from the radio buttons

## Author checklist

### DIFF
--- a/org.eclipse.jdt.junit/src/org/eclipse/jdt/junit/wizards/NewTestSuiteWizardPage.java
+++ b/org.eclipse.jdt.junit/src/org/eclipse/jdt/junit/wizards/NewTestSuiteWizardPage.java
@@ -197,11 +197,8 @@ public class NewTestSuiteWizardPage extends NewTypeWizardPage {
 				isJunit4= true;
 			}
 			isJunit5= CoreTestSearchEngine.hasJUnit5TestAnnotation(project);
-			if (!isJunit5 && !CoreTestSearchEngine.hasTestCaseType(project) && JUnitStubUtility.is18OrHigher(project)) {
-				isJunit5= true;
-			}
 		}
-		setJUnit4or5(isJunit5, isJunit5, true);
+		setJUnit4or5(isJunit4, isJunit5, true);
 		doStatusUpdate();
 	}
 
@@ -216,7 +213,7 @@ public class NewTestSuiteWizardPage extends NewTypeWizardPage {
 			setAddComments(StubUtility.doAddComments(project), true); // from project or workspace
 		}
 
-		setTypeName(ALL_TESTS, true);
+		setTypeName(generateName(ALL_TESTS), true);
 		fTypeNameStatus= typeNameChanged(); //set status on initialization for this dialog - user must know that suite method will be overridden
 	}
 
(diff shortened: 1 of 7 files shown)


org.eclipse.jdt.junit/src/org/eclipse/jdt/junit/wizards/NewTestSuiteWizardPage.java
```diff
@@ -294,7 +309,8 @@ protected void createClassesInSuiteControl(Composite parent, int nColumns) {
 			gd.horizontalSpan= nColumns-1;
 
 			fClassesInSuiteTable.getTable().setLayoutData(gd);
-			fClassesInSuiteTable.setContentProvider(new SuiteClassesContentProvider(isJUnit4()));
+			System.out.println("is junit4 " + isJUnit4() + " is junit5 " + isJUnit5()); //$NON-NLS-1$ //$NON-NLS-2$
```
noopur2507, line 312: Please remove the sysout statement.

noopur2507: The default radio button selection in the wizard should be New JUnit 5 suite whenever it's possible to create a JUnit 5 suite in the workspace, or the selection in Package Explorer, etc. You can refer to the behavior of New Test Case wizard.

For example, if the selected project contains junit suite bundle then New JUnit 5 suite should be selected by default. 
And when the wizard is invoked on a package or test class(es), if the selection contains a JUnit 5 test, then the wizard should select JUnit 5 suite.

Also, noticed that while editing something in the wizard (like selecting/deselecting test classes, changing suite name, etc.), the radio button selection is getting reset to 4 if 5 was selected before editing. I don't have exact steps to reproduce but it happened a couple of times.

noopur2507 (changes requested):

org.eclipse.jdt.junit/src/org/eclipse/jdt/junit/wizards/NewTestSuiteWizardPage.java
jjohnstn: @noopur2507 Done and done.  I also fixed the wizard to not always select AllTests which causes an error or warning when the file already exists.  It will default to pick AllTestsxxxx where xxxx is a number that isn't already used.  The user has the option to remove xxxx and overwrite the existing file if desired  I found this annoying while testing and figure a New TestSuite wizard should default to creating a new testsuite.

noopur2507: Thanks, Jeff. I tried out the latest patch and it looks good now. We can release it for M3.

noopur2507 (approved):

jjohnstn:
Thx @noopur2507 .

PR #177, opened 2022-07-26
Whitespaceissue

jjohnstn:
- for all white-space selection, don't set a white-space offset
  in SelectionConverter.getElementAtOffset() and don't cause IOBException

## What it does

Fixes problem with selecting all white-space characters

## How to test

Select area with only white-space characters and right-click

## Author checklist

### DIFF
--- a/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/actions/SelectionConverter.java
+++ b/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/actions/SelectionConverter.java
@@ -283,7 +283,7 @@ public class SelectionConverter {
 		String selectedString= selection.getText();
 		if (selectedString != null && !selectedString.isEmpty()) {
 			int i= 0;
-			while (Character.isWhitespace(selectedString.charAt(i))) {
+			while (i < selectedString.length() && Character.isWhitespace(selectedString.charAt(i))) {
 				++i;
 		    }
 			whiteSpaceOffset= i;
--- a/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/actions/SelectionConverter.java
+++ b/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/actions/SelectionConverter.java
@@ -288,6 +288,9 @@ public class SelectionConverter {
 		    }
 			whiteSpaceOffset= i;
 		}
+		if (selectedString != null && whiteSpaceOffset == selectedString.length()) {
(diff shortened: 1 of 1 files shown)


org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/actions/SelectionConverter.java
```diff
@@ -283,11 +283,14 @@ public static IJavaElement getElementAtOffset(ITypeRoot input, ITextSelection se
 		String selectedString= selection.getText();
 		if (selectedString != null && !selectedString.isEmpty()) {
 			int i= 0;
-			while (Character.isWhitespace(selectedString.charAt(i))) {
+			while (i < selectedString.length() && Character.isWhitespace(selectedString.charAt(i))) {
```
carstenartur, line 286: Wouldn't it be better to change coding to allow support full java specification while on it?
Don't know if there is any plan to support full java spec for jdt though.
This is using the char based method that does not support full unicode character set
https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/Character.html

https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/71/files

jjohnstn: I think it would be best to do a mass JDT change in an M1 milestone to ensure all instances are covered with some form of testing rather than doing this piecemeal hidden within other patches.  If @iloveeclipse disagrees,  I'll be happy to add the supplementary char logic in as I have already coded it using Character.codePointAt().  This required additional calls and would add time to perform the loop, otherwise it would be a no-brainer to add it in now.

iloveeclipse: Jeff, feel free whatever you want, as long as it would reduce bug count :)

Please don't wait for my review if you want merge.

iloveeclipse (approved):

org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/actions/SelectionConverter.java
jjohnstn: Ok, thx @iloveeclipse.  I added you as a reviewer because there wasn't a particular scenario for the issue and though I was able to reproduce I just wanted to make sure it worked for your use case.  Weirdly enough, adding you as a reviewer also adds Till Brychy too ??

Issue #184, opened 2022-07-30
Content assist in nested try/catch block sometimes doesn't work


iloveeclipse:
Given example below
```
import java.io.File;
import java.util.stream.Stream;

public class X {
	public static void main(String[] args) throws Exception {
		try {
			try {
				Stream<File> stream = Stream.of(new File("")); // place cursor anywhere here
				System.out.println(stream);
			} catch (Exception e) {
			}
		} catch (Exception e) {
		}
	}
}
```
place cursor inside the first line inside inner try block and try to invoke `Ctrl+1` for content assist.
There will be only one "No suggestions available" entry shown and in the error log error below will be reported twice.

This is a regression from f42020ec00b26af423e9f055036e5cf19f0c3cdb 
```
java.lang.IndexOutOfBoundsException: Index 0 out of bounds for length 0
	at java.base/jdk.internal.util.Preconditions.outOfBounds(Preconditions.java:64)
	at java.base/jdk.internal.util.Preconditions.outOfBoundsCheckIndex(Preconditions.java:70)
	at java.base/jdk.internal.util.Preconditions.checkIndex(Preconditions.java:266)
	at java.base/java.util.Objects.checkIndex(Objects.java:359)
	at java.base/java.util.ArrayList.get(ArrayList.java:427)
	at org.eclipse.jdt.internal.ui.text.correction.QuickAssistProcessor.getTryWithResourceProposals(QuickAssistProcessor.java:3168)
	at org.eclipse.jdt.internal.ui.text.correction.QuickAssistProcessor.getTryWithResourceAssistProposals(QuickAssistProcessor.java:3093)
	at org.eclipse.jdt.internal.ui.text.correction.QuickAssistProcessor.getAssists(QuickAssistProcessor.java:371)
	at org.eclipse.jdt.internal.ui.text.correction.JavaCorrectionProcessor$SafeAssistCollector.safeRun(JavaCorrectionProcessor.java:404)
	at org.eclipse.jdt.internal.ui.text.correction.JavaCorrectionProcessor$SafeCorrectionProcessorAccess.run(JavaCorrectionProcessor.java:341)
	at org.eclipse.core.runtime.SafeRunner.run(SafeRunner.java:45)
	at org.eclipse.jdt.internal.ui.text.correction.JavaCorrectionProcessor$SafeCorrectionProcessorAccess.process(JavaCorrectionProcessor.java:330)
[...]

PR #185, opened 2022-07-30
Use right statements list (#184)

iloveeclipse:
QuickAssistProcessor.getTryWithResourceProposals() used wrong variable
to process proposals: coveredNodes was used instead of
coveredStatements, and beside this, the check for the collections size
was missing. Fixed both and added regression test case.

Fixes https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/184


iloveeclipse:
@jjohnstn : could you please take a look?

jjohnstn (approved):
Looks good.

Issue #192, opened 2022-08-02
Some comments get lost on "enhanced for loop cleanup"


carstenartur:
Here an example
![image](https://user-images.githubusercontent.com/3164220/182440569-32d8ae9d-d99a-40ee-90df-b1d619e32f91.png)

See
https://github.com/eclipse-pde/eclipse.pde/pull/251/files

for more samples

PR #193, opened 2022-08-02
fix lost comments issue 192

carstenartur:
see https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/192

## What it does
It solves the lost comments issue and adds/changes a test
While it solves the issue with the comments the testWhileNested5() test is failing. It is unclear if it is really related.

## How to test
contains junit test

## Author checklist

- [ x] The change is following the [coding conventions](https://wiki.eclipse.org/Platform/How_to_Contribute#Coding_Conventions)


jjohnstn:
Hi Carsten.  The Tight source range computer is necessary to keep the error you are seeing from happening.  I ran into this while making the fix that added it because failures occur when comments are at the end of block.  That said, the accepted strategy is that line comments preceding a line, belong to that line,and if it is removed, they are also removed even though there are times when it could be a general comment.  In sync with this, is the strategy of keeping a comment for a line that is modified even though it is possible it may require updating.  So, in your added test, it is the correct thing to remove Comment 1 while keeping Comment 2.  For example,

```
// iterator for loop
Iterator<Integer> x = k.iterator();
// loop until some condition
while (x.hasNext()) {
   Integer k = x.next();
   System.out.println(k);
}
```

carstenartur:
Ok, I wasn't aware of that to be accepted.
In the real world test

https://github.com/eclipse-pde/eclipse.pde/pull/251/files

I got the impression it should keep some more comments but it is fine for me to accept it as it is given this is generally accepted. So I close it.
Thanks for having a look!

PR #195, opened 2022-08-03
create junit test for failing case

carstenartur:
This provides a junit test for the failing case. It can be discussed if it is better (easier) at this point just to detect the assignment of the iterator after the first hit and stay away from refactoring or detect that there is no real problem (just reuse of the variable) and create another loop.

## What it does
so for only adds @Ignored test

## How to test
--

## Author checklist


jjohnstn:
Hi @carstenartur I have made a fix that prevents the conversion if the iterator variable is reassigned or if two while loops use the same iterator.  This test is no longer viable.

PR #199, opened 2022-08-04
Sort build path container entries by label

kjsmita6:
## What it does
Sorts the build path "Libraries" entries alphabetically by its label. When you have a project with a lot of library dependencies (say, OSGi plugin dependencies), having them not be ordered makes it very difficult to find anything. One of our plugins has 508 plugin dependencies, for example. 
It's nice to be able to find things reliably and not have them placed in a seemingly-random spot. As before, ordering is delegated to the label provider's `getText(Object)` method.

Example before:
image

Example after:
image

## How to test
1. Have any project in the IDE that has some kind of build path "Library" dependency. JRE library included (i.e., any Java project can work).
2. Select Project Properties-->Java Build Path-->Libraries, expand one of the containers (such as "JRE System Library"), and see that its entries are in alphabetical order.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/wizards/buildpaths/CPListElementSorter.java
+++ b/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/wizards/buildpaths/CPListElementSorter.java
@@ -74,8 +74,8 @@ public class CPListElementSorter extends ViewerComparator {
         if (cat1 != cat2)
             return cat1 - cat2;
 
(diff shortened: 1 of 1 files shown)


Bananeweizen:
Looks good to me. Strictly speaking, I would also change the comparator to be case insensitive (you want A_foo before z_bar), but that will not have relevance for the module names.

noopur2507:
I see the expected behaviour as the default behaviour in I20230913-1800. Please reopen if anything is still pending.

Issue #201, opened 2022-08-05
Refactor rename reports shadowing when there is none


trancexpress:
**To reproduce:**

1. Import projects from: [RefactorRenameShadowingBug.zip](https://github.com/eclipse-jdt/eclipse.jdt.ui/files/9268291/RefactorRenameShadowingBug.zip)
2. Open type ```test1.impl.ITes`t``.
3. Refactor rename ```foo()``` to ```foo1()```.
4. Observe reported shadowing problems:

![RefactorRenameShadowingBug](https://user-images.githubusercontent.com/24752155/183072217-2859c609-534e-4da3-b54a-eed65219d864.png)

There is no shadowing though, the rename is fine.

**Cause:**

Looks like JDT UI "thinks" there are no search matches for the new method in the mentioned files, after the rename. So it reports shadowing (as it thinks the rename did nothing).

This is the case due to the order of elements passed to: ```RenameMethodProcessor.batchFindNewOccurrences()``` (and in particular, delegated to ```RefactoringScopeFactory.create(IMember[]))```. ```test2.Foo``` seems to be found first by ```RippleMethodFinder2.getRelatedMethods()``` and so only its project is used for the search. I.e. the scope is created with only the project ```Test2``` and so the search misses some occurrences. The refactor rename then thinks those missed occurrences are shadowing, instead of subject to the rename.

The problem is created with this stack trace:

```
"ModalContext" #199 prio=6 os_prio=0 cpu=45.89ms elapsed=16.76s tid=0x00007ffff3240000 nid=0x3dce at breakpoint [0x00007fff29cf4000]
   java.lang.Thread.State: RUNNABLE
        at org.eclipse.jdt.internal.corext.refactoring.rename.RenameAnalyzeUtil.addShadowsError(RenameAnalyzeUtil.java:387)
        at org.eclipse.jdt.internal.corext.refactoring.rename.RenameAnalyzeUtil.analyzeRenameChanges2(RenameAnalyzeUtil.java:277)
        at org.eclipse.jdt.internal.corext.refactoring.rename.RenameMethodProcessor.analyzeRenameChanges(RenameMethodProcessor.java:572)
        at org.eclipse.jdt.internal.corext.refactoring.rename.RenameMethodProcessor.doCheckFinalConditions(RenameMethodProcessor.java:412)
[...]

PR #204, opened 2022-08-05
Refactor rename wrongly reports shadowings#201

trancexpress:
During a refactor rename, its possible that new occurrences of the
renamed methods are not correctly detected by
RenameMethodProcessor.batchFindNewOccurrences().

This is due to batchFindNewOccurrences() creating a scope based on the
set of new (renamed) methods, using
RefactoringScopeFactory.create(IMember[]). This create() method
essentially picks the project of the first of the specified IMember
objects, at least for trivial refactor renames. However that project
might not be sufficient to search for all new renamed method
occurrences, it might instead be a project that is a leaf dependency and
not a root dependency.

This change adjusts the used scope to be that of the project scopes of
all new methods provided to batchFindNewOccurrences(), instead of just
using the first method.

Fixes: #201

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/refactoring/RefactoringScopeFactory.java
+++ b/org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/refactoring/RefactoringScopeFactory.java
@@ -15,8 +15,10 @@
 package org.eclipse.jdt.internal.corext.refactoring;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.HashSet;
+import java.util.LinkedHashSet;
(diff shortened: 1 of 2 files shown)


trancexpress:
Mainly opening the PR to check for test fails. I'm not 100% sure this is the right approach to fix the problem.

So far I don't understand why the search scope is created the way it is, so I'm also not sure what the best fix would be.

trancexpress:
OK, I think the current change makes sense. @iloveeclipse please review.

iloveeclipse:
@jjohnstn : Jeff, is this something you would like to look at?

jjohnstn:
@iloveeclipse Will do

jjohnstn (approved):
Looks good.  Merging.

trancexpress:
> Looks good. Merging.

Thanks!

PR #226, opened 2022-08-21
Use tree selection for picking the nested test classes

gzsombor:
instead of a list, which makes picking test cases easier.

## What it does

It changes the dialog which is shown when "Run As/Debug As" -> "JUnit Test" action is executed in a file, which has multiple test 
classes. Previously, only a simple listing of the found types are shown, which is not the most optimal solution, if there are multiple nested classes, the original structure of the classes are lost. Specially, finding the root class is not trivial. 
 Changing the dialog to show the tree structure of the classes, helps tremendously.

## How to test

 Create a **JUnit** test class with multiple inner classes, with **Nested** annotations and test methods in it. Try to start them...

## Author checklist

### DIFF
--- a/org.eclipse.jdt.junit.core/META-INF/MANIFEST.MF
+++ b/org.eclipse.jdt.junit.core/META-INF/MANIFEST.MF
@@ -3,7 +3,7 @@ Automatic-Module-Name: org.eclipse.jdt.junit.core
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.jdt.junit.core;singleton:=true
-Bundle-Version: 3.11.400.qualifier
+Bundle-Version: 3.11.500.qualifier
 Bundle-Activator: org.eclipse.jdt.internal.junit.JUnitCorePlugin
 Bundle-ActivationPolicy: lazy
 Bundle-Vendor: %providerName
--- a/org.eclipse.jdt.junit.core/pom.xml
+++ b/org.eclipse.jdt.junit.core/pom.xml
@@ -18,6 +18,6 @@
   </parent>
   <groupId>org.eclipse.jdt</groupId>
   <artifactId>org.eclipse.jdt.junit.core</artifactId>
-  <version>3.11.400-SNAPSHOT</version>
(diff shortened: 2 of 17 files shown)


jjohnstn:
Hi @gzsombor, I am having trouble getting this to work.  When I create the following nested test class, I don't see anything when I try Run as except for a pointer to Run Configurations.  What am I doing wrong according to your description above?  I tried within the file and by clicking on the file in the Project Explorer and neither worked.  Tests in the same package activate fine with Run as.

```
package test;

import static org.junit.Assert.fail;
import static org.junit.jupiter.api.Assertions.*;

import org.junit.jupiter.api.Test;

class TestMulti5 {

	public class TestMulti2 {
		@Test
		public void test() {
			fail("Not yet implemented");
		}
	}
	public class TestMulti3 {
		@Test
		public void test() {
			fail("Not yet implemented");
		}
	}
	public class TestMulti4 {
		@Test
		public void test() {
			fail("Not yet implemented");
		}
	}

}
```

gzsombor:
You need to use the `@Nested` annotation, like this:

```
package test;

import static org.junit.Assert.fail;

import org.junit.jupiter.api.Nested;
import org.junit.jupiter.api.Test;

class TestMulti5 {

	@Nested
	public class TestMulti2 {
		@Test
		public void test() {
			fail("Not yet implemented");
		}
	}
	@Nested
	public class TestMulti3 {
		@Test
		public void test() {
			fail("Not yet implemented");
		}
	}
	@Nested
	public class TestMulti4 {
		@Test
		public void test() {
			fail("Not yet implemented");
		}
	}

}

```

jjohnstn:
Ok, thanks, that worked.  I am unable to choose multiple items in the tree view.  I can select a class with nested tests but with the example above, I can't choose TestMulti2 and TestMulti4 and exclude TestMulti3.  I assume that is expected/desired behavior.

gzsombor:
Yes, we can pick only one test class - as previously https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/226/files#diff-58ab62bd5e3526882a6b511cb8db88cbdd4e8fc1ee2242f4bdfa2f773111cb61L257

jjohnstn (approved):
Looks good.

Issue #78, opened 2022-06-01
[content assist] Investigate why default completion list is slower than VSCode Java when completion is enabled for all keys


gayanper:
When completing same code in both VSCodeJava and Eclipse, Eclipse JDT completion popup is slow to be populated compared to VSCode completions. This needs to be investigated and improved to provide same or more better level of response time in JDT.

This could be because of lack of debouncing in the JDT.UI editor for keystrokes.

humphreygao:
When there are errors in Java Editor, the content assist is extremely slow, so I have to press `Esc`  to quit content assist and invoke content assist again.

gayanper:
> When there are errors in Java Editor, the content assist is extremely slow, so I have to press `Esc` to quit content assist and invoke content assist again.

Could you give a example code snippet. Then we can see what’s happening in side the completion engine.

PR #239, opened 2022-08-26
fix ClassCastException in some PropertyDialogAction #78

jukzi:
If element is PackageFragmentRoot then resource is Folder.
If resource is already IProject then getProject() returns this - as
before.

This solution was already suggested in bug 467761 and 579428.

To test: show Properties of "src" folder. Then Select "Java Compiler".


jjohnstn (approved):
Looks fine.  Will have to wait until master is open for 4.26 M1 to merge.

jukzi:
> Looks fine. Will have to wait until master is open for 4.26 M1 to merge.

@jjohnstn time to submit?

PR #279, opened 2022-09-24
dogfooding enhanced for loop

carstenartur:
## What it does
dogfooding: applies enhanced for loop cleanup to jdt.ui code base.
Shows a handful cases where a comment before the changed code gets lost. 

## How to test
This is to be tested in the junit tests.

## Author checklist

- [o] I have thoroughly tested my changes


carstenartur:
Hi @jjohnstn,
Regarding the comments before a change that get lost because of the cleanup we cannot do better, right?

jjohnstn:
I wouldn't say that.  The current behavior is delete a line, delete the comment it owns.  We could add preferences.  For example, we could have a preference to not remove a line comment for a removed statement in general.  We could also be more specific (e.g. don't remove a line comment for iterators that directly precede the loop).  We could alternatively make the latter idea part of the current code by default and only remove if the iterator is separated from the loop by other statements or try and be smart and look for the word iterator in the comment and remove if that is the case.  What do you think?

I would suggest you manually add back any comments you thought should have been left and modify the patch.

Issue #261, opened 2022-09-15
Outline view: don't show full annotations for method parameters


mmatela:
The Outline view doesn't show information about annotations on classes, methods or fields - which is good, because it's just an overview that doesn't need to go into such details. But for methods instead of a simple list of parameter types, we do see full annotations (including annotation parameters and their values). In projects that use annotations extensively, it makes the view less useful as we don't see the list of parameters, just the first part of the first parameter's annotations list.

I'm not even sure if a configuration would be necessary for this. Although as I was looking around the Outline settings, the "Visible Categories..." option looked promising. But it only shows a dialog with an empty list and nothing can be changed - what is that about?

PR #280, opened 2022-09-24
Outline view: don't show full annotations for method parameters.

mateusz-matela:
Fix for #261 

Turns out the behavior was the result of a mistake in change https://bugs.eclipse.org/bugs/show_bug.cgi?id=530689. A new constant `JavaElementLabelsCore.MOD_CATEGORY` was given the same value as an existing constant `JavaElementLabelsCore.M_PARAMETER_ANNOTATIONS`.
Easy mistake to make when the existing value of `1L << 52` was hidden at the top of the file, between `1L << 1` and `1L << 2`. Thus I also took the liberty to reorder the constants by their values to avoid similar problems in the future.


mateusz-matela:
Now I'm wondering, maybe the constants should not be reordered, but their values should be reorganized to be consecutive bits? Would that be safe? There are some unused bits and it's close to the 64 bit limit of long.

mateusz-matela:
Oh, as this is public API, it wouldn't be safe to reorganize all the constants.
Even changing the problematic constant could affect someone out there, but I guess it's still better than keeping a bug-causing conflict.

mateusz-matela:
Build errors:
```
00:13:47  [API ERROR] File MANIFEST.MF at line 6: The major version should be incremented in version 3.28.0, since API breakage occurred since version 3.27.0 - The constant value 8444249301319680 of the field org.eclipse.jdt.ui.JavaElementLabels.ALL_CATEGORY has been changed (location: /home/jenkins/agent/workspace/eclipse.jdt.ui-github_PR-280/org.eclipse.jdt.ui/META-INF/MANIFEST.MF)
00:13:47  [API ERROR] File JavaElementLabels.java at line 328: The constant value 8444249301319680 of the field org.eclipse.jdt.ui.JavaElementLabels.ALL_CATEGORY has been changed (location: /home/jenkins/agent/workspace/eclipse.jdt.ui-github_PR-280/org.eclipse.jdt.ui/ui/org/eclipse/jdt/ui/JavaElementLabels.java)
```

Does this mean changing the constant is only allowed with version 4.0.0?
So what now?

mateusz-matela:
Thanks for the suggestion @jjohnstn, but it won't help. The `JavaElementLabelsCore.M_PARAMETER_TYPES` flag is set to true for the outline view - thus methods are displayed with their parameters as lists of types. Even if it happened to help in this particular case, the API would still be broken for other cases.

I understand @noopur2507 can be busy, but I'd assume she could always find half a minute to delegate to someone else who can have a look at this. If not, I guess there is no one available?

jjohnstn:
I'll take over the review and look at it for 4.27 M1 as moving/redefining constants is too major for last-minute M3 which has it's deadline today and no major changes allowed.

mateusz-matela:
I just realized that `JavaElementLabelsCore` is internal so it's not necessary to preserve deprecated constants there. Only the `JavaElementLabels.ALL_CATEGORY` must be marked as deprecated and its value forced to be the same as was in old `JavaElementLabelsCore.ALL_CATEGORY`. But I'm not sure, maybe current approach will be slightly less confusing.

@jjohnstn will you be able to review before M1?

jjohnstn:
Hi @mateusz-matela, was on vacation.  Even though this is a day after the usual M1 freeze, I will merge as I promised I would look at it for 2023-03 M1 and would have merged if not for vacation.  If there are any unforeseen issues, I will revert.

PR #288, opened 2022-10-05
Move addTypeAsPermittedSubTypeProposal into jdt.core.manipulation.

rgrunber:
- Use CompilationUnitRewriteOperationsFixCore to define the AST rewrite required for the quick fix
- Use helper methods for retrieving the sealed type, subtype, and compilation unit for the sealed type
- Instead of 'problem.getProblemArguments()[1]', use the Java element name of the sealed type

## What it does

Moves `addTypeAsPermittedSubTypeProposal(..)` (from `LocalCorrectionsSubProcessor` of jdt.ui) to (to a new class, `SealedClassFixCore` of jdt.core.manipulation)

## How to test

The code being refactored is tested in [`QuickFixTest17`](https://github.com/eclipse-jdt/eclipse.jdt.ui/blob/master/org.eclipse.jdt.ui.tests/ui/org/eclipse/jdt/ui/tests/quickfix/QuickFixTest17.java), which I used to confirm there was no breakage.

You can also simply create a new class that inherits from a sealed class. The quick-fix that adds the new class to the `permits` expression of the sealed class should continue to function as before.

### DIFF
--- a/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/text/correction/LocalCorrectionsSubProcessor.java
+++ b/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/text/correction/LocalCorrectionsSubProcessor.java
@@ -126,6 +126,7 @@ import org.eclipse.jdt.core.dom.rewrite.ImportRewrite;
 import org.eclipse.jdt.core.dom.rewrite.ImportRewrite.ImportRewriteContext;
 import org.eclipse.jdt.core.dom.rewrite.ImportRewrite.TypeLocation;
 import org.eclipse.jdt.core.dom.rewrite.ListRewrite;
+import org.eclipse.jdt.core.refactoring.CompilationUnitChange;
 
 import org.eclipse.jdt.internal.core.manipulation.StubUtility;
 import org.eclipse.jdt.internal.core.manipulation.dom.ASTResolving;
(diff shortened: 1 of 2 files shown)


org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/text/correction/LocalCorrectionsSubProcessor.java
```diff
@@ -1353,65 +1355,24 @@ public static void createNewTypeAsPermittedSubTypeProposal(IInvocationContext co
...
-		ASTNode selectedNode= problem.getCoveringNode(context.getASTRoot());
-		if (selectedNode == null) {
-			return;
-		}
-		if (!ASTHelper.isSealedTypeSupportedInAST(selectedNode.getAST())) {
-			return;
-		}
+	public static void addTypeAsPermittedSubTypeProposal(IInvocationContext context, IProblemLocation problem, Collection<ICommandAccess> proposals) {
+		SealedClassFixCore fix = SealedClassFixCore.addTypeAsPermittedSubTypeProposal(context.getASTRoot(), (IProblemLocationCore)problem);
+		if (fix != null) {
```
jjohnstn, line 1360: In the future, when you want to do more refactoring like this where you move the code into a CompilationUnitRewriteOperationsFixCore, the following code could/should be put into a utility method where you pass it the fix, the image or image id, ICompilationUnit, whatever, and you get back a CUCorrectionProposal to add.

jjohnstn (approved):
Just the one comment about making the fix core wrapping a utility but ok if you want to wait until the next refactoring.

org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/text/correction/LocalCorrectionsSubProcessor.java
rgrunber: I could probably fix this to get it right. I actually intended to add another 1 or 2 quick-fixes into `SealedClassFixCore`.

rgrunber: Actually, I may keep this as-is for now and maybe we can adjust in a future PRs. If we provide some utility that does that I think it's just a matter of ensuring future refactorings use it.

PR #294, opened 2022-10-12
Move addSealedAsDirectSuperTypeProposal into jdt.core.manipulation.

rgrunber:
- Use SealedClassFixCore to define the AST rewrite required for the quick fix
- Use helper methods for retrieving the sealed type and permitted type
- Instead of 'problem.getProblemArguments()[0]', use the type declaration's identifier to get the simple name
- Add createCorrectionProposalFromCURewriteOperation(..) helper method

## What it does
Moves addSealedAsDirectSuperTypeProposal(..) (from LocalCorrectionsSubProcessor of jdt.ui) to (to a new class, SealedClassFixCore of jdt.core.manipulation)

## How to test
The code being refactored is tested in [QuickFixTest17](https://github.com/eclipse-jdt/eclipse.jdt.ui/blob/master/org.eclipse.jdt.ui.tests/ui/org/eclipse/jdt/ui/tests/quickfix/QuickFixTest17.java), which I used to confirm there was no breakage.

You can also simply create a sealed class that permits some class to inherit from it. The quick-fix that adds the sealed class as a direct super-class should continue to function as before.

### DIFF
--- a/org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/fix/SealedClassFixCore.java
+++ b/org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/fix/SealedClassFixCore.java
@@ -73,6 +73,38 @@ public class SealedClassFixCore extends CompilationUnitRewriteOperationsFixCore 
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


rgrunber:
Thanks. I'll have a last look and see if maybe I adjust the names on some of the helper methods, but after that I'll probably merge.

jjohnstn:
Ok, feel free.

PR #89, opened 2022-06-06
Fix "Remove overridden assignment" cleanup preview

howlger:
Fix the preview of the _Remove overridden assignment_ cleanup (in the _Unnecessary Code_ tab) when the option _Move declaration if necessary_ is not enabled.

See also https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/27

PR #296, opened 2022-10-16
Fix the Java editor not saving HTML to Clipboard (#89)

mihnita:
There are 3 tests failing (in my environment), but I checked and the same tests fail identically without my changes.

## Author checklist


mihnita:
The original bug was filed here:
https://github.com/eclipse-platform/eclipse.platform.debug/issues/89

Very likely that is the wrong place, but at the time I didn't know where the fix will end up.
Let me know if you want me to file an equivalent bug here (or close than one and open it here)

Thank you,
Mihai

PR #297, opened 2022-10-17
Fix "Add import" for static references through subclasses

fapdash:
I don't think the removed check was necessary in the first place. The containerName (the class that the method is called on) shouldn't have to be equal the methods declaringClass, otherwise we can't go up the class hierarchy.

fixes #276

PR #298, opened 2022-10-17
Remove use of deprecated SubProgressMonitor from AddImportsOperation

fapdash:
Following the [boy scout rule](https://www.informit.com/articles/article.aspx?p=1235624&seqNum=6) I removed the warnings from AddImportsOperation, which I changed in https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/297.

### DIFF
--- a/org.eclipse.jdt.ui/core extension/org/eclipse/jdt/internal/corext/codemanipulation/AddImportsOperation.java
+++ b/org.eclipse.jdt.ui/core extension/org/eclipse/jdt/internal/corext/codemanipulation/AddImportsOperation.java
@@ -22,7 +22,6 @@ import org.eclipse.core.runtime.Assert;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.core.runtime.IStatus;
-import org.eclipse.core.runtime.NullProgressMonitor;
 import org.eclipse.core.runtime.OperationCanceledException;
 import org.eclipse.core.runtime.Status;
 import org.eclipse.core.runtime.SubMonitor;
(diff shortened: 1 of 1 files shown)


jjohnstn (changes requested):
According to the Javadoc:
```

	/**
	 * <p>Converts an unknown (possibly null) IProgressMonitor into a SubMonitor allocated
	 * with the given number of ticks. It is not necessary to call done() on the result,
	 * but the caller is responsible for calling done() on the argument. Calls beginTask
	 * on the argument.</p>
	 * 
	 * <p>This method should generally be called at the beginning of a method that accepts
	 * an IProgressMonitor in order to convert the IProgressMonitor into a SubMonitor.</p> 
	 *
	 * <p>Since it is illegal to call beginTask on the same IProgressMonitor more than once,
	 * the same instance of IProgressMonitor must not be passed to convert more than once.</p>
	 * 
	 * @param monitor monitor to convert to a SubMonitor instance or null. Treats null
	 * as a new instance of <code>NullProgressMonitor</code>.
	 * @param work number of ticks that will be available in the resulting monitor
	 * @return a SubMonitor instance that adapts the argument
	 */
	public static SubMonitor convert(IProgressMonitor monitor, int work) {
		return convert(monitor, "", work); //$NON-NLS-1$
	}

```

The same monitor cannot be passed to multiple convert calls in a method which means you cannot call it multiple times as in 172 and 180 or am I reading this wrong above?

fapdash:
@jjohnstn Good catch, I was actually using `SubMonitor` completely wrong. I applied changes following the JavaDoc instructions.

org.eclipse.jdt.ui/core extension/org/eclipse/jdt/internal/corext/codemanipulation/AddImportsOperation.java
```diff
@@ -354,15 +348,12 @@ private TextEdit evaluateEdits(CompilationUnit root, ImportRewrite importRewrite
...
 		IJavaSearchScope searchScope= SearchEngine.createJavaSearchScope(new IJavaElement[] { fCompilationUnit.getJavaProject() });
 
-		TypeNameMatch[] types= findAllTypes(simpleName, searchScope, nameNode, new SubProgressMonitor(monitor, 1));
+		TypeNameMatch[] types= findAllTypes(simpleName, searchScope, nameNode, subMonitor.split(1));
 		if (types.length == 0) {
 			fStatus= JavaUIStatus.createError(IStatus.ERROR, Messages.format(CodeGenerationMessages.AddImportsOperation_error_notresolved_message, BasicElementLabels.getJavaElementName(simpleName)), null);
 			return null;
 		}
 
-		if (monitor.isCanceled()) {
```
jjohnstn, line 363: Any reason this is no longer valid?

jjohnstn (changes requested):

org.eclipse.jdt.ui/core extension/org/eclipse/jdt/internal/corext/codemanipulation/AddImportsOperation.java
fapdash: From what I understand, one of the features of `SubMonitor` is to provide an abstraction where the caller doesn't have to do explicit cancellation handling.

From the doc of `SubMonitor#split`:

> This method is much like newChild, but it will additionally check for cancellation and will throw an OperationCanceledException if the monitor has been cancelled.

This change moves the cancellation check (implicit check by calling `SubMonitor#split`) above the if block, but in my mind that doesn't really make a difference as the if block is just a guard clause that doesn't do a costly computation.

jjohnstn: Seems reasonable.

jjohnstn (approved):

PR #308, opened 2022-10-25
Allow OrganizeImportsOperation keeping existing imports

CsCherrYY:

## What it does

Allow `OrganizeImportsOperation` to have a change to restore existing imports. Currently it directly call ` CodeStyleConfiguration.createImportRewrite(astRoot, false)` to create corresponding `ImportRewrite`. If we do not want to change the current import statements but only want to add all missing imports, there is no easy way to achieve this. 

There is a scenario in JDT.LS to support "Add all missing imports" quick fix without changing the current import statements, see the related enhancement issue https://github.com/redhat-developer/vscode-java/issues/2748. With this change the implementation can make use of `OrganizeImportsOperation` directly, see the draft PR https://github.com/eclipse/eclipse.jdt.ls/pull/2292.
## How to test

With this change, we can use the new constructor to create a `OrganizeImportsOperation` instance with `restoreExistingImports` is true. The operation will not remove or sort the current imports, but only add missing imports.
## Author checklist

cc: @rgrunber

### DIFF
--- a/org.eclipse.jdt.core.manipulation/META-INF/MANIFEST.MF
+++ b/org.eclipse.jdt.core.manipulation/META-INF/MANIFEST.MF
@@ -3,7 +3,7 @@ Automatic-Module-Name: org.eclipse.jdt.core.manipulation
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.jdt.core.manipulation; singleton:=true
-Bundle-Version: 1.16.250.qualifier
+Bundle-Version: 1.17.0.qualifier
 Bundle-Vendor: %providerName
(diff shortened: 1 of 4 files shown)


rgrunber (commented):
`restoreExistingImports` says "specifies if the existing imports should be kept or removed." . Is this just another way of saying that no sorting will occur if `true` ? Just want to confirm this is the case as I can't think of any other side effect of this other than sorting and performance.

The version bump due to API seems correct to me.

CsCherrYY:
>  Is this just another way of saying that no sorting will occur if true?

Yes. I have add a test case to show what will happen when `restoreExistingImports` is set to `true` :)

rgrunber (approved):
Change looks fine to me. Just exposes existing functionality via. constructor parameter. Thanks for adding the test for this.

PR #309, opened 2022-10-25
Make some performance enhancements to test finders

### LINKED ISSUE #302: [JUnit] `JUnit4TestFinder` large project performance
Original discussion: https://bugs.eclipse.org/bugs/show_bug.cgi?id=540712

I have been investigating _Run As > JUnit Test_ start times for a while now.  I think there are some opportunities in `JUnit4TestFinder` to make some improvements.  

1. After `newTypeHierarchy`, all of the types from a source folder, package or project are added to the search scope.  Later, abstract or inaccessible types are excluded from the result.  Q: Is it necessary to search abstract types or anonymous types for test methods?  Is it possible in JUnit 4 to find any test cases this way?  [RxJava](https://github.com/ReactiveX/RxJava) has about 2000 test classes and over 7000 anonymous inner classes.  Minimally, I am proposing that `createJavaSearchScope` be passed only non-anonymous types.  This can be checked without fetching element info using `aClass.getElementName().isEmpty()` -- anonymous inner classes return an empty name.  The abstract test requires creating element info, so I can understand why you might wait until after.

1. After finding all test types, the `candidates` set is filtered by accessibility, abstractness and containment.  At the very least, the abstract check should be the least expensive and can be moved up front:
`if (!Flags.isAbstract(curr.getFlags()) && CoreTestSearchEngine.isAccessibleClass(curr) && region.contains(curr))`

1. Similarly in `JUnit5TestFinder` there are some expensive things like parsing within `internalIsTest`, so I propose changing the order of checks within `findTestsInContainer`:
`if (region.contains(type) && internalIsTest(type, pm))`

1. Lastly, within `JUnit5TestFinder#internalIsTest` is there a way to determine "is test" using only the `IType` passed in?  Is the `TypeBinding` the only way to make the determination?

https://github.com/eclipse-jdt/eclipse.jdt.ui/blob/master/org.eclipse.jdt.junit.core/src/org/eclipse/jdt/internal/junit/launcher/JUnit4TestFinder.java#L132
https://github.com/eclipse-jdt/eclipse.jdt.ui/blob/master/org.eclipse.jdt.junit.core/src/org/eclipse/jdt/internal/junit/launcher/JUnit5TestFinder.java#L221

jjohnstn:
- fixes #302

## What it does
Does some minor changes to JUnit4TestFinder and JUnit5TestFinder classes to make the finders perform better for very large projects with many anonymous classes.

## How to test
Try running tests in a package that has many classes and many anonymous classes.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.junit.core/src/org/eclipse/jdt/internal/junit/launcher/JUnit4TestFinder.java
+++ b/org.eclipse.jdt.junit.core/src/org/eclipse/jdt/internal/junit/launcher/JUnit4TestFinder.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2006, 2021 IBM Corporation and others.
+ * Copyright (c) 2006, 2022 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -14,8 +14,10 @@
  *******************************************************************************/
 package org.eclipse.jdt.internal.junit.launcher;
 
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Set;
 
 import org.eclipse.core.runtime.CoreException;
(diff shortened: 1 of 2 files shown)


jjohnstn:
Merging after confirmation from original reporter that enhancement brings down start time of tests from ~20s to ~13s.

PR #314, opened 2022-10-28
WIP: Refactor for moving quickfixe logic into JDT

gayanper:
## What it does
Move core components from jdt.ls into org.eclipse.jdt.core.manipulation so that quickfix and refactoring logic in jdt.ls can be moved back into jdt.ui

## How to test
N/A

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.manipulation/common/org/eclipse/jdt/core/corrections/IInvocationContext.java
+++ b/org.eclipse.jdt.core.manipulation/common/org/eclipse/jdt/core/corrections/IInvocationContext.java
@@ -1,0 +1,87 @@
+/*******************************************************************************
+ * Copyright (c) 2000, 2013 IBM Corporation and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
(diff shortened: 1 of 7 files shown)


gayanper:
@rgrunber how does this looks like in relation to what we discussed in https://github.com/eclipse/eclipse.jdt.ls/pull/2257

rgrunber:
It seems to me like this is just copying `getSplitVariableProposals(..)` so that it's here and in https://github.com/eclipse-jdt/eclipse.jdt.ui/blob/f33d12e0bf97384ac97e71df290684814555db5c/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/text/correction/QuickAssistProcessor.java#L2045

Have a look at https://github.com/eclipse-jdt/eclipse.jdt.ui/commit/0a4cf67215674550ccee01d00b33b0ab30a62502 & https://github.com/eclipse-jdt/eclipse.jdt.ui/commit/679803405493442cd649a239bd6834edcb843235 . These moved the logic into jdt.core.manipulation and allowed JDT UI + JDT-LS to consume the same part of the code. See https://github.com/eclipse/eclipse.jdt.ls/commit/716faa3dcd3a5d011686f81b4999bfc6a9621398 for how it looks on the JDT-LS side.

It would be good to reduce the amount of duplicate code in JDT-LS but i'm not sure if just creating more duplicates in jdt.core.manipulation is the right approach.

Issue #331, opened 2022-11-08
Improve the Safety of "Extract Local Variable" Refactorings concering ClassCasts


chixiaye:
# Improve the Safety of "Extract Local Variable" Refactorings concerning ClassCasts

## Description
In previous work (commit ID: [2987007](https://github.com/eclipse-jdt/eclipse.jdt.ui/commit/2987007d9703de02eea83081a0b7b65fdc6e4a84)), we improved the safety of extract variable refactorings in JDT that otherwise may have resulted in NullPointerException. However, as specified in the issue(issue ID: [39](https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/39)), if the extracted expression is a classcast, the extract variable refactoring conducted by JDT may result in NULLPOINTER EXCEPTION as well. This problem has not yet been fixed. 

Considering such example:
```java
1   /* CS1: Original Code Snippet */
2    void foo(Object obj){
3       if(obj instanceof Integer && ((Integer)obj).intValue() > 0){
4             System.out.println(((Integer)obj).intValue());
5       }else if(obj instanceof Float && ((Float)obj).floatValue() > 0.0){
6            System.out.println(((Float)obj).floatValue());
7       }
8    }
``` 

If we select the expression `((Integer)obj).intValue()` (line `4` in `CS1`), and conduct refactoring "extract local variable",and name the new variable as `value`, we will get the following code snippet:
```java
1   /* CS2: Refactored Code Snippet */
2    void foo(Object obj){
3       int value = ((Integer) obj).intValue();
4       if(obj instanceof Integer && value > 0){
5             System.out.println(value);
6       }else if(obj instanceof Float && ((Float)obj).floatValue() > 0.0){
7            System.out.println(((Float)obj).floatValue());
8       }
9    }
``` 

Such refactoring is incorrect and dangerous. If the method is called via `foo(new Float(3.5))`, the refactored version of the method would result in ClassCastException whereas the original version would not.

## Solution
The main difference between NullPointerException case and ClassCastException case is the condition that would cause ClassCastException. We defined the case as sub-condition `obj` instanceof `Class1` and sub-condition `((Class2)obj).method()` occur sequentially in the same logical expression which is between the declaration position and the position where the first expression appears, where `Class2` is the superclass of `Class1`.

The correct and safe refactorings would be as:
```java
1   /* CS3: Safe Refactored Code Snippet */
2    void foo(Object obj){
[...]

PR #333, opened 2022-11-14
Improve the Safety of Extract Local Variable Refactorings concering ClassCasts #331

chixiaye:
**This is a patch based on the solution explained in the bug report.**
The main changed file is `Checker.java`, the description of the modification of this file is as follows:

## 1. Implement Void ClassCasts Exception
Compared with the solution proposed in the original issue, we apply a more generalized solution. First, we traverse the selected expression and create a mapping from expression to type for each `CastExpression`. Second, we validate the code snippet between the variable declaration and the first expression matching the selected expression. If we detect the `InstanceofExpression` and value corresponding to the key value, `leftOperand` of  `InstanceofExpression`, is equal to the `rightOperand` of  `InstanceofExpression` or its supertype, we would assert that such extracting is unsafe.

## 2. Refactoring
Considering this class not only aims at avoiding NullPointerException but also ClassCastExecption, we renamed it to `Check` and renamed the inner class `NullMiddleCodeVisitor` to `MiddleCodeVisitor`. Besides, we removed some private fields of the inner classes to make the code more concise.

### DIFF
--- a/org.eclipse.jdt.ui.tests.refactoring/resources/ExtractTemp/canExtract/A_test126_in.java
+++ b/org.eclipse.jdt.ui.tests.refactoring/resources/ExtractTemp/canExtract/A_test126_in.java
@@ -1,0 +1,12 @@
+package p; //6, 32, 6, 58
+
+class A {
+	void foo(Object obj) {
+		if (obj instanceof Integer && ((Integer) obj).intValue() > 0) {
+			System.out.println(((Integer) obj).intValue());
+		} else if (obj instanceof Float && ((Float) obj).floatValue() > 0.0) {
+			System.out.println(((Float) obj).floatValue());
+		}
+
+	}
+}
--- a/org.eclipse.jdt.ui.tests.refactoring/resources/ExtractTemp/canExtract/A_test126_out.java
+++ b/org.eclipse.jdt.ui.tests.refactoring/resources/ExtractTemp/canExtract/A_test126_out.java
@@ -1,0 +1,13 @@
+package p; //6, 32, 6, 58
+
+class A {
+	void foo(Object obj) {
+		if (obj instanceof Integer && ((Integer) obj).intValue() > 0) {
+			int intValue= ((Integer) obj).intValue();
+			System.out.println(intValue);
+		} else if (obj instanceof Float && ((Float) obj).floatValue() > 0.0) {
+			System.out.println(((Float) obj).floatValue());
+		}
+
+	}
+}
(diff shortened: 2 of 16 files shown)


jjohnstn (changes requested):
Looks good but a few comments to address.

chixiaye:
Hi, as you suggested, we improved the relevant code. I think the check of resolveBinding can be merged into the check that `getSuperClass()` does not return, so I did not add an extra resolveBinding check.

liuhuigmail:
@jjohnstn Please let us known if there is anything deserving further improvement.  Thank you.

jjohnstn:
@liuhuigmail Will look at it today or tomorrow.  I was on vacation last week and will be on vacation from Dec 19th onward.

jjohnstn:
Hi, everything looks good except you haven't fixed hasInheritanceRelationship() to handle interfaces.  For example, you could have:

public interface A {
    public int getInt();
}

public class B implements A {
    public int getInt() {
       return 7;
    }
}

public class C {
    public void foo(Object obj) {
        if ((obj instanceof A) && ((A)obj).getInt() > 3) {
           ....
        }
   }
}

If an instance of B is passed to C.foo() and the user asks to put the 2nd part of the if statement in a local variable,
your logic for hasInheritanceRelationship() won't catch it because getSuperClass() only finds those types specified
in an "extends" clause, but not an "implements" clause so you will never find a match with interface A.  There is a separate call for getting interfaces getInterfaces() that needs to be called in addition to getSuperClass() and you should make the method recursive.  Essentially you can check the passed in class and all of its interfaces, then recursively call for each of the interfaces and with the super class if it exists.   This will ensure you check the whole inheritance tree.  Each recursive call can check if you get a match and stop the recursion.

jjohnstn:
@chixiaye BTW: I added a N&N entry for your previous null check fix in 2022-12 and it made the release video:
https://youtu.be/jJau4kUoLrA

liuhuigmail:
@jjohnstn  Thank you for the comments. We will fix it quickly and resubmit the patch.  
By the way, we have submitted another issue concering the Extract Local Varialbe refactoring (https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/348). Would you please kindly have a look and let us known whether it is ok for us to prepare a patch (pull request)  as we specified in the issue report.  Thanks.

chixiaye:
Hi, the method `hasInheritanceRelationship` has been improved to identify the inheritance concerning the interface. And the implementation of `hasInheritanceRelationship` has also been changed to a recursive form. Besides, Test-128 and Test-129 were added to verify the inheritance relationship between classes and interfaces. Test-130 and Test-131 were added to test the inheritance relationship between interfaces.

jjohnstn (approved):

jjohnstn (approved):
Looks good.  Just need to get build working.

jjohnstn:
Hi @chixiaye, please look at the results of the latest build.  There is a warning and the build is failing due to -failonwarning being specified.  It is complaining about using int as a key for a Map.

PR #335, opened 2022-11-18
Move the logic of MissingAnnotationAttributes to o.e.j.core.manipulation

datho7561:
## What it does

Moves all the text editing logic of the "add missing attributes to annotation" QuickFix to org.eclipse.jdt.core.manipulation so that it can be reused in jdt.ls.

## How to test

Test that `MissingAnnotationAttributesFixCore` can be accessed from another project without needing to require any ui packages.

## Author checklist

Signed-off-by: David Thompson <davthomp@redhat.com>

### DIFF
--- a/org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/fix/MissingAnnotationAttributesFixCore.java
+++ b/org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/fix/MissingAnnotationAttributesFixCore.java
@@ -1,0 +1,49 @@
+/*******************************************************************************
+ * Copyright (c) 2022 Red Hat Inc. and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Red Hat Inc. - initial API and implementation
+ *******************************************************************************/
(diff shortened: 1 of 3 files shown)


rgrunber:
`LocalCorrectionsQuickFixTest#testMissingAnnotationAttributes*` seem to be passing. I also like that this requires no changes to the call sites that continue to make the call to the jdt.ui layer.

org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/fix/MissingAnnotationAttributesFixCore.java
```diff
@@ -0,0 +1,49 @@
...
+	 * location.
+	 *
+	 * @param cu the AST of the compilation unit where an attribute is missing required attributes
+	 * @param problemLocation the error marker that marks the annotation that's missing attributes
+	 * @return a fix that adds the missing required attributes to the annotation at the problem
+	 *         location
+	 */
+	public static MissingAnnotationAttributesFixCore addMissingAnnotationAttributesProposal(CompilationUnit cu, IProblemLocationCore problemLocation) {
+		ASTNode selectedNode= problemLocation.getCoveringNode(cu);
+		Assert.isTrue(selectedNode instanceof Annotation);
```
rgrunber, line 42: Do we need this (when we eventually implement it in a language server, or some core component) ? Could we just return `null` and check for before creating the proposal.

datho7561: Yeah, that makes sense to me. I'll update it.

PR #340, opened 2022-12-03
Add support for inline return tags

coehlrich:
## What it does

This adds support for inline return tags which were introduced in java 16.
Inline return tags along with being added with the normal tags like the normal return tag they also get inserted into the main text where the inline return tag is but prepended with "Returns " when using English.
The spec for the inline return tag is: https://docs.oracle.com/en/java/javase/19/docs/specs/javadoc/doc-comment-spec.html#return

## How to test

Add a Javadoc to a method containing an inline return tag and either hover over the method or open the Javadoc view.

## Author checklist


akurtakov:
@jjohnstn would you please review this one.

jjohnstn:
@akurtakov Ok

jjohnstn (approved):
Looks good.  The inline tag is supposed to be the first thing in description but I tested and found that Javadoc just warns against something like XYZ {@return value} and processes it to be:
XYZ Returns value.

PR #344, opened 2022-12-12
Move "Extract Interface" logic to o.e.j.core.manipulation

CsCherrYY:

## What it does

Move extract interface refactoring logic to org.eclipse.jdt.core.manipulation so that it can be reused in jdt.ls.

## How to test

use existing test cases in jdt.ui.
## Author checklist


carstenartur:
How is a third party bundle supposed to have access to the migrated classes in both bundles involved (source and target) when the migrated classes still are in the same java package after relocation? Are all migrated classes really an internal implementation detail that a third party bundle never accesses directly? The osgi bundle access configuration is in trouble when you have to import the same java package from two different bundles, isn't it?

CsCherrYY:
> How is a third party bundle supposed to have access to the migrated classes in both bundles involved (source and target) when the migrated classes still are in the same java package after relocation?

AFAIK, the classes under package o.e.j.core.manipulation will be packaged into a standalone file org.eclipse.jdt.core.manipulation_*.jar, so that it can be used in other bundles.

> Are all migrated classes really an internal implementation detail that a third party bundle never accesses directly?

In this case, yes. All we need is to reuse the type `ExtractInterfaceProcessor`, and use it to create a **Change** with given parameters in jdt.ls. Other files are necessary to that type.

> The osgi bundle access configuration is in trouble when you have to import the same java package from two different bundles, isn't it?

Now the bundle access configuration of jdt.ls is configured to access the types in o.e.j.core.manupution. See: https://github.com/eclipse/eclipse.jdt.ls/blob/3400ec33d2b05e6ee639d90e2d3e1ed509c4b495/org.eclipse.jdt.ls.core/META-INF/MANIFEST.MF#L25

carstenartur:
That is not where I see a problem. I wonder if the amount of unresolvable warnings will increase for bundles depending on both changed bundles.
Something like ewe see in

https://github.com/carstenartur/sandbox

rgrunber:
If the exports exist in both package, it should be possible declare the one in jdt.core.manipulation as split so that clients can use `Require-Bundle` if they need both jdt.ui and jdt.core.manipulation. There are already such packages existing, so if we intend to reduce such splits, we should move them over entirely.

This is basically what is being done for `org.eclipse.jdt.internal.corext.refactoring.structure.constraints`.

org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/refactoring/structure/constraints/SuperTypeRefactoringProcessor.java
```diff
@@ -433,12 +431,12 @@ protected final String createTypeSource(final ICompilationUnit copy, final IType
 				source= buffer.toString();
 			}
 			final IDocument document= new Document(source);
-			final TextEdit edit= CodeFormatterUtil.format2(CodeFormatter.K_COMPILATION_UNIT, source, 0, delimiter, FormatterProfileManager.getProjectSettings(copy.getJavaProject()));
+			final TextEdit edit= CodeFormatterUtil.format2(CodeFormatter.K_COMPILATION_UNIT, source, 0, delimiter, copy.getJavaProject().getOptions(true));
```
jjohnstn, line 434: This change omits some logic that is found in FormatterProfileManager.getProjectSettings() which looks at the preferences.  Admittedly, there is logic there that uses JDT UI classes but some way around it should be found.

jjohnstn (changes requested):

org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/refactoring/structure/constraints/SuperTypeRefactoringProcessor.java
rgrunber: Yeah, we need to take a closer look here. I think we might have migrated some classes in the past and we probably had to deal with this, so we just need to determine what we did, or what the best approach is to support the behaviour.

CsCherrYY: I just move the implementation of method `getProjectSettings()` to a new type FormatterProfileManagerCore via https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/344/commits/03bfc3ea777b8248b811f32fa4df68364d837aec, since moving whole `FormatterProfileManager` will cause a lot of other changes. Now we can reach the logics there.

org.eclipse.jdt.core.manipulation/common/org/eclipse/jdt/internal/ui/preferences/formatter/FormatterProfileManagerCore.java
```diff
@@ -0,0 +1,40 @@
...
+public class FormatterProfileManagerCore {
+	public final static String FORMATTER_SETTINGS_VERSION= "formatter_settings_version"; //$NON-NLS-1$
+
+	public static Map<String, String> getProjectSettings(IJavaProject javaProject) {
+		Map<String, String> options= new HashMap<>(javaProject.getOptions(true));
+		IEclipsePreferences prefs= new ProjectScope(javaProject.getProject()).getNode(JavaManipulation.getPreferenceNodeId());
+		if (prefs == null)
+			return options;
+		int profileVersion= prefs.getInt(FORMATTER_SETTINGS_VERSION, ProfileVersionerCore.getCurrentVersion());
+		if (profileVersion == ProfileVersionerCore.getCurrentVersion())
```
CsCherrYY, line 36: Here we can remove the usage of `CustomProfile` and directly use `ProfileVersionerCore.updateAndComplete()`.

org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/refactoring/structure/constraints/SuperTypeRefactoringProcessor.java
jjohnstn: Looks good.

jjohnstn (approved):

CsCherrYY:
thank you @jjohnstn !

PR #352, opened 2022-12-16
Make quickfix that suggests importing a class smarter

### LINKED ISSUE #349: Take returntype of method into account for suggestions
I already noticed this several times and it is a constant source of "what the ... " moments of eclipse:

I have a local variable and it is missing the type import returned from a method, then I open the QuickFix and it get multiple choices where obviously only one can be the right choice (that is the return type of the method) see the following screenshot:

grafik

even worse, not only I now have the choice, the first option in this case is also the wrong one, so if I choose this "fix" I end up with an error telling me that the types are not assignable.

jjohnstn:
- change UnresolvedElementsSubProcessor.addSimilarTypeProposals() to check if an interface is expected in which case make sure that each suggested class to import has the interface in its hierarchy
- fixes #349

## What it does

Fixes the quickfix for an undefined type that is used to initialize or return an interface so that any suggested import class has the given interface in its hierarchy.

## How to test

Create the following:

```
package tests;

public class BundleInfo {
	
	public int getInt() {
		return 43;
	}

}

package tests2;

public interface ISomePath {
	
	public String getPath();

}

package tests2;

public class BundleInfo implements ISomePath {

	@Override
	public String getPath() {
		return null;
	}

}

package tests3;

import tests2.ISomePath;

public class TestBundleInfo {
	
	public ISomePath getSomePath() {
		return new BundleInfo();
	}

}
```

Hover over the missing BundleInfo() creation in TestBundleInfo().  The quickfix should only suggest the version of BundleInfo that implements ISomePath.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/text/correction/UnresolvedElementsSubProcessor.java
+++ b/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/text/correction/UnresolvedElementsSubProcessor.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2021 IBM Corporation and others.
+ * Copyright (c) 2000, 2022 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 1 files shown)


jjohnstn:
Test failure is known issue in jdt.core that is being addressed.

laeubi:
I can't wait until this is released :+1:

PR #355, opened 2022-12-19
Support specifying package for extract interface refactor

CsCherrYY:

cc @rgrunber 

## What it does

Add a new method `ExtractInterfaceProcessor.setPackageFragment()` to set the destination package fragment of extract interface refactor. It can be used by jdt.ls.

## How to test

see the added test case `testInterface7`.

## Author checklist


org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/refactoring/structure/ExtractInterfaceProcessor.java
```diff
@@ -1043,6 +1048,14 @@ protected void rewriteTypeOccurrences(final TextEditBasedChangeManager manager,
 					JavaManipulationPlugin.log(exception);
 				}
 				subUnit.getBuffer().setContents(document.get());
+				IPackageFragment fragment = this.getPackageFragment();
+				if (fragment != null && !fragment.equals(this.fSubType.getPackageFragment())) {
```
CsCherrYY, line 1052: if the destination package is not the same as the subtype's package, we should manaully add a import statement about the new interface so that bindings can be resolved correctly.

org.eclipse.jdt.ui.tests.refactoring/test cases/org/eclipse/jdt/ui/tests/refactoring/ExtractInterfaceTests.java
```diff
@@ -135,6 +135,40 @@ protected void validatePassingTest(String className, String[] cuNames, String ne
 		assertEqualLines("(interface cu)", getFileContents(getOutputTestFileName(newInterfaceName)), interfaceCu.getSource());
 	}
 
+	protected void validatePassingTest(String className, String[] cuNames, String newInterfaceName, boolean replaceOccurrences, String[] extractedMethodNames, String[][] extractedSignatures, String[] extractedFieldNames, IPackageFragment newPackageFragment) throws Exception {
```
datho7561, line 138: I think it would be a good idea to range format this method, since there is some odd formatting (mixed tabs/spaces, long lines).

org.eclipse.jdt.ui.tests.refactoring/test cases/org/eclipse/jdt/ui/tests/refactoring/ExtractInterfaceTests.java
```diff
@@ -135,6 +135,40 @@ protected void validatePassingTest(String className, String[] cuNames, String ne
...
+				cus[i]= clas.getCompilationUnit();
+			else
+				cus[i]= createCUfromTestFile(clas.getPackageFragment(), cuNames[i]);
+		}
+		processor.setReplace(replaceOccurrences);
+		processor.setAnnotations(fGenerateAnnotations);
+		IMethod[] extractedMethods= getMethods(clas, extractedMethodNames, extractedSignatures);
+	    IField[] extractedFields= getFields(clas, extractedFieldNames);
+		processor.setExtractedMembers(merge(extractedMethods, extractedFields));
+		processor.setPackageFragment(newPackageFragment);
```
datho7561, line 159: Maybe it's worthwhile to check if the passed `packageFragment` is null, and if so, don't set the processor's package fragment. Then, the other `validatePassingTest` could just call this one with `newPackageFragment = null`, so there is less repeated code.

I don't know the conventions of this code base well, so maybe this is frowned upon in this code base.

datho7561 (changes requested):

org.eclipse.jdt.ui.tests.refactoring/test cases/org/eclipse/jdt/ui/tests/refactoring/ExtractInterfaceTests.java
CsCherrYY: currently when the field `fFragment` is null, we will fall back to the subtype's fragment, see https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/355/files#diff-4ff695c2e81ce4025c310ebb45028b14ed249febccde313de659114d6c8130d8R1128.

do you mean 

- we initialize the `fFragment` with the subtype's fragment
- check if the passed `packageFragment` is null. If not, set the `fFragment`
- when executing, we directly use `fFragment`, and there is no need for the current fall back

datho7561: I just meant for this helper method in the tests, instead of having a copy paste of an existing method with an additional parameter, we have the old method (that was copied) call the the new method that you introduced, but pass `null` as the `packageFragment`. Then, only call `processor.setPackageFragment(packageFragment)` if the `packageFragment` that was supplied is non-null. This way there is less code duplication.

CsCherrYY: resolved in https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/355/commits/ffd3b23edae1675249ed33f51128e79a79c9a322 :)

CsCherrYY:
@datho7561 I have fixed the failed test cases, could you please take a look?

rgrunber (approved):
Changes works well for me. I tried just manually using the API as well and seems like it shouldn't be too difficult to use it with the corresponding package explorer widget in the UI, if someone were interested in adding the feature in JDT.

Issue #348, opened 2022-12-13
Improve the Safety of "Extract Local Variable" Refactorings by Identifying the Side Effect of Selected Expression


chixiaye:
# Improve the Safety of "Extract Local Variable" Refactorings by Identifying the Side Effect of Selected Expression

# Description
"Extract Variable" refactoring in Eclipse provides an option to replace all occurrences of the selected expression with references to the newly extracted local variable. However, it may result in semantic errors when the selected expression (to be extracted as a variable) has side effect, i.e., it may change the states of the system. 

 A simple and straightforward example is `Lines[i++]`: 

```java
1   /* CS1: Original Code Snippet */
2   String[] Lines;
3   int i;
4   private String foo(String msg) {
5   	return msg+": "+ Lines[i++];
6   }
7   ...
8   log.warn(foo("Format warning"));
9   ...
10  log.warn(foo("Format warning"));
```
Extracting `foo("Format warning")` as local variable (enabling "replace all occurrences") would result in the following code:

```java
1   /* CS2: Code Snippet After Refactoring */
2   String[] Lines;
3   int i;
4   private String foo(String msg) {
5   	return msg+": "+ Lines[i++];
6   }
7   ...
8   String foo= foo("Format warning");
9   log.warn(foo);
10  ...
11  log.warn(foo);
```
The resulting code is semantically different from the original code, which results in semantic errors. 

# Solution
We should check the potential side effect of the selected expression before the actual conduction of the Extract Local Variable refactoring. 

The key of the checking is composed of two parts. The first is to identify all statements that may be executed by the selected expression, which may include the expression itself and all statements directly or indirectly invoked by the expression. To guarantee the performance of the IDE, we would confine the method call stack to the first four levels (if there are any).  

The second part is to validate whether the collected statements would change any status of the systems, i.e., updating variables/parameters/fields or generating outputs.   Notably, if we cannot access the body of the method invoked by the retrieved statements, we simply assume the method is safe (having no side effect) to minimize the chance of false alarms. 

To guarantee the performance (response time) of the refactoring engine, we can set a maximum analyzing time (e.g., 3 or 5 seconds). If the analysis terminates within the given time slot, that is fine. If fail, it would return what the current version of Eclipse returns (i.e., allowing the extracting of the selected expression).

**The above is our proposed solution, if you have any comments or suggestions, please let us know.**

chixiaye:
[...]

PR #360, opened 2022-12-26
Improve the Safety of "Extract Local Variable" Refactorings by Identifying the Side Effect of Selected Expression #348

chixiaye:
Here is a PR based on the discussion in the [issue-348](https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/348). 

If  "replace all occurrences" option is enabled and more than one expression is extracted, `SideEffectChecker` would be created to check the side effects of the selected expression. A warning message will be thrown to remind the user if it identifies the potential side effects. 

Besides, several test cases have been appended to validate the correctness concerning side effects detection. Test-132 is about the side effects direct occurring in the selected expression. Test-133 and  Test-134 were designed to ensure that only when the called method's implementation is completely sure, a warning message would be thrown. Test-135 is a test for the assignment that the left value depends on itself. And test-136 was added to test the static variable situation.

Finally, there are two points that need further discussion: 
- Although only the expression and its directly called method are taken into consideration, the method implementation still has the possibility to cause huge computational costs. I suggest adding a metric to confine, but I am not sure which metric is suitable(e.g. the number of lines in the method body). 
- In some cases, side effects can be caused by modifying parameter values. Considering the following code, obviously, the expression `m(arr, index)` may have a side effect, but `m(new int[index+1], index)` doesn't. A more complex example is about `m(arr == null? new int[index+1]:arr, Math.abs(index-1))`. I originally included the code for detecting parameter value modification. But considering the complexity of the actual situation and possible false positives, I commented them out in the current commit. If this situation also needs attention, I will add it in subsequent revisions.

```java
    int m(int[] array,int index) {
    	return array[index]--;
    }
```


org.eclipse.jdt.ui/core refactoring/org/eclipse/jdt/internal/corext/refactoring/util/SideEffectChecker.java
```diff
@@ -0,0 +1,324 @@
...
+							continue;
+						if (res != null)
+							return null;
+						res= md;
+					}
+				}
+			}
+			return res;
+		} catch (JavaModelException e) {
+			e.printStackTrace();
```
jjohnstn, line 155: This might be better logged if it is unexpected.

org.eclipse.jdt.ui/core refactoring/org/eclipse/jdt/internal/corext/refactoring/util/SideEffectChecker.java
```diff
@@ -0,0 +1,324 @@
...
+	public SideEffectChecker(ASTNode astNode) {
+		fExpression= astNode;
+		fSideEffect= false;
+	}
+
+	boolean fSideEffect;
+
+	ASTNode fExpression;
+
+	private final int THRESHOld= 1000;
```
jjohnstn, line 64: Not sure about this threshold as 80 char lines, it would be 12 lines total.  I do agree in a threshold due to the cost of parsing the code, but perhaps a little higher.

org.eclipse.jdt.ui/core refactoring/org/eclipse/jdt/internal/corext/refactoring/util/SideEffectChecker.java
```diff
@@ -0,0 +1,324 @@
...
+			return res;
+		} catch (JavaModelException e) {
+			e.printStackTrace();
+		}
+		return null;
+	}
+
+	private static void findTypes(IType it, ITypeHierarchy ith, ArrayList<IType> iTypes) throws JavaModelException {
+		iTypes.add(it);
+		for (IType i : ith.getAllSubtypes(it)) {
```
jjohnstn, line 162: This can be quite expensive if someone is using a basic Java method.  Have you tried something like a call to an Object method like hashCode()?  You might need to put this in a thread to be able to have it cancellable.  Another possibility is to screen out Java library classes such as java.lang, java.util, java.io, etc... and don't do a type hierarchy in such a case.

org.eclipse.jdt.ui/core refactoring/org/eclipse/jdt/internal/corext/refactoring/util/SideEffectChecker.java
```diff
@@ -0,0 +1,324 @@
...
+				if (qn.resolveBinding() instanceof IVariableBinding) {
+					ivb= (IVariableBinding) qn.resolveBinding();
+				}
+				return ivb != null && Modifier.isStatic(ivb.getModifiers())
+						|| isNoTemp(qn.getQualifier());
+			} else if (expr instanceof ArrayAccess) {
+				ArrayAccess aa= (ArrayAccess) expr;
+				return isNoTemp(aa.getArray());
+			}
+//			if (ivb != null && ivb.isParameter()) {
```
jjohnstn, line 295: This should be removed if no longer needed.

org.eclipse.jdt.ui.tests.refactoring/test cases/org/eclipse/jdt/ui/tests/refactoring/ExtractTempTests.java
```diff
@@ -876,6 +876,43 @@ public void test125() throws Exception {
 		helper1(9, 32, 9, 42, true, false, "length", "length");
 	}
 
+
```
jjohnstn, line 879: You should add a test that has a method with parameter(s) and have the parameter expression be a prefix expression so it cannot be done in multiple.  You should also add another test which is the same code but doesn't call with a prefix expression so that all calls are replaced.

org.eclipse.jdt.ui/core refactoring/org/eclipse/jdt/internal/corext/refactoring/util/SideEffectChecker.java
```diff
@@ -0,0 +1,324 @@
...
+			return false;
+		}
+		if (node instanceof MethodInvocation) {
+			MethodInvocation node2= (MethodInvocation) node;
+			final IMethodBinding resolveMethodBinding= node2.resolveMethodBinding();
+			if (MARKEDMETHODSET.contains(getQualifiedName(resolveMethodBinding))) {
+				fSideEffect= true;
+				return false;
+			}
+			MethodDeclaration md= findFunctionDefinition(resolveMethodBinding.getDeclaringClass(), resolveMethodBinding);
```
jjohnstn, line 95: I think if you cannot parse a function and it isn't part of a set you have pre-cleared  you should assume side-effects and only change one instance.  This is safe and you are at least changing one instance.

jjohnstn (changes requested):
Looks quite good and works good for the samples I tried, but there are some performance issues looming if someone tries to extract a basic method call such as Object.hashCode().  I think that if your method parsing fails or can't be run for any reason you need to assume side-effects and don't do multiple changes.  That will always be safe.

chixiaye:
Hi @jjohnstn, I have submitted the new changes. The following changes were made:

1.Updated threshold. The threshold changed to 2500.

2.Fix the performance issue concerning the finding method. The following code was added to avoid finding source code in java library classes.

```java
IJavaElement root= it.getAncestor(IJavaElement.PACKAGE_FRAGMENT_ROOT);
if (root instanceof IPackageFragmentRoot) {
    IClasspathEntry cpEntry= ((IPackageFragmentRoot) root).getRawClasspathEntry();
    if (cpEntry.getEntryKind() == IClasspathEntry.CPE_CONTAINER
            && cpEntry.getPath().toString().startsWith("org.eclipse.jdt.launching.JRE_CONTAINER")) { //$NON-NLS-1$
        return null;
    }
}
```

3.Two test cases are supplemented. Test-138 and Test-139 were added to validate the correctness of subexpressions with side effects that occur as parameters in extracted expression.

4.Other code details adjustments. 

I am sorry that I am a little bit confused about the action when the function can not be parsed. As we discussed before, the code `System.out.println(list.size()+","+list.size());` would be refactored as `int v= list.size(); System.out.println(v+","+v);`. `list.size()` is the method that we cannot find its implementation, but we can do multiple changes. Such a clear method is not only in Java library classes but also in third-party libraries. In my personal view, for these methods, it is better to keep the same as the original JDT version does, that is, allowing to extract all occurrences.

jjohnstn:
For item 4, I am referring to a method from user-code that you cannot get the source (e.g. binding failures, null pointers returned).  If you have decided to clear Java library code except for marked methods, that is fine, but for user code you should be able to get access to the source and if due to error, it is a reasonable precaution to treat it as a side-effect (e.g. you have a routine that is self-modifying, but another compilation error causes the binding to come back null).

chixiaye:
@jjohnstn Thanks so much for your patient explanation. I have added the relevant code. If the method cannot be parsed, it is also considered to have side effects.

org.eclipse.jdt.ui/core refactoring/org/eclipse/jdt/internal/corext/refactoring/util/SideEffectChecker.java
jjohnstn: With the check for JRE prior, this is fine now.

org.eclipse.jdt.ui/core refactoring/org/eclipse/jdt/internal/corext/refactoring/util/SideEffectChecker.java
```diff
@@ -0,0 +1,332 @@
...
+						MethodDeclaration md= (MethodDeclaration) perform;
+						if (Modifier.isAbstract(md.resolveBinding().getModifiers()))
+							continue;
+						res= md;
+					} else {
+						fSideEffect= true;
+						return null;
+					}
+				} else if (tmp != null && res != null) {
+					return null;
```
jjohnstn, line 166: What scenario does this represent where you don't set sideEffect to true?

jjohnstn (changes requested):
Hi @chixiaye looks better.  There was one scenario where I didn't know why you didn't set sideEffect to true but returned null.  I'll let you comment or modify the code.  You should also rebase on top of master and I think we're done.

org.eclipse.jdt.ui/core refactoring/org/eclipse/jdt/internal/corext/refactoring/util/SideEffectChecker.java
chixiaye: It represents the scenario for multiple possible function implementations due to run-time polymorphism in Java. A typical example is:
```java
Person p = new Man();   
p.toString();
class Person{
    public String toString() {
        String name = "Person";
        return name;
    }
}
 
class Man extends Person{
    public String toString(){
        String name = "Man";
        return name;
    }
}
```
`p.toString()` executes `toString()` in the `Man` class rather than `Person` class. Considering this uncertainty, we adopt a conservative strategy, which is considered to have no side effects. In addition, there are some other strategies that can be used:
1. Scan for all potential methods. This strategy is the most accurate but significantly increases the computational cost.
2. Only choose one of the methods for detection. The overriding methods have similar effects that guarantee a certain accuracy.
3. Marked as having side effects. It can guarantee enough security, but it may cause a bad user experience because of the common use of polymorphism.

jjohnstn: I don't think that an overridden method should be assumed to be side-effect free which is what you are doing by returning null.  I think option 2 is a reasonable assumption but you should check for an abstract method and continue on until you find a real implementation.

org.eclipse.jdt.ui/core refactoring/org/eclipse/jdt/internal/corext/refactoring/util/SideEffectChecker.java
jjohnstn: Ok now.

jjohnstn (approved):

jjohnstn:
@chixiaye  Looks good.  Thanks for handling all the comments.

liuhuigmail:
@jjohnstn  Thank you for the help. Concering the safety of extract local variable refacotoring,  we have submitted the last issue at https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/432  

Would you please kindly have a look and let us knonw whether you are fine with the proposed solution?

PR #362, opened 2023-01-03
Show option to create a new java project in the project explorer

laeubi:
Currently if one opens the "Package Explorer" and rightclick an item the user is offered to create a new java project. On the other hand, in the "Project Explorer" this option is not offered what is confusing.

This adds a commonWizzard with a filter for the active perspective that shows this option in the "Project Explorer" as well.

### DIFF
--- a/org.eclipse.jdt.ui/plugin.xml
+++ b/org.eclipse.jdt.ui/plugin.xml
@@ -6351,6 +6351,16 @@
 					</or>
 				</enablement>
 			</commonWizard>
+			<commonWizard
+	           menuGroupId="org.eclipse.jdt.ui.java"
+	           type="new"
+	           wizardId="org.eclipse.jdt.ui.wizards.JavaProjectWizard">
+		       <enablement>
+					<with variable="activeWorkbenchWindow.activePerspective">
+   						<equals value="org.eclipse.jdt.ui.JavaPerspective"/>
+   					</with>
+				</enablement>
+		    </commonWizard>
   			<dropAssistant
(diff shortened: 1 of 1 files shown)


jjohnstn:
Hi @laeubi I am trying to test the patch but it is not working for me.  When I right-click on a project, for example, in the Project Explorer and I have the Java perspective selected, I am only seeing Project... in the New menu list, not Java Project.  I am using a child Eclipse off of current github sources and I can see your changes in the org.eclipse.jdt.ui/plugin.xml are there.  Is it working for you?

laeubi:
I think for my local tests I have to reset the perspective once to get it working (no idea why):

- open java perspective
- reset the perspective
- open the project explorer view
- right click on a project

jjohnstn:
Hi @laeubi I am able to get it working now.  It would be preferable if you can get the Java Project... to show up before the Project... as it does in the Package Explorer (i.e. the first option), otherwise it is buried in the Java items.  I tried an experiment setting the menu group id to be org.eclipse.a as an experiment and was able to get it to appear before the other java items, but below the File and Folder options.  I also tried empty string but that was an error.  Perhaps the extension needs to be altered.

laeubi:
The position is nothing jdt can really influence but I have already opened an issue/pr here to fix that in platform:
- https://github.com/eclipse-platform/eclipse.platform.ui/pull/506

laeubi:
The fix should show up in the next ibuild, and then project-wizards will be shown in the "New Project.." category.

jjohnstn (approved):
Looks good after the platform change.

PR #375, opened 2023-01-09
Move tryWithResourceCleanUp into jdt.core.manipulation

JessicaJHee:
## What it does
Moves the "try-with-resource" clean up logic to org.eclipse.jdt.core.manipulation so that it can be used in jdt.ls as well.

## How to test
Confirm that TryWithResourceFixCore can be accessed without needing to import any jdt.ui packages.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/fix/TryWithResourceFixCore.java
+++ b/org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/fix/TryWithResourceFixCore.java
@@ -1,0 +1,289 @@
+/*******************************************************************************
+ * Copyright (c) 2023 Red Hat Inc. and others.
+ *
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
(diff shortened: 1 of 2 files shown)


rgrunber:
The change looks pretty good to me. You added the source range computation logic that deals with `UNTOUCH_COMMENT`, and that was the main thing that needed to be included in addition to moving the class. I think I'll just take an example where every other token is a comment (eg. `try /* comment 1 */ { /* comment 2 */ ...`), and confirm the refactoring behaves exactly the same as before.

Also just wondering for the future : I know there's a lot of instances where the refactorings being moved into jdt.core.manipulation have simply copied over the functionality mentioned above, but would it also have been possible to somehow copy from https://github.com/eclipse-jdt/eclipse.jdt.ui/blob/07373996f857569bb55af61381d3e2f074c7b79c/org.eclipse.jdt.ui/core%20extension/org/eclipse/jdt/internal/corext/fix/CompilationUnitRewriteOperationsFix.java#L33-L47 into https://github.com/eclipse-jdt/eclipse.jdt.ui/blob/07373996f857569bb55af61381d3e2f074c7b79c/org.eclipse.jdt.core.manipulation/core%20extension/org/eclipse/jdt/internal/corext/fix/CompilationUnitRewriteOperationsFixCore.java#L46 ? Just wondering if it's possible to have everyone inherit that as opposed to having to remember to copy it at every migration.

rgrunber:
Seems to be working well for me. The unit tests dealing with try-with also work locally for me as expected.

PR #397, opened 2023-01-17
Add new preference to ignore inexact refs for Code Minings

### LINKED ISSUE #370: Code Minings - Method call occurences for common named methods are not correctly calculated
# Description #
----------

Hello there, 

I noticed Code Minings is not calculating correct occurences of, at least, methods. 
This has likely something to do with the way occurences are found.
I don't know when it started working like this.

## Expected Outcome ##

I expect that an occurence of a method call is calculated in the same way as **"Open Call Hierarchy"** option or at least that shows the call occurences in the workspace.

## Current Outcome ##

Some methods, depending on their name and/or signature will be marked has having `"X occurences"` but these don't match any actual call of that method, but rather any other method with the same name that can be somehow found in the classpath.

## Steps to reproduce ## 

I have prepared a sample application that has a very simple Payara Micro application, whose kind is irrelevant to the case in point.
It can be found [here](https://github.com/CarlosMOGoncalves/eclipse-issue-tester)

This app does nothing special. It has a Jakarta REST endpoint but that is just for show.

1. Fresh install Eclipse for **Eclipse IDE for Enterprise Java and Web Developers 2022-12**
2. Checkout the project from my repository
3. Run `mvn clean install` on it
4. Activate Code Minings for method references, like this:

image

5. Then navigate to the class `NumericUtils.java`, method `getType()`:

image

6. It marks 1 occurence. However, that occurence is not correct, but rather just an occurence a similarly named method on a comment:

image

7. If we use "Open Call Hierarchy" on that method we clearly see that no one is calling it:

image

8. The same can be said for methods with common names, like on `TestResource#get()`:

image

9. Which is not in fact the right callee:

image

**Final note**

- sorry for all the Issue ceremony. I used a template from Payara issues as I think it is pleasant to work with.

## Environment ##
- **Distribution**:  Eclipse IDE for Enterprise Java and Web Developers 2022-12 2022-12 for Windows x64
- **JDK Version**: Corretto JDK 11
- **Operating System**:  Windows 10

jjohnstn:
- add a new preference to ignore inexact results for Code Mining references and have this used instead of the Search preference
- fixes #370

## What it does

Adds a Code Mining preference to ignore inexact references which will be used instead of reading the similar preference
in the Search preferences which is not intuitive to the end-user.

## How to test

Perform Code Mining and specify to show method references.  The new option will match the results of setting the Search preference regarding ignore potential matches.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/javaeditor/codemining/JavaReferenceCodeMining.java
+++ b/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/javaeditor/codemining/JavaReferenceCodeMining.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2018, 2020 Angelo Zerr and others.
+ * Copyright (c) 2018, 2023 Angelo Zerr and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -36,8 +36,6 @@ import org.eclipse.ui.IEditorPart;
 
 import org.eclipse.ui.texteditor.ITextEditor;
 
-import org.eclipse.search.ui.NewSearchUI;
-
 import org.eclipse.jdt.core.IJavaElement;
 import org.eclipse.jdt.core.IJavaProject;
 import org.eclipse.jdt.core.JavaCore;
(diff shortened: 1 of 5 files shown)


CarlosMOGoncalves:
Thanks for this feature @jjohnstn

PR #406, opened 2023-01-21
Improve names of new views

Bananeweizen:
* Never use the _type_ of a UI element in its name (view, wizard, dialog, ...).
* Also use a blank for "Java Element", even though that is related to the JavaElement class. People not knowing that class will assume a typo otherwise.

grafik

The new names will still work with Ctrl-3 and typing the old name, so committers don't have to re-learn muscle memory.

### DIFF
--- a/org.eclipse.jdt.astview/plugin.xml
+++ b/org.eclipse.jdt.astview/plugin.xml
@@ -5,7 +5,7 @@
      <extension
          point="org.eclipse.ui.views">
       <view
-            name="ASTView"
+            name="Abstract Syntax Tree"
             icon="$nl$/icons/view.png"
             category="org.eclipse.jdt.ui.java"
             class="org.eclipse.jdt.astview.views.ASTView"
--- a/org.eclipse.jdt.jeview/plugin.properties
+++ b/org.eclipse.jdt.jeview/plugin.properties
@@ -1,15 +1,15 @@
 ###############################################################################
 # Copyright (c) 2000, 2005 IBM Corporation and others.
 #
-# This program and the accompanying materials 
+# This program and the accompanying materials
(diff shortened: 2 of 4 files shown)


iloveeclipse:
See https://github.com/eclipse-tycho/tycho/pull/2013#issuecomment-1399243667

I assume either it is a problem after recent SDK build failures, or something directly from tycho changes.

Bananeweizen:
Since there is a NoSuchMethodError both in this PR and on platform.ui master, I suspect the Tycho snapshot jars used in the build are not all from the same build or the Tycho publishing somehow broke in the middle, leaving the latest jar in an inconsistent state. The last changes in the class mentioned in the error message were like 5 hours ago. Somehow forcing the Maven snapshots to be updated in the build should then fix it, but I have no idea how to try that.

iloveeclipse:
> Since there is a NoSuchMethodError both in this PR and on platform.ui master, I suspect the Tycho snapshot jars used in the build are not all from the same build or the Tycho publishing somehow broke in the middle, leaving the latest jar in an inconsistent state. The last changes in the class mentioned in the error message were like 5 hours ago. Somehow forcing the Maven snapshots to be updated in the build should then fix it, but I have no idea how to try that.

Could you please report a bug in tycho please? I haven't noticed the NoSuchMethodError.

iloveeclipse:
> Could you please report a bug in tycho please?

I've reported https://github.com/eclipse-tycho/tycho/issues/2014

laeubi:
All checks seem to pass now, so is the problem fixed with that?

iloveeclipse:
> All checks seem to pass now, so is the problem fixed with that?

We will see after infra starts to work again. If you see official SDK build working again, feel free to assume the problem is fixed.

PR #407, opened 2023-01-22
Add option to update snapshots and print a stacktrace on failure

laeubi:
@iloveeclipse if you don't mind, I would recommend to add these options permanently to the Jenkisn build to be able to better identify issues in the build.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -17,7 +17,7 @@ pipeline {
 			steps {
 				wrap([$class: 'Xvnc', useXauthority: true]) {
 					sh """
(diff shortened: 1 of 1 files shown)


iloveeclipse:
I've rebased https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/406 on this commit. Let see if that helps to understand the problem.

iloveeclipse:
Unfortunately this fix was broken.
Jenkins didn't allowed your change to run (because of security), but it failed after merge with

```
[ERROR] Error executing Maven.
09:43:12  org.apache.commons.cli.MissingArgumentException: Missing argument for option: f
```
Proposed fix: https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/408

PR #408, opened 2023-01-22
Fix maven options to be in the right order in Jenkinsfile

iloveeclipse:
Fix for https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/407


Bananeweizen:
Is there any good reason for using "-f pom.xml" at all? That's the default, so it can be left out.

akurtakov:
> Is there any good reason for using "-f pom.xml" at all? That's the default, so it can be left out.

Please submit a PR about that.

PR #409, opened 2023-01-22
remove superfluous maven argument

Bananeweizen:
pom.xml is the default file, so it is not needed


Bananeweizen:
Coming from discussion in https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/408

iloveeclipse:
Thanks.

PR #415, opened 2023-01-25
Create new template permitting alternative method body statements.

JessicaJHee:
## What it does

* Adds code template that can be used to generate a method stub in the case of a super method invocation.
* Adds a parameter to`createImplementationStubCore()` to indicate whether the method to be generated is a super method invocation.

## How to test

The use of the modified methods are tested in https://github.com/eclipse-jdt/eclipse.jdt.ui/blob/master/org.eclipse.jdt.ui.tests/ui/org/eclipse/jdt/ui/tests/quickfix/AssistQuickFixTest1d8.java, and the behavior is retained. 

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.manipulation/common/org/eclipse/jdt/internal/core/manipulation/StubUtility.java
+++ b/org.eclipse.jdt.core.manipulation/common/org/eclipse/jdt/internal/core/manipulation/StubUtility.java
@@ -140,7 +140,13 @@ public class StubUtility {
 	 * Don't use this method directly, use CodeGeneration.
 	 */
 	public static String getMethodBodyContent(boolean isConstructor, IJavaProject project, String destTypeName, String methodName, String bodyStatement, String lineDelimiter) throws CoreException {
-		String templateName= isConstructor ? CodeTemplateContextType.CONSTRUCTORSTUB_ID : CodeTemplateContextType.METHODSTUB_ID;
+		return getMethodBodyContent(false, isConstructor, project, destTypeName, methodName, bodyStatement, lineDelimiter);
+	}
+
+	public static String getMethodBodyContent(boolean useAlternativeMethodBody, boolean isConstructor, IJavaProject project, String destTypeName, String methodName, String bodyStatement, String lineDelimiter) throws CoreException {
+		String templateName= isConstructor ? CodeTemplateContextType.CONSTRUCTORSTUB_ID
+							: useAlternativeMethodBody ? CodeTemplateContextType.METHODSTUB_ALTERNATIVE_ID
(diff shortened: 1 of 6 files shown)


JessicaJHee:
@rgrunber relating to https://github.com/redhat-developer/vscode-java/issues/1792

org.eclipse.jdt.core.manipulation/common/org/eclipse/jdt/internal/core/manipulation/StubUtility.java
```diff
@@ -140,7 +140,13 @@ public class StubUtility {
 	 * Don't use this method directly, use CodeGeneration.
 	 */
 	public static String getMethodBodyContent(boolean isConstructor, IJavaProject project, String destTypeName, String methodName, String bodyStatement, String lineDelimiter) throws CoreException {
-		String templateName= isConstructor ? CodeTemplateContextType.CONSTRUCTORSTUB_ID : CodeTemplateContextType.METHODSTUB_ID;
+		return getMethodBodyContent(false, isConstructor, project, destTypeName, methodName, bodyStatement, lineDelimiter);
+	}
+
+	public static String getMethodBodyContent(boolean isSuperMethodInvocation, boolean isConstructor, IJavaProject project, String destTypeName, String methodName, String bodyStatement, String lineDelimiter) throws CoreException {
+		String templateName= isConstructor ? CodeTemplateContextType.CONSTRUCTORSTUB_ID
+							: isSuperMethodInvocation ? CodeTemplateContextType.METHODSTUB_SUPER_ID
```
rgrunber, line 148: This doesn't really represent `isSuperMethodInvocation`. It's more like preserve-the-old-method-stub-for-parent-methods-when-the-stub-changes . I think the approach of creating a second template will probably work, but I'm wondering if we can avoid it.

I need to think about if there's another way to do this. I was thinking of the fact that each template has a context type, and each context type defines a set of variables that it can resolve. What if we just added another variable to the context type for the method body that represents the unsupported operation string. Between JDT-LS & JDT, we can have different templates, so maybe there's a way to just resolve that variable in both cases.

eg. If we set the "unsupported operation variable" here then evaluate the template as needed with it. Might need to discuss this further.

rgrunber: Ok I've come to see this as probably the best approach for now, but I still think we should rename the parameter because it's a lot more general than the use JDT-LS would have for it.

rgrunber (approved):
Looks good. Let's wait until the freeze is over to merge.

Issue #423, opened 2023-02-06
QuickFixTest1d8.testRemoveRedundantNonNull fails since I20230202-1800


iloveeclipse:
QuickFixTest1d8.testRemoveRedundantNonNull fails since I20230202-1800.
```
Wrong number of problems, is: 2, expected: 1 Pb(922) The nullness annotation is redundant with a default that applies to this location[91 ,105] Pb(922) The nullness annotation is redundant with a default that applies to this location[91 ,105]

java.lang.AssertionError: Wrong number of problems, is: 2, expected: 1
Pb(922) The nullness annotation is redundant with a default that applies to this location[91 ,105]
Pb(922) The nullness annotation is redundant with a default that applies to this location[91 ,105]

at org.junit.Assert.fail(Assert.java:89)
at org.eclipse.jdt.ui.tests.quickfix.QuickFixTest.assertNumberOfProblems(QuickFixTest.java:254)
at org.eclipse.jdt.ui.tests.quickfix.QuickFixTest.collectCorrections(QuickFixTest.java:229)
at org.eclipse.jdt.ui.tests.quickfix.QuickFixTest.collectCorrections(QuickFixTest.java:224)
at org.eclipse.jdt.ui.tests.quickfix.QuickFixTest1d8.testRemoveRedundantNonNull(QuickFixTest1d8.java:1743)
```

Regression from https://github.com/eclipse-jdt/eclipse.jdt.core/pull/683

iloveeclipse:
@stephan-herrmann : I've pushed https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/425 in case it would take longer to fix JDT core problem, so that JDT UI builds can succeed in Jenkins (right now they all fail due the broken test).

Just give a hint if that should be merged or not, I will wait with merging for now, especially ECA doesn't work anymore today: https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/2635

stephan-herrmann:
Thanks for the heads-up. Of course no warning should be reported twice.

I'll look into it later today. If you need to fix JDT/UI right now, feel free to merge #425 - I'd revert that when the issue in JDT/Core is fixed.

iloveeclipse:
@stephan-herrmann : I will wait, no one can merge anyway now. 
In best case you will have a fix in near future and we can close #425 without merging.

stephan-herrmann:
[...]

PR #425, opened 2023-02-06
Adopt test expectation after eclipse.jdt.core/pull/683 (#423)

iloveeclipse:
Fixes https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/423


iloveeclipse:
Original problem fixed in core.

PR #429, opened 2023-02-07
Make Text Block indentation consistent

### LINKED ISSUE #414: Indent action indents unformatted text block wrongly
This is a continuation of #30, which was fixed only partially with #33. We use the `Correct Indentation` save action and its applying odd indentations in our code.

E.g. with default formatter settings, the following (already formatted) snippet is no longer indented wrongly due to #30:

```
		String a = """
				    line1
				    line2
				""";
		System.out.println(a);
	}
```

However, for a snippet that is not formatted, the formatting done by the indent action is wrong when compared to formatting the source.

Unformatted snippet:

```
		String a =
				"""
				    line1
				    line2
				""";
		System.out.println(a);
	}
```

Formatting the whole snippet: 

```
		String a = """
				    line1
				    line2
				""";
		System.out.println(a);
	}
```

Indent action after selecting the snippet, this is also done for the `Correct Indentation` save action:

```
		String a =
				"""
						    line1
						    line2
						""";
		System.out.println(a);
	}
```

This is due to the indent action formatting each line separately, in which case the default formatter options result in the misaligned lines seen above. The fix for #30 addresses some cases of this (e.g. select the text block and `String a =`), but not all.

The expected outcome of the save action is to have the source look like formatted as a whole, since its supposed to correct indentations for all of the file.

jjohnstn:
- make text block indentation the same whether the open quotes occur on a separate line or at the end of another line
- fixes #414

## What it does

Fixes #414

## How to test

See issue.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/IndentActionTest15.java
+++ b/org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/IndentActionTest15.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2005, 2022 IBM Corporation and others.
+ * Copyright (c) 2005, 2023 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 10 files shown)


jjohnstn:
Test failure is a known issue with jdt.core and has been addressed recently.

trancexpress:
I checked, the save action works as expected now. Its consistent with formatting the file with `Ctrl+Shift+F`.

Oddly selecting text and running the indent action behaves differently to the save action. If I select the entire file and run the indent action `Ctrl+I` nothing happens with the unformatted snippet from the description.

Both cases are fine for me, its just odd to see the 2 types of behavior, I thought the code was the same for both cases.

PR #438, opened 2023-02-25
Apply semantic syntax coloring in Java source hover viewer. Fixes #437

### LINKED ISSUE #437: Limited syntax coloring in JavaSourceHover
Java Source hover has only limited syntax coloring from only basic java source content parsing.
Advanced semantic coloring is not applied.

Example:
image

Expected:
Java source in hover viewer should be colored the same way as if it was opened in source code editor.

RedeemerSK:
## What it does
Adds full semantic coloring to content of Java source hover viewer.
~~Functionality can be toggled with new checkbox added to Preferences / Java / Editor / Syntax Coloring / "Full semantic syntax coloring in Java source hover" at the very bottom, enabled by default.~~ edit: removed

Behaves exactly like current hover in terms of presented content, size and reaction when trying to "enter into" hover viewer with mouse cursor.
Only difference is that when hovering over class / method / field / constant / variable name, if the element has JavaDoc or single/multi line comment above it's declaration, it will also be part of hover viewer content, but will not be presented / visible right away, but only after user "enters into" viewer and scrolls up.
This difference is intentional since I believe comments, especially JavaDoc, can be also useful piece of information to present in hover viewer.

## How to test
Compare contents, dimensions and behavior (when "entering into" hover viewer with mouse cursor) of Java source hover viewer with one displayed when running current release of Eclipse IDE when (+Shift) hovering over :

- class name
- method name
- field / constant name
- variable name
- end bracket of any loop
- any end bracket within if() {} else{} block (*)
- any end bracket within  if() {} else if() {} (*) (may be multiple "else if" sub-blocks)
- module name

(* starting "if" must be out of source code range displayed by the editor)

## Author checklist

### DIFF
--- a/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/javaeditor/SemanticHighlightingManager.java
+++ b/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/javaeditor/SemanticHighlightingManager.java
@@ -310,16 +310,34 @@ public class SemanticHighlightingManager implements IPropertyChangeListener {
 	}
 
 	/**
+	 * Create semantic highlighting presenter instance to use.
+	 *
+	 * @return semantic highlighting presenter to use
+	 */
+	protected SemanticHighlightingPresenter createSemanticHighlightingPresenter() {
+		return new SemanticHighlightingPresenter();
+	}
+
+	/**
+	 * Create semantic highlighting reconciler instance to use.
+	 *
+	 * @return semantic highlighting reconciler to use
+	 */
+	protected SemanticHighlightingReconciler createSemanticHighlightingReconciler() {
+		return new SemanticHighlightingReconciler();
+	}
+
+	/**
 	 * Enable semantic highlighting.
 	 */
 	private void enable() {
 		initializeHighlightings();
 
-		fPresenter= new SemanticHighlightingPresenter();
+		fPresenter= createSemanticHighlightingPresenter();
 		fPresenter.install(fSourceViewer, fPresentationReconciler);
 
 		if (fEditor != null) {
-			fReconciler= new SemanticHighlightingReconciler();
+			fReconciler= createSemanticHighlightingReconciler();
 			fReconciler.install(fEditor, fSourceViewer, fPresenter, fSemanticHighlightings, fHighlightings);
 		} else {
 			fPresenter.updatePresentation(null, createHardcodedPositions(), new HighlightedPosition[0]);
--- a/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/javaeditor/SemanticHighlightingReconciler.java
+++ b/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/javaeditor/SemanticHighlightingReconciler.java
@@ -537,11 +537,21 @@ public class SemanticHighlightingReconciler implements IJavaReconcilingListener,
 		fSourceViewer= sourceViewer;
 
 		if (fEditor instanceof CompilationUnitEditor) {
-			((CompilationUnitEditor)fEditor).addReconcileListener(this);
-		} else if (fEditor == null) {
+			if (registerAsEditorReconcilingListener()) {
+				((CompilationUnitEditor)fEditor).addReconcileListener(this);
+			}
+		} else if (fEditor != null) {
 			fSourceViewer.addTextInputListener(this);
 			scheduleJob();
 		}
+	}
+
+	/**
+	 * Decides if this reconciler should also register itself as a reconciling listener on the editor as part of {@link #install} process.
+	 * @return whether this instance should register itself as a reconciling listener on the editor
+	 */
+	protected boolean registerAsEditorReconcilingListener() {
+		return true;
 	}
 
 	/**
@@ -566,10 +576,19 @@ public class SemanticHighlightingReconciler implements IJavaReconcilingListener,
 	}
 
 	/**
+	 * Get the type root java element for which reconciling job should be scheduled.
+	 *
+	 * @return root element for reconciling
+	 */
+	protected ITypeRoot getElement() {
+		return fEditor.getInputJavaElement();
+	}
+
+	/**
 	 * Schedule a background job for retrieving the AST and reconciling the Semantic Highlighting model.
 	 */
 	private void scheduleJob() {
-		final ITypeRoot element= fEditor.getInputJavaElement();
+		final ITypeRoot element= getElement();
 
 		synchronized (fJobLock) {
 			final Job oldJob= fJob;
--- a/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/text/java/hover/JavaSourceHover.java
+++ b/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/text/java/hover/JavaSourceHover.java
@@ -14,6 +14,10 @@
 package org.eclipse.jdt.internal.ui.text.java.hover;
 
 import java.io.IOException;
+import java.util.ArrayList;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.stream.IntStream;
 
 import org.eclipse.swt.SWT;
 import org.eclipse.swt.SWTException;
@@ -21,6 +25,7 @@ import org.eclipse.swt.custom.StyledText;
 import org.eclipse.swt.graphics.Point;
 import org.eclipse.swt.graphics.Rectangle;
 import org.eclipse.swt.widgets.Display;
+import org.eclipse.swt.widgets.Shell;
 
 import org.eclipse.core.runtime.CoreException;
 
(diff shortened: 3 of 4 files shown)


RedeemerSK:
### Implementation details:
`JavaSourceHover` is modified to provide (as source viewer information input) enhanced sub-classes of existing `JavaSourceInformationInput`, that are used by `SourceViewerInformationControl` to prepare and display hover viewer widget. 

`SourceViewerInformationControl` is made to use combination of existing `SemanticHighlightingManager` with sligthly modified `SemanticHighlightingReconciler` & `SemanticHighlightingPresenter` classes so that it can trigger semantic highlighting reconciliation of source code that does not come from active editor (manager's current / default use-case), but from origin (compilation unit) of hovered element that may or may not be the same as one for active editor.

`SourceViewerInformationControl`, if instructed to do so (by enhanced input), uses `SemanticHighlightingManager` to do semantic highlighting reconciliation of whole (from what I can tell that's the only way) source code of the origin of hovered element, which also applies semantic coloring to the contents of the viewer. After application of styles, hover viewer content is limited to only part corresponding to the hovered element and further modified as needed (line trimming and/or skipped lines substitutions).

If the origin of the hovered element is the same as one of active editor, the 1st step of reconciliation job, that is analysis of code by calling `SharedASTProviderCore.getAST`, takes the "isActiveElement" code path and just returns already cached AST. If however origin differs, display of the hover viewer widget is postponed until after reconciliation = syntax coloring and further modifications of the viewer content are done.
Not having this delay would result in coloring "popping in" (being applied onto) already displayed viewer content which I found slightly disturbing. But if retaining the speed of displaying the viewer is preferred, it's an easy change.

### Current:
image

### New:
image

### Additional JavaDoc "hidden" above inside viewer, visible after scroll (or resize):
image

iloveeclipse:
I haven't reviewed the logic, but saw that you've added "since" tags all over. They aren't needed on internal (non API) code.

Note: we are in RC2 time now, so it can't be merged yet and has to wait till master opens.

RedeemerSK:
One thing on the topic of speed (started in [#437](https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/437#issuecomment-1445224071)):

Current behavior when user "enters into" displayed non-resizable hover viewer with mouse cursor is that `JavaSourceHover` creates new `SourceViewerInformationControl` that goes through exactly same initialization (and now also semantic coloring) process as the one being entered into.

I was thinking maybe this is unnecessary and that either existing `SourceViewerInformationControl` instance could be re-used (just making it resizable) or if new instance turns out to be necessary, then it could just re-use existing Shell / StyledText / SourceViewer / Document  (whatever would work). But I haven't yet looked into what that would mean for uses of `SourceViewerInformationControl` other than from `JavaSourceHover`.

What do you think about this optimization @iloveeclipse, is it worth looking into it ? Thx

RedeemerSK:
> One thing on the topic of speed (started in [#437](https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/437#issuecomment-1445224071)):
> 
> Current behavior when user "enters into" displayed
> ....

Made sense for me, so it's added to PR already.

org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/javaeditor/SemanticHighlightingReconciler.java
```diff
@@ -537,13 +537,23 @@ public void install(JavaEditor editor, ISourceViewer sourceViewer, SemanticHighl
 		fSourceViewer= sourceViewer;
 
 		if (fEditor instanceof CompilationUnitEditor) {
-			((CompilationUnitEditor)fEditor).addReconcileListener(this);
-		} else if (fEditor == null) {
```
RedeemerSK, line 541: This was I believe always wrong since if `scheduleJob()` is run when `fEditor` is `NULL`, it will thrown NPE there.

org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/javaeditor/SemanticHighlightingManager.java
```diff
@@ -309,17 +309,37 @@ public void install(JavaSourceViewer sourceViewer, IColorManager colorManager, I
 		install(null, sourceViewer, colorManager, preferenceStore);
 	}
 
+	/**
+	 * Create semantic highlighting presenter instance to use.
+	 *
+	 * @return semantic highlighting presenter to use
+	 * @since 3.28
```
RedeemerSK, line 316: Regarding unnecessary `@since` tags on "non API" code ... I considered new non-private methods as API code - API of the classes usable from potential extensions (subclasses). If this reasoning is too strict and thus wrong here, let me know and I'll remove remaining tags.

iloveeclipse: "Since" tags are used by API tooling to check for API breakage between releases, and so make only sense on real API code. Put on internal code they only conffuse. So please remove them.

RedeemerSK:
I'm also wondering what was the reasoning to not make it possible to enter hover viewer displayed over curly bracket and being able to see the whole source code of the AST node without skipped lines in (now) resizable viewer. Seems like missed opportunity for another useful feature to me.

iloveeclipse:
Hi, I see there are now 7? commits with partial reworked code after review. Could we have less commits by combining all "review" changes into ome main commit? 

I see value in separated bug fix commit at beginning and improvement commit next to it, that would be two commits, but not 7?

RedeemerSK:
> Hi, I see there are now 7? commits with partial reworked code after review. Could we have less commits by combining all "review" changes into ome main commit?
> 
> I see value in separated bug fix commit at beginning and improvement commit next to it, that would be two commits, but not 7?

Turned out that changes that would be left for the 2nd commit (following the 1st main one that includes all the review changes) are very few, so I just made it into single commit.

Rebased on current HEAD.

org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/text/java/hover/JavaSourceHover.java
```diff
@@ -575,59 +873,93 @@ private IInformationControlCreator createInformationControlCreator(final boolean
 			if (editor instanceof IWorkbenchPartOrientation)
 				orientation= ((IWorkbenchPartOrientation) editor).getOrientation();
 
-			return new SourceViewerInformationControl(parent, isResizable, orientation, statusFieldText) {
```
RedeemerSK, line 578: Only indentation changed for existing methods.

RedeemerSK:
Any chance this will make it into 2023-06 ?
(From looking at SimRel-2023-06 plan I can't really tell when code freeze for JDT starts)

iloveeclipse:
> Any chance this will make it into 2023-06 ? (From looking at SimRel-2023-06 plan I can't really tell when code freeze for JDT starts)

I'm sorry, there is no chance because we have RC1 tomorrow and no features should be merged, only regressions / bug fixes.

I hope you can understand it, and don't give up. Ping again once 4.29 starts please.

RedeemerSK:
Ping @iloveeclipse 
Rebased on HEAD

jjohnstn:
@iloveeclipse sure, will do

RedeemerSK:
> Behaves exactly like current hover in terms of presented content, size and reaction ...

except this ignored indentation (trimming) annoyance that is also fixed now
old:
image
new:
image

jjohnstn:
Hi @RedeemerSK First of all, you need to review any review comments and either fix them or discuss why the comment is not valid.  Second, you need to sign the ECA for this patch to ever be merged.  See under details of eca failure above.

Now, regarding latest changes to the patch, you fixed a number of issues I found earlier which was indentation and hovering over classes with errors didn't work previously (e.g. class with missing parentheses).  I think your change to remove the delay is the right idea, but there is a problem I am seeing:  indentation movement occurring after the hover displays.  I believe the solution to consider is to do the line trimming for the initially displayed lines before displaying and reconciling.  You can do the rest of the line trimming after.

RedeemerSK:
> you need to sign the ECA

Done (I did not notice previously that my second commit I made from my other computer did not have proper author & committer set, sorry)

> there is a problem I am seeing: indentation movement occurring after the hover displays

That's exactly what I was trying to avoid by disabling rendering of the hover until everything is trimmed & indented. I don't see any such movement on my setup (win 11). What's your platform ? Maybe it will turn out that setRedraw(false) does not work as execpted for others (even its JavaDoc says it may not be supported on some platforms).

> do the line trimming for the initially displayed lines before displaying and reconciling. You can do the rest of the line trimming after.

At first I thought you're suggesting exactly how it's implemented in my commits right now, because really, text displayed at first is JavaSourceInformationInput.fHoverInfo, which is already properly line-trimmed. But then I understood that you probably meant something like setting full origin source as hover's document, but with (what will become) visible area already properly line-trimmed and then showing it + starting reconciling job. So basically instead of setting & showing fHoverInfo, set & show line-trimmed fFullContent.

If that's the case, then the problem is that for reconciling job to be able (at the end) to apply coloring properly= so that it aligns with text, the source viewer has to contain **unmodified** origin source code. Otherwise coloring would not align with text, but most often would even fail with BadLocationException. That's why 2nd pass of (same) line-trimming has to be done again after reconciling on at that moment already colored origin source. And the ultimate reason for this is that I am (atm) unaware of any way how I could do AST "analysis" of custom text, so I am using origin source.

RedeemerSK:
> > there is a problem I am seeing: indentation movement occurring after the hover displays
> 
> That's exactly what I was trying to avoid by disabling rendering of the hover until everything is trimmed & indented. I don't see any such movement on my setup (win 11). What's your platform ? Maybe it will turn out that setRedraw(false) does not work as execpted for others (even its JavaDoc says it may not be supported on some platforms).

On Win 10 no movement as well. But I've confirmed your observation on Ubuntu 22.04 (running inside WSL, but well, still enough to reproduce it).

I'll try to come up with a way to fix it ... I'll probably need to use different (dummy) source viewer for purposes of reconciliation and then just grab coloring ranges and apply them to displayed one. I would like to avoid going back to previous approach that was to display source viewer only once everything is done.

jjohnstn:
> > > there is a problem I am seeing: indentation movement occurring after the hover displays
> > 
> > 
> > That's exactly what I was trying to avoid by disabling rendering of the hover until everything is trimmed & indented. I don't see any such movement on my setup (win 11). What's your platform ? Maybe it will turn out that setRedraw(false) does not work as execpted for others (even its JavaDoc says it may not be supported on some platforms).
> 
> On Win 10 no movement as well. But I've confirmed your observation on Ubuntu 22.04 (running inside WSL, but well, still enough to reproduce it).
> 
> I'll try to come up with a way to fix it ... I'll probably need to use different (dummy) source viewer for purposes of reconciliation and then just grab coloring ranges and apply them to displayed one. I would like to avoid going back to previous approach that was to display source viewer only once everything is done.

Thanks.  Agreed that waiting to display source viewer isn't a good approach.

jjohnstn:
Hi @RedeemerSK Any progress on fixing the indentation movement?  4.30 M1 is now open for patches and there is time to get this in.

RedeemerSK:
Hi
sorry for long silence, summer happened.

I have the code change prepared, I just want to check few last things before pushing the commit. Should not take more than a day or two ... hopefully.

jjohnstn:
> Hi sorry for long silence, summer happened.
> 
> I have the code change prepared, I just want to check few last things before pushing the commit. Should not take more than a day or two ... hopefully.

Ok, the deadline for M1 is approaching.  It would be best if you could have this by tomorrow.  Remember to rebase as the Java 21 branch has recently been merged plus a number of changes are getting in for M1.

RedeemerSK:
Rebased the branch and added the fix commit. Looks good to me also on Linux now (with the limited environment I have), but that has to be confirmed by people running on Linux natively. Same with MacOS.

jjohnstn:
Thanks @RedeemerSK It is working much better on Linux with your new fixes.  Merging for 4.30 M1.

RedeemerSK:
Found https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/936, PR for fix https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/937

iloveeclipse:
Unfortunately this change caused regression, see https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1125

RedeemerSK:
> Unfortunately this change caused regression, see https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1125

Fix PR #1126

PR #443, opened 2023-02-27
In Java editor avoid flickering and key conflict for Alt+Shift+R

howlger:
In Eclipse 2023-03 M3 there is a **conflict** for the key **Alt+Shift+R** **causing a flickering** for one or two seconds (at least on Windows 11 and when using a fresh workspace):
- **_Rename_** (Category: _File_)
- **_Rename - Refactoring_** (Category: _Refactor Java_)
both with _When_ set to **_In Windows_**.

Steps to reproduce this issue:
1.  In Eclipse 2023-03 M3 with a fresh workspace create a Java file containing `class Sample {}`
2. Set the cursor somewhere into `Sample` 
3. Hit Alt+Shift+R

The key conflict is logged in the error log:
```
A conflict occurred for ALT+SHIFT+R:
Binding(ALT+SHIFT+R,
	ParameterizedCommand(Command(org.eclipse.jdt.ui.edit.text.java.rename.element,Rename - Refactoring ,
		Rename the selected element,
		Category(org.eclipse.jdt.ui.category.refactoring,Refactor - Java,Java Refactoring Actions,true),
		WorkbenchHandlerServiceHandler("org.eclipse.jdt.ui.edit.text.java.rename.element"),
		,,true),null),
	org.eclipse.ui.defaultAcceleratorConfiguration,
	org.eclipse.ui.contexts.window,,,system)
Binding(ALT+SHIFT+R,
	ParameterizedCommand(Command(org.eclipse.ui.edit.rename,Rename,
		Rename the selected item,
		Category(org.eclipse.ui.category.file,File,null,true),
		WorkbenchHandlerServiceHandler("org.eclipse.ui.edit.rename"),
		,,true),null),
	org.eclipse.ui.defaultAcceleratorConfiguration,
	org.eclipse.ui.contexts.window,,,system)
```

Setting _When_ of **_Rename - Refactoring_** to **_Editing Java Source_** **fixes** the **flickering** and the **key conflict**. But the drawback is that Alt+Shift+R no longer works in the _Overview_ view.


iloveeclipse (changes requested):
Please try to do same as in https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/476

howlger:
> Please try to do same as in #476

It binds JDT's rename command to the more specific context `org.eclipse.jdt.ui.javaEditorScope` of JDT. So doesn't it do the same ?

Should JDT's rename command better be a handler of the general platform rename command of the platform, rather than a JDT command of its own?

iloveeclipse:
> It binds JDT's rename command to the more specific context `org.eclipse.jdt.ui.javaEditorScope` of JDT. So doesn't it do the same?

Not. https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/476 **adds** another key binding with smaller scope, so it works in both global window context and in a local editor one.

howlger:
> > It binds JDT's rename command to the more specific context `org.eclipse.jdt.ui.javaEditorScope` of JDT. So doesn't it do the same?
> 
> Not. #476 **adds** another key binding with smaller scope, so it works in both global window context and in a local editor one.

I see. But that would solve the conflict only for the editors, but not in the _Outline_ view of a Java editor and not in the _Project/Package Explorer_, right?

Whatever, shouldn't the command be a handler instead of a command of its own?

iloveeclipse:
> But that would solve the conflict only for the editors, but not in the _Outline_ view of a Java editor and not in the _Project/Package Explorer_, right?

According to @mickaelistria the problem is supposed to happen only in presence of active editor.

> Whatever, shouldn't the command be a handler instead of a command of its own?

Not sure what do you mean by that and how that is related to the current issue.

howlger:
> > According to @mickaelistria the problem is supposed to happen only in presence of active editor.
> 
> Yes, at least with latest LSP4E snapshots (upcoming release in a few days). It is probable that previous release had a bug that caused the handlers to be enabled more often than necessary; but in theory, they should only be enabled when active editor is about a document that has some language server support.

As I understand it, the situation here is different than in [#476](https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/476), since JDT uses the rename command not only in the editor.

> > Whatever, shouldn't the command be a handler instead of a command of its own?
> 
> Not sure what do you mean by that and how that is related to the current issue.

I am not an expert, but shouldn't a specialization of an existing command be implemented as a handler (without a key binding) rather than a command of its own and with the same key binding that causes the key bindings conflict?

iloveeclipse:
> As I understand it, the situation here is different than in [#476](https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/476), since JDT uses the rename command not only in the editor.

No, it is not, it is same, Ctrl+Shift+T is not bound to editor context only and can be used in an empty window.

howlger:
> > As I understand it, the situation here is different than in [#476](https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/476), since JDT uses the rename command not only in the editor.
> 
> No, it is not, it is same, Ctrl+Shift+T is not bound to editor context only and can be used in an empty window.

You are right. Adding an additional key binding for _Rename - Refactoring_ with _When_ set to _Editing Java Source_ (in addition to the key binding for the same command, but with _When_ set to _In Windows_) fixes the issue. Eclipse needs to be restarted to apply this fix also to the _Package Explorer_ and _Projects Explorer_ views (for the Java editor no restart is needed). Also, with this fix, there is no longer a key conflict logged for Alt+Shift+R. So, the only known drawback is that there are several commands and key bindings in the preferences for _Rename_.

iloveeclipse:
Thanks Holger.

oleosterhagen:
Does this change really solves the key binding conflict when hitting `Shift+Alt+R` for a Java file in the **Package Explorer**? Or is this a remaining issue?

**Some background**

There exists two commands with different key bindings:

1. `org.eclipse.ui.edit.rename` (Rename) with key binding **F2** (defined in `org.eclipse.ui`)
2. `org.eclipse.jdt.ui.edit.text.java.rename.element` (Rename - Refactoring) with key binding **Shift+Alt+R** (defined in `org.eclipse.jdt.ui`)

For both commands `org.eclipse.jdt.ui` registers action handlers which invoke the `RenameAction`. So far no ambiguities.

Then `org.eclipse.lsp4e` introduces an additional key binding **Shift+Alt+R** for `org.eclipse.ui.edit.rename`. From now on the key binding **Shift+Alt+R** is associated with _two_ commands:

1. org.eclipse.ui.edit.rename (Rename)
2. org.eclipse.jdt.ui.edit.text.java.rename.element (Rename - Refactoring)

A little popup appears and the user has to select the desired command. It doesn't matter that both commands will invoke the same rename action. This is no new problem. I see it since Eclipse 2021-12 JEE.

A possible solution would be to introduce an **own command id** for the LSP4E rename command (like org.eclipse.jdt.ui.edit.text.java.rename.element in JDT) and associate the key binding **Shift+Alt+R** there only to this new command. Because the handler for the rename command in LSP4E already has an `activeWhen` expression to narrow down the applicability, no window should be displayed anymore.

iloveeclipse:
> Does this change really solves the key binding conflict when hitting `Shift+Alt+R` for a Java file in the **Package Explorer**? Or is this a remaining issue?

Please grab next nightly platform build and validate that.

mickaelistria:
> Does this change really solves the key binding conflict when hitting Shift+Alt+R for a Java file in the Package Explorer? Or is this a remaining issue?

No, but this one is supposed to be fixed if you use latest release or snapshot of LSP4E.

oleosterhagen:
I have
- installed Eclipse nightly build from [https://download.eclipse.org/eclipse/downloads/drops4/I20230317-1800/](https://download.eclipse.org/eclipse/downloads/drops4/I20230317-1800/)
- validated that the change from this PR is included
- installed latest LSP4E snapshot from [https://download.eclipse.org/lsp4e/snapshots/](https://download.eclipse.org/lsp4e/snapshots/)

Selecting a Java class in the **Package Explorer** and pressing `Shift+Alt+R` still opens the popup with options _Rename_ and _Rename - Refactoring_ (both from JDT UI).

I think there are several ways to solve this inconvenience. Adjusting this in LSP4E would lead only to few changes. I have created a PR: eclipse/lsp4e/pull/559

howlger:
I can reproduce what @oleosterhagen says: installing latest LSP4E snapshot into latest Eclipse SDK nightly causes the issue to re-appear in the _Project Explorer_.

Instead of having different _rename_ commands (the latest Eclipse SDK nightly has four _rename_ commands bound to Alt+Shift+R; https://github.com/eclipse/lsp4e/pull/559 would add another one), I would prefer to have only one _rename_ command which would make changing the Alt+Shift+R key binding easier. If needed, Eclipse platform might add a _Rename without Refactoring_ command without key binding by default.

Apart from that, there should be no flickering when there is a key binding conflict. I was not able to reproduce this with other commands. So I have no glue whether this is caused by JDT's _Rename - Refactoring_ command or by something in the Eclipse platform.

oleosterhagen:
For me, a very good **first time user experience** in Eclipse is important. Like before the release of Eclipse JEE 2021-12.

The current situation with the popup window is not that nice. When you know, you can work around this problem by using only `F2` for rename in the _Package Explorer_. But this is no permanent solution.

howlger:
> For me, a very good **first time user experience** in Eclipse is important. Like before the release of Eclipse JEE 2021-12.
> 
> The current situation with the popup window is not that nice. When you know, you can work around this problem by using only `F2` for rename in the _Package Explorer_. But this is no permanent solution.

Yes, this issue causes a really bad user experience. I wonder whether we are heading in the wrong direction with this fix, forcing other plugins like LSP4E (maybe also CDT, etc.) offering a _rename_ in the _Project Explorer_ to create their own _rename_ commands instead of specifying a command handler. So the question is, why does JDT have its own _Rename - Refactoring_ command and would this usability issue be fixed by JDT using a command handler instead of its own command?

PR #476, opened 2023-03-14
Forces key binding to Open Type inside JDT editor

mickaelistria:
Preventing conflicts with other plugins that provide same binding for other commands in more generic context

## Author checklist

### DIFF
--- a/org.eclipse.jdt.ui/plugin.xml
+++ b/org.eclipse.jdt.ui/plugin.xml
@@ -4909,6 +4909,13 @@
             sequence="M1+M2+T"
             commandId="org.eclipse.jdt.ui.navigate.open.type"
             schemeId="org.eclipse.ui.defaultAcceleratorConfiguration"/>
+      <!-- Forces M1+M2+T to Open Type inside JDT editor, preventing conflicts with other
+           plugins that provide same binding for other commands in more generic contexts -->
+      <key
+            sequence="M1+M2+T"
+            contextId="org.eclipse.jdt.ui.javaEditorScope"
+            commandId="org.eclipse.jdt.ui.navigate.open.type"
+            schemeId="org.eclipse.ui.defaultAcceleratorConfiguration"/>
       <key
             sequence="M1+M2+H"
             commandId="org.eclipse.jdt.ui.navigate.open.type.in.hierarchy"


mickaelistria:
Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/513

iloveeclipse:
Few things:
- please add link to original issue **in the commit message**
- what about ctrl+shift+r which was also reported to be broken?
- does this also solve a problem with conflicting key bindings if there is no editor active?

mickaelistria:
> please add link to original issue in the commit message

Done.

> what about ctrl+shift+r which was also reported to be broken?

Next time, but probably same solution.

> does this also solve a problem with conflicting key bindings if there is no editor active?

I don't think it solves such a problem, if it exists. At least with LSP4E, the command has no handler when no document is being edited.

jukzi:
Thanks for worknig on a solution. Wouldn't it be better to ship any IDE that includes JDT UI without conflicting key bindings? I guess people get confused when the well known key binding sometimes opens another dialog that wasn't supposed to be opened. You could probably add an additional plugin (say "LSP4E4WithoutJdtUI") in any non-JDT IDE that registers the key binding - without having any conflicts then.

jjohnstn:
The patch would solve the problem when in the Java editor but does not solve the problem when one is in the Package Explorer, the Outline View, the Javadoc view, etc...

Interestingly enough, using the Eclipse Java EPP for 2022-12, if I actually select to use "Go to symbol in Workspace", I get a dialog that indicates the operation is not enabled.  After that dialog, the binding opens the Open Type dialog in all the views mentioned without requiring me to choose from the conflicts.  If I quit and restart, I have to choose to use it again, get the dialog, then have it disabled from then on.

Could there be a way to dynamically disable this from the beginning for Java EPP users?

mickaelistria:
> The patch would solve the problem when in the Java editor but does not solve the problem when one is in the Package Explorer, the Outline View, the Javadoc view, etc...

Do you see this issue when the selection is in the mentioned views?
Many things have changed in LSP4E snapshots lately compared to latest release that is in SimRel/EPP; so a part of the fix may actually be also in LSP4E. However, even with latest LSP4E, the problem persists in the editor when[ eclipseide-jdtls ](https://github.com/redhat-developer/eclipseide-jdtls) is installed, and this patch fixes it.

mickaelistria:
> Could there be a way to dynamically disable this from the beginning for Java EPP users?

The problem is not about Java EPP, it's about JDT + any language server that affects .java files (Spring LS, LSP4MP, LSP4Jakarta, QuarkusLS...)

mickaelistria:
I noticed #443 already covers rename shortcut

jjohnstn:
> > The patch would solve the problem when in the Java editor but does not solve the problem when one is in the Package Explorer, the Outline View, the Javadoc view, etc...
> 
> Do you see this issue when the selection is in the mentioned views? Many things have changed in LSP4E snapshots lately compared to latest release that is in SimRel/EPP; so a part of the fix may actually be also in LSP4E. However, even with latest LSP4E, the problem persists in the editor when[ eclipseide-jdtls ](https://github.com/redhat-developer/eclipseide-jdtls) is installed, and this patch fixes it.

I was just using the 2022-12 Java EPP to see if it occurred there and it did.  With your patch, it solved the problem with focus in the editor, but I was still seeing the conflict when focus was in the aforementioned views until I chose to use the "Go to Symbol" to cause the disablement.  I see there is a patch mentioned above by @oleosterhagen which would solve that so that is good.

I have no issues with this patch as it would at least be better for the eclipseide-jdtls editor case and wouldn't cause any additional issues.  You allude that the LSP4E code may have fixed the issue from occurring in the other views mentioned, but if not confirmed, this is still troubling.

A nicer conflict strategy might also be useful.  For example, on first conflict, a dialog could open to allow a user to select one as default and disable the other(s) from then on or to narrow context or to leave the conflict as-is and allow the user to select.  A bit easier than forcing the user to manually edit the key bindings.  I myself would choose to disable the Go to Symbol binding if given the option.

mickaelistria:
> A nicer conflict strategy might also be useful.

The point here is not really to provide a nicer conflict strategy, but more to enforce that in the Java editor, Open Type is used *without* risk of conflict.

jjohnstn:
> > A nicer conflict strategy might also be useful.
> 
> The point here is not really to provide a nicer conflict strategy, but more to enforce that in the Java editor, Open Type is used _without_ risk of conflict.

I should have put my suggestion in the issue instead.  As a Java developer, I use CTRL+SHIFT+T everywhere, every day, all the time.  I have never used the other function for the binding and don't want to be offered it, ever.  I imagine other Java developers used to Open Type might agree with that as well.  There may be no way to perfectly solve this and then both sets of users of the binding are probably going to be annoyed unless they manually change one of the bindings.  It was an idea for a compromise that might alleviate some user frustration if there isn't a way to resolve this.

mickaelistria:
> if there isn't a way to resolve this.

Other issues got fixed in LSP4E so the issue in other views should now be fixed. But the case of the editor is still bogus when eclipseide-jdtls is installed; this patch fixes that particular case.

kriegaex:
Sorry for commenting on a PR rather than on an issue, I would not know which one to comment on, as there are too many of them, relating to the same issue, as this `Shift+Ctr-R` problem has been present across several Eclipse IDE releases at least since 4.23 and never been taken seriously. Just look at the cluster of old bugs https://bugs.eclipse.org/bugs/show_bug.cgi?id=579168 and its [16(!) alleged duplicates](https://bugs.eclipse.org/bugs/buglist.cgi?bug_id=579300%2C579316%2C579321%2C579339%2C579352%2C579365%2C579366%2C579367%2C579373%2C579380%2C579389%2C579400%2C579411%2C579539%2C579569%2C579613).

The original bug was something in connection with Lombok. After that one was resolved by telling users to upgrade Lombok, many new issues totally unrelated to Lombok were immediately marked as duplicates of that one without ever having been triaged properly. Most issues have not even been commented on. Hence, the problem was never properly investigated and fixed. It still occurs in 4.27 (2023-03) and can be worked around by changing the keyboard shortcuts for 3 refactor/rename actions from `Shift-Ctrl-R` to something different. In my case, I chose `Shift-F6`, because in IntelliJ it is the same. Interestingly, this works, and even for the conflicting action, I get a popup menu for me to make a choice:

image

With the original key binding, there just was flickering, but nothing happened, not even an error in the "Error log" view.

In addition to the 16 so-called duplicates in Bugzilla, there are several bugs here on GitHub, all pointing to the same problem. I suggest to the JDT Core team to really take each single issue seriously and not to dismiss them too quickly as duplicates, which leads to real problems being ignored for far too long. I think that at the end of the day, it saves more time to triage than to spend lots of time, falsely dismissing issues as duplicates, just because they look similar. The same symptom does not automatically mean that we are dealing with the same root cause.

Having said that, please also know that I do acknowledge the fact that the JDT Core dev team is stretched thin for resources. I know the situation from AspectJ. So please do not mistake this post for a rant, because it is not. I am not angry or annoyed, not even a bit. I just want us all to learn something for the future.

trancexpress:
> In addition to the 16 so-called duplicates in Bugzilla, there are several bugs here on GitHub, all pointing to the same problem. I suggest to the JDT Core team to really take each single issue seriously and not to dismiss them too quickly as duplicates, which leads to real problems being ignored for far too long. I think that at the end of the day, it saves more time to triage than to spend lots of time, falsely dismissing issues as duplicates, just because they look similar. The same symptom does not automatically mean that we are dealing with the same root cause.

From what I see in the bugzilla ticket, aside from @iloveeclipse no one else from the JDT maintainers has spent time on the issues linked as duplicates (not even to close them as duplicates). Andrey, do you plan on allocating resources for this? Or do you know of any JDT maintainer who plans to do this? If not, maybe this should be brought up on the mailing list: jdt-dev@eclipse.org

I assume the topic will be ignored due to lack of resources, but maybe bringing it up on the mailing list has a better chance of getting attention and resources.

iloveeclipse:
> Andrey, do you plan on allocating resources for this? 

No

> Or do you know of any JDT maintainer who plans to do this?

No

oleosterhagen:
Hi @kriegaex,

do you have installed the _Language Server Protocol client for Eclipse IDE_ (`org.eclipse.lsp4e`)? If so, there may be conflicts for the key bindings `Shift+Ctrl+T` (JDT: Open Type / LSP4E: Go to Symbol in Workspace) and `Shift+Alt+R` (JDT: Rename - Refactoring / LSP4E: Rename). The chance for conflicts have been reduced recently with https://github.com/eclipse/lsp4e/pull/551 and https://github.com/eclipse/lsp4e/pull/564.

The flickering of the popup window in the case of a conflict was a general problem and has now been resolved with https://github.com/eclipse-platform/eclipse.platform.ui/pull/655.

kriegaex:
> do you have installed the _Language Server Protocol client for Eclipse IDE_ (`org.eclipse.lsp4e`)?

Yes, it looks like it. I never installed that on purpose. I have no idea where it comes from.

image

oleosterhagen:
There may be different ways which lead to installation of the LSP4E, e.g. _Eclipse IDE for Enterprise Java and Web Developers_ (JEE Package) includes _Wild Web Developer_ and this feature needs LSP4E.

PR #480, opened 2023-03-17
Create a workflow to check for merge commits in PR

laeubi:
@iloveeclipse I think we can just try it out here as you even has a test-case PR ...

### DIFF
--- a/.github/workflows/checkMergeCommits.yml
+++ b/.github/workflows/checkMergeCommits.yml
@@ -1,0 +1,18 @@
+on: pull_request
+
+name: Check for merge commits
+
+permissions:
+  pull-requests: read
+
+jobs:
+  message-check:
+    name: Block Merge Commits
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Block Merge Commits
+        uses: Morishiri/block-merge-commits-action@v2
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/checkMergeCommits.yml
+++ b/.github/workflows/checkMergeCommits.yml
@@ -13,6 +13,6 @@ jobs:
 
     steps:
       - name: Block Merge Commits
-        uses: Morishiri/block-merge-commits-action@v1
+        uses: Morishiri/block-merge-commits-action@a4554c78def8d874966a8d1e20e2971121443755
         with:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/checkMergeCommits.yml
+++ b/.github/workflows/checkMergeCommits.yml
@@ -13,6 +13,6 @@ jobs:
 
     steps:
       - name: Block Merge Commits
-        uses: Morishiri/block-merge-commits-action@v2
+        uses: Morishiri/block-merge-commits-action@v1
         with:
           repo-token: ${{ secrets.GITHUB_TOKEN }}


iloveeclipse:
@laeubi : are you done with this PR, or there are some more changes you would like to push?

laeubi:
Seems it runs now, so its fine to merge, Should I squash things or do you like to squash merge it?

iloveeclipse:
@laeubi : would you mind to push a dummy PR with merge commit for checking?

trancexpress:
According to @stephan-herrmann this was not necessary for JDT, see: https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1077#issuecomment-1585758946

> Why would merge commits ever appear in the final history if squash and merge is properly used? Given that direct merging without squash is not enabled for this repo I don't see what problem is being fixed here.

Original trigger was: https://github.com/eclipse-platform/eclipse.platform.releng/issues/214

Please discuss what is necessary and what isn't.

stephan-herrmann:
In my understanding git is much better able to make sense of arbitrarily complex merge graphs, than I am as a reviewer able to follow changes when everything is rebased time and again, mixing changes from all different sources into factually a new branch (of the same name), with no change history of its own (no the Compare link after a rebase is not sufficient).

I have positive experience with merge commits being eliminated from **gitlab**. I'd be much surprised if **github** is worse in this respect.

iloveeclipse:
Stephan, I don't think I can follow what is wrong with rebase and why diff on rebase should be worse as diff after multiple merges. In rebase one sees only own changes on top of master. Nothing is easier to understand as that.

trancexpress:
> Stephan, I don't think I can follow what is wrong with rebase and why diff on rebase should be worse as diff after multiple merges. In rebase one sees only own changes on top of master. Nothing is easier to understand as that.

I believe its the missing diffs between commits, as a complicated review progresses? I.e. patch sets in gerrit.

iloveeclipse:
In that case one can wait with rebase or squash till the end. No one is forced to rebase **immediately**, it is just a red flag for those who will do a final merge.

stephan-herrmann:
> Stephan, I don't think I can follow what is wrong with rebase and why diff on rebase should be worse as diff after multiple merges. In rebase one sees only own changes on top of master. Nothing is easier to understand as that.

Imagine Joe is working on issue A. Sometime during the development and review process, issue B is merged to master, which is found to trigger some unwanted interaction with issue A. Or maybe there's just a potential risk of that.

Now Joe wants to work in a workspace that combines issues A and B. When he does this by rebasing on master (incl. B), his previous work can no longer be effectively compared with any updates, because after rebase you will have two branches, which mainly differ by changes brought in from rebasing on master. Git can no longer help you to understand where a particular change originated.

Of course if you are *very* disciplined you **can** use rebase in a non-obstructive way, but the way I see people favoring rebase for *every* change, this would take a lot of educating.


IMHO, I shouldn't have to argue against rebase, but anybody wanting to invalidate public histories (in a pull request where contributor and reviewer collaborate) needs a very good excuse why he wants to re-write such history. When doing a review, I want to see each individual step, each change as a new commit, each merge from master as a merge commit, because all that is relevant information for reviewing. If that's too much information we have tools in abundance to abstract from that.

As you rightly say, cleaning up the history can happen after development and reviewing have completed. That's what squash and merge is all about, isn't it? Plus git provides lots of options for creating a custom history in a new PR etc. pp.

Perhaps I'm the only one where the review process takes several rounds and significant time to completion. If thus I'm the only one needing to see individual updates, perhaps I should go back to requesting patch files, which I can collect on my HD, and directly compare those patch files :) (seriously, that was a useful approach, "back then").

stephan-herrmann:
> In rebase one sees only own changes on top of master. Nothing is easier to understand as that.

to elaborate on that I'll have to expand this:

> Of course if you are very disciplined you can use rebase in a non-obstructive way

I would be mostly happy, if rebase were **only ever** used for syncing your branch with master. But I see plenty of modification of s.o.'s pending change via rebase as well. The result is: yes it's super easy to start a new review because you can easily see all accumulated changes in just one commit. But if the previous round of reviewing already cost significant efforts, I am clueless how to do incremental reviews in that model.  Thus reviewing time will be multiplied with the number of iterations. Seriously.

PR #481, opened 2023-03-17
Test merge action check

laeubi:
lets see what happens @iloveeclipse


laeubi:
grafik

trancexpress:
Looks great, thanks!

trancexpress:
Looks great, thanks! Can you mention using rebase instead of merging as part of the error message (that is seen when clicking "Details")?

laeubi:
grafik

trancexpress:
Great!

iloveeclipse:
@laeubi : great. Is there a way to run this action first and don't run jenkins build if action reports an error, to avoid that PRs block shared resources (nodes) with tests?

laeubi:
It could block Github runs probably but Jenkins would be a bit more complicated.

If this is really crucial, one could enhance the Jenkinsfile to perform such checks but it will require setup of credentials and a like so it is a bit bigger task, 

Maybe one can archive something similar with some git magic:
https://stackoverflow.com/questions/74096323/get-origin-master-and-pr-commit-id-in-jenkins

but maybe it does not occurs that much often and test-results are still valuable for a "mergecommited" one?

iloveeclipse:
Ok, I think we can close this PR now.

Issue #505, opened 2023-03-31
Contribute to "onboarding" extension in Java perspective


iloveeclipse:
See https://github.com/eclipse-platform/eclipse.platform.ui/issues/633 and changes to "Resources" perspective https://github.com/eclipse-platform/eclipse.platform.ui/pull/676.

It make sense to add exact same things what is contributed to "Resources" perspective **plus** "Open Type" command "Ctrl+Shift+T".

PR #507, opened 2023-03-31
Add Editor Area Onboarding for Java Perspective

BeckerWdf:
Fixes: #505


vogella:
Awesome. Thanks @BeckerWdf 

@iloveeclipse please review

org.eclipse.jdt.ui/plugin.xml
```diff
@@ -610,10 +610,25 @@
          <showInPart id="org.eclipse.jdt.ui.PackageExplorer"/>
       </perspectiveExtension>
 
-      <perspectiveExtension targetID="org.eclipse.jdt.ui.JavaPerspective">
+      <perspectiveExtension
+            editorOnboardingImage="$nl$/icons/full/onboarding_jperspective.png"
+            editorOnboardingText="perspectiveExtension.editorOnboardingText"
```
iloveeclipse, line 615: The "%" is missing

org.eclipse.jdt.ui/plugin.xml
```diff
@@ -610,10 +610,25 @@
...
+         </editorOnboardingCommand>
+         <editorOnboardingCommand
+               id="org.eclipse.ui.newWizard">
+         </editorOnboardingCommand>
+         <editorOnboardingCommand
+               id="org.eclipse.jdt.ui.navigate.open.type">
+         </editorOnboardingCommand>         
       </perspectiveExtension>
 
       <perspectiveExtension targetID="org.eclipse.jdt.ui.JavaBrowsingPerspective">
```
iloveeclipse, line 634: Two perspectives below would also profit from same addition.

iloveeclipse (changes requested):

org.eclipse.jdt.ui/plugin.xml
BeckerWdf: Good catch. I fixed it.

org.eclipse.jdt.ui/plugin.xml
BeckerWdf: I need to create the icons first. This will take some more time.

BeckerWdf:
Can someone on windows pls. also check if the icon also looks good on windows so that we don't run into the same issue as in https://github.com/eclipse-platform/eclipse.platform.ui/issues/687

org.eclipse.jdt.ui/plugin.xml
BeckerWdf: The different perspectives have different icons and so should the onboarding info.

iloveeclipse (approved):
LGTM.

Here is how it looks like on Linux / no themes:

image

iloveeclipse:
> Can someone on windows pls. also check if the icon also looks good on windows so that we don't run into the same issue as in [eclipse-platform/eclipse.platform.ui#687](https://github.com/eclipse-platform/eclipse.platform.ui/issues/687)

Looks almost same as on Linux (on Widows "Light" theme, with and without Eclipse themes).

org.eclipse.jdt.ui/plugin.xml
BeckerWdf: > Two perspectives below would also profit from same addition.

This is done now.

iloveeclipse (approved):
Thanks a lot. Looks good.

BeckerWdf:
> Thanks a lot. Looks good.

Can you post screenshots how it look in light and dark?

iloveeclipse:
> Can you post screenshots how it look in light and dark?

I don't use dark theme and won't mess up with my Gnome config to set it in the system.
But I will post the "light" version.

iloveeclipse:
image
image
image

BeckerWdf:
> I don't use dark theme and won't mess up with my Gnome config to set it in the system.

Ok. No problem.

> But I will post the "light" version.

They look nice

BeckerWdf:
From my point of view this can be merged.
Can somebody with jdt-ui committer rights merge?

vogella:
The inner white filling looks bad in the dark theme. @BeckerWdf could you do the same as in the Resource perspective and remove it?

image

BeckerWdf:
> you do the same as in the Resource perspective and remove it?

I am not sure. Because then it would not longer be a "solid" ball but only a circle with not fill - especially in the dark theme. Maybe I can play wit transparency and put in a fill which is almost transparent. By doing this. You would also have a fill in the dark theme but a much darker one.

BeckerWdf:
> The inner white filling looks bad in the dark theme. @BeckerWdf could you do the same as in the Resource perspective and remove it?
> 
> image

See https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/516

Issue #374, opened 2023-01-09
Remove $author from code template for class and module


Bananeweizen:
The default code templates for new type and new module (maybe others) contain a placeholder to insert the author of that new class/type/module. This was useful at times where many people did not have version control or a version control system that was not easy to use and query.

Today I see more negative than positive benefit:
* The same information is _better_ available in the version control history, with attribution by line instead of assuming this one author for every line currently in the file
* People copy/paste classes to create "something similar", without fixing this comment.
* It may leak information of the private user account to the public (at least on Windows, this defaults to your local user account id, and my company admin would surely not like me posting that account id everywhere).
* Even developers without an underlying version control system should not rely on this information, but rather be nudged to set up a real version control system.

I therefore want to suggest to remove this from all Javadoc-like comment templates (i.e. not from file header copyright banners, it does have benefit there).

Be aware this is not about the JDT or Eclipse platform source code, but about the templates provided by JDT.

erik-brangs:
I think this would be a good change. Personally, I've configured my templates to remove the author information because I just ended up deleting it every time.

PR #512, opened 2023-04-04
remove author from default code templates (#374)

Bananeweizen:
This was useful at times where many people did not have version control or a version control system that was not easy to use and query. Today that leads to lots of problems:

* The same information is better available in the version control history, with attribution by line instead of assuming this one author for every line currently in the file
* People copy/paste classes to create "something similar", without fixing this comment.
* It may leak information of the private user account to the public (at least on Windows, this defaults to your local user account id, and my company admin would surely not like me posting that account id everywhere).
* Even developers without an underlying version control system should not rely on this information, but rather be nudged to set up a real version control system.

## What it does
#374: remove the author tag from default code templates

## How to test
Create a new class (in a new workspace). Notice there is no more author tag in the Javadoc comment.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.ui/templates/default-codetemplates.xml
+++ b/org.eclipse.jdt.ui/templates/default-codetemplates.xml
@@ -35,8 +35,6 @@
  */</template>
 
 <template name="typecomment" id="org.eclipse.jdt.ui.text.codetemplates.typecomment" description="%CodeTemplates.typecomment" context="typecomment_context" enabled="true">/**
- * @author ${user}
- *
  * ${tags}
  */</template>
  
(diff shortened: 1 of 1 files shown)


Bananeweizen:
Be aware this has one effect that doesn't look nice: There are now two empty comments in modules:
grafik

However, the module-info creation code wants two comments there, so this seems to be very intentional: https://github.com/eclipse-jdt/eclipse.jdt.ui/blob/master/org.eclipse.jdt.ui/core%20extension/org/eclipse/jdt/internal/corext/util/InfoFilesUtil.java#L51 (notice the comment in line 64 clearly differentiates the 2 comments).

iloveeclipse (approved):
Thanks.

PR #516, opened 2023-04-05
Make Onboarding Icons smaller and look better in dark theme

BeckerWdf:


BeckerWdf:
@vogella and @iloveeclipse: Can you pls. test?

iloveeclipse:
Just noticed that browsing image contains some artifact that seem to be unneeded (was also before in the bigger image): 
image

iloveeclipse:
The PR title looks wrong: "Make Onboarding Icons look better in dark theme" - has nothing to do with dark theme, but with image size.

BeckerWdf:
> The PR title looks wrong: "Make Onboarding Icons look better in dark theme" - has nothing to do with dark theme, but with image size.

It also makes them better in dark theme - see the changes in the opacity of the fill color

iloveeclipse:
> So I also took it over to this icon.

On 32x32 pixel there are only 2 pixels visible, which probably was never noticed.
But on a bigger size picture this element in background simply looks like a graphical design artifact (which it is most likely is?).

iloveeclipse:
> > which it is most likely is?.
> 
> No it's not it's some kind of a hook. See here:
> 
> image

O. O-o. Well, not sure how good the resulting icon matches the original picture / idea of it, but for me in 300x300 size it looks like a bug :-)
@trancexpress : WDYT?

BeckerWdf:
> > > which it is most likely is?.
> > 
> > 
> > No it's not it's some kind of a hook. See here:
> > image
> 
> O. O-o. Well, not sure how good the resulting icon matches the original picture / idea of it, but for me in 300x300 size it looks like a bug :-) @trancexpress : WDYT?

See #524

PR #536, opened 2023-04-18
Enable Github Verification and cleanup build scripts

laeubi:
This do the follwoing:
- enable the unified github verification build
- removes obsolete workflows
- move mandatory maven options into the maven.config file

FYI @iloveeclipse


laeubi:
Some tests seem to show different offsets on `windows-latest` but the action itself works as desired.

iloveeclipse (changes requested):
OK, it happened what I've expected.
As long as github actions fail and reporting is as bad as it is, I would not merge this.

I personally have hard time to understand the "fail" output, and which build produced which "error" (which is not an error but test fail, but that's known issue).

I don't want / don't have time to explain every contributor on every PR that the various fails that we report are **to be ignored** (!!!), because later it causes that even experienced committer assume the PR is "safe" just because it fails **everywhere else**. 

See for example https://github.com/eclipse-platform/eclipse.platform.ui/pull/703#issuecomment-1511195592

=>

The way to proceed is to analyze fails and why they differ between maven / jenkins / official builds and to provide patches that would fix test fails or at least mute the known issues on github only.

Issue #528, opened 2023-04-14
JavaDoc hover spawns a WebKit process per open tab on Linux 


trancexpress:
To reproduce, open many Java editors and hover on code from each editor. Observe many WebKit processes are spawned and running (one per editor tab). E.g.:

```
sandreev   25004   23553 52 13:36 pts/2    00:00:55 /usr/lib/jvm/java-17-openjdk/bin/java -XX:+ShowCodeDetailsInExceptionMessages -agentlib:jdwp=transport=dt_socket,suspend=y,address=localhost:41765 -Dorg.eclipse.swt.graphics.Resource.reportNonDisposed=true -Declipse.pde.launch=true --add-modules=ALL-SYSTEM -Djava.security.manager=allow -Dfile.encoding=UTF-8 -classpath /data/eclipses/eclipseSDK_428_m20230404/plugins/org.eclipse.equinox.launcher_1.6.400.v20210924-0641.jar org.eclipse.equinox.launcher.Main -launcher /data/eclipses/eclipseSDK_428_m20230404/eclipse -name Eclipse -showsplash 600 -product org.eclipse.platform.ide -data /data/workspaces/contributor_workspace/../runtime-Eclipse -configuration file:/data/workspaces/contributor_workspace/.metadata/.plugins/org.eclipse.pde.core/Eclipse/ -dev file:/data/workspaces/contributor_workspace/.metadata/.plugins/org.eclipse.pde.core/Eclipse/dev.properties -os linux -ws gtk -arch x86_64 -nl en_US -consoleLog
sandreev   25303   25004  0 13:37 pts/2    00:00:00 /usr/lib/webkit2gtk-4.0/WebKitNetworkProcess 8 129
sandreev   25309   25004  0 13:37 pts/2    00:00:00 /usr/lib/webkit2gtk-4.0/WebKitWebProcess 12 129
sandreev   25342   25004  0 13:37 pts/2    00:00:00 /usr/lib/webkit2gtk-4.0/WebKitWebProcess 20 145
sandreev   25374   25004  0 13:37 pts/2    00:00:00 /usr/lib/webkit2gtk-4.0/WebKitWebProcess 27 154
sandreev   25407   25004  0 13:37 pts/2    00:00:00 /usr/lib/webkit2gtk-4.0/WebKitWebProcess 34 157
[...]

PR #548, opened 2023-04-25
Dispose javadoc hover widgets after delay

iloveeclipse:
The hovers shown for content assist **are** immediately disposed, and that seem to be OK. 
So why not the editor hovers?
The idea is just to dispose the controls once they are hidden for at least 5 seconds. This seem to work for me, but I haven't checked why we actually don't do this by default.

Fixes https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/528

### DIFF
--- a/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/text/java/hover/JavadocHover.java
+++ b/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/text/java/hover/JavadocHover.java
@@ -66,6 +66,7 @@ import org.eclipse.jface.text.TextPresentation;
 import org.eclipse.ui.IEditorPart;
 import org.eclipse.ui.ISharedImages;
 import org.eclipse.ui.IWorkbenchPage;
+import org.eclipse.ui.IWorkbenchPreferenceConstants;
 import org.eclipse.ui.IWorkbenchSite;
 import org.eclipse.ui.PartInitException;
 import org.eclipse.ui.PlatformUI;
@@ -298,6 +299,15 @@ public class JavadocHover extends AbstractJavaEditorTextHover {
 		}
 	}
 
+	/**
+	 * @return timeout that we want to have on javadoc hover to avoid two browser instances hanging
+	 *         around for every opened java editor
+	 */
+	private static int getDisposeHoverTimeout() {
+		String key = IWorkbenchPreferenceConstants.DISPOSE_CLOSED_BROWSER_HOVER_TIMEOUT;
+		int timeout = Platform.getPreferencesService().getInt("org.eclipse.ui", key, -1, null);  //$NON-NLS-1$
+		return timeout;
+	}
 
 	/**
 	 * Presenter control creator.
@@ -318,16 +328,13 @@ public class JavadocHover extends AbstractJavaEditorTextHover {
 			fSite= site;
 		}
 
-		/*
-		 * @see org.eclipse.jdt.internal.ui.text.java.hover.AbstractReusableInformationControlCreator#doCreateInformationControl(org.eclipse.swt.widgets.Shell)
-		 */
 		@Override
 		public IInformationControl doCreateInformationControl(Shell parent) {
 			if (BrowserInformationControl.isAvailable(parent)) {
 				ToolBarManager tbm= new ToolBarManager(SWT.FLAT);
 				String font= PreferenceConstants.APPEARANCE_JAVADOC_FONT;
 				BrowserInformationControl iControl= new BrowserInformationControl(parent, font, tbm);
-
+				iControl.setDisposeTimeout(getDisposeHoverTimeout());
 				final BackAction backAction= new BackAction(iControl);
 				backAction.setEnabled(false);
 				tbm.add(backAction);
@@ -423,15 +430,13 @@ public class JavadocHover extends AbstractJavaEditorTextHover {
 			if (BrowserInformationControl.isAvailable(parent)) {
 				String font= PreferenceConstants.APPEARANCE_JAVADOC_FONT;
 				iControl= new BrowserInformationControl(parent, font, tooltipAffordanceString) {
-					/*
-					 * @see org.eclipse.jface.text.IInformationControlExtension5#getInformationPresenterControlCreator()
-					 */
+
 					@Override
 					public IInformationControlCreator getInformationPresenterControlCreator() {
 						return fInformationPresenterControlCreator;
 					}
 				};
-
+				iControl.setDisposeTimeout(getDisposeHoverTimeout());
 				JFaceResources.getColorRegistry().addListener(this); // So propertyChange() method is triggered in context of IPropertyChangeListener
 				setHoverColors();
 


cdietrich:
TODO: check if this needs to be transposed to Xtext too

trancexpress:
Should we do this for Linux only? Since the problem is only with GTK+ and WebKit.

From my POV maybe 10 seconds instead of 5, but other than that I don't see a problem. Maybe the hover won't come up as fast as it normally does, for a lot of JavaDoc contents. And maybe we want to do this if there is a specific number of hovers already?

iloveeclipse:
- Simeon, if you have Windows, could you check how it works there (both problem and solution)?
- 5 or 10 seconds is not a big deal, we may even have 30.
- What I haven't checked is: where in this complex framework that is **reusing** the browser we could do same as proposed here, so for example Xtext could be "fixed" too (or anyone else)? **If** that would be possible (with some clean code), we could add some preference to close the browser instances after some timeout and enable that for Linux only (or even in specific product only)?

trancexpress:
> * Simeon, if you have Windows, could you check how it works there (both problem and solution)?

Not anymore, I only have Linux at this point.

>     * 5 or 10 seconds is not a big deal, we may even have 30.

30 seconds might be even better yes.

>     * What I haven't checked is: where in this complex framework that is **reusing** the browser we could do same as proposed here, so for example Xtext could be "fixed" too (or anyone else)? **If** that would be possible (with some clean code), we could add some preference to close the browser instances after some timeout and enable that for Linux only (or even in specific product only)?

Hard to say, maybe somehow in `AbstractReusableInformationControlCreator.createInformationControl(Shell)`? Or maybe `BrowserInformationControl` can get a flag "dispose after N seconds when closed".

iloveeclipse:
> maybe somehow in AbstractReusableInformationControlCreator.createInformationControl(Shell)?

`AbstractReusableInformationControlCreator` seem to be too generic, implementations don't always use browser.

> maybe BrowserInformationControl can get a flag "dispose after N seconds when closed".

This one seem to be a better choice. If that can have a flag & check it, it should be doable.
The concrete creators (like JDT / Xtext/ whoever) could define what they want looking on some preference in IDE.
We don't even need to expose this preference to user, and keep it totally internal / for products.

trancexpress:
> This one seem to be a better choice. If that can have a flag & check it, it should be doable. The concrete creators (like JDT / Xtext/ whoever) could define what they want looking on some preference in IDE. We don't even need to expose this preference to user, and keep it totally internal / for products.

Something like this in JFace then?

```
diff --git a/org.eclipse.jface.text/src/org/eclipse/jface/internal/text/html/BrowserInformationControl.java b/org.eclipse.jface.text/src/org/eclipse/jface/internal/text/html/BrowserInformationControl.java
index b3640458e..a14172217 100644
--- a/org.eclipse.jface.text/src/org/eclipse/jface/internal/text/html/BrowserInformationControl.java
+++ b/org.eclipse.jface.text/src/org/eclipse/jface/internal/text/html/BrowserInformationControl.java
@@ -169,6 +169,7 @@ public class BrowserInformationControl extends AbstractInformationControl implem
         */
        private final String fSymbolicFontName;
 
+       private int fDisposeTimeout= 0;
 
        /**
         * Creates a browser information control with the given shell as parent.
@@ -335,6 +336,7 @@ public class BrowserInformationControl extends AbstractInformationControl implem
                if (!visible) {
                        super.setVisible(false);
                        setInput(null);
+                       startDisposeTimeout(shell);
                        return;
                }
 
@@ -596,4 +598,23 @@ public class BrowserInformationControl extends AbstractInformationControl implem
                return new Point((int) (widthInChars * width), heightInChars * height);
        }
 
+       /**
+        * Sets a timeout, after which a not visible control will be disposed.
+        *
+        * @param disposeTimeout The timeout in milliseconds. Non-positive values result in no timeout,
+        *            i.e. don't dispose non-visible controls.
+        */
+       public void setDisposeTimeout(int disposeTimeout) {
+               fDisposeTimeout= disposeTimeout;
+       }
+
+       private void startDisposeTimeout(Shell shell) {
+               if (fDisposeTimeout > 0) {
+                       shell.getDisplay().timerExec(fDisposeTimeout, () -> {
+                               if (!isVisible()) {
+                                       dispose();
+                               }
+                       });
+               }
+       }
 }
```

And then JDT needs to set this timeout. We can also define JFace preference for it?

iloveeclipse:
Yes, something like this. 
Notes: 
- ideally the timeout should be set only once (assuming a series of set visible on/off).
- jface has no preferences, so it should be some IDE bundle preference (workbench).

trancexpress:
> Yes, something like this. Notes:
> 
>     * ideally the timeout should be set only once (assuming a series of set visible on/off).

Like this?

```
diff --git a/org.eclipse.jface.text/src/org/eclipse/jface/internal/text/html/BrowserInformationControl.java b/org.eclipse.jface.text/src/org/eclipse/jface/internal/text/html/BrowserInformationControl.java
index b3640458e..352748248 100644
--- a/org.eclipse.jface.text/src/org/eclipse/jface/internal/text/html/BrowserInformationControl.java
+++ b/org.eclipse.jface.text/src/org/eclipse/jface/internal/text/html/BrowserInformationControl.java
@@ -169,6 +169,9 @@ public class BrowserInformationControl extends AbstractInformationControl implem
         */
        private final String fSymbolicFontName;
 
+       private int fDisposeTimeout= 0;
+
+       private Runnable fDisposeTask= null;
 
        /**
         * Creates a browser information control with the given shell as parent.
@@ -335,6 +338,7 @@ public class BrowserInformationControl extends AbstractInformationControl implem
                if (!visible) {
                        super.setVisible(false);
                        setInput(null);
+                       startDisposeTimeout(shell.getDisplay());
                        return;
                }
 
@@ -596,4 +600,29 @@ public class BrowserInformationControl extends AbstractInformationControl implem
                return new Point((int) (widthInChars * width), heightInChars * height);
        }
 
+       /**
+        * Sets a timeout, after which a not visible control will be disposed.
+        *
+        * @param disposeTimeout The timeout in milliseconds. Non-positive values result in no timeout,
+        *            i.e. don't dispose non-visible controls.
+        */
+       public void setDisposeTimeout(int disposeTimeout) {
+               fDisposeTimeout= disposeTimeout;
+       }
+
+       private void startDisposeTimeout(Display display) {
+               class DisposeTask implements Runnable {
+                       @Override
+                       public void run() {
+                               fDisposeTask= null;
+                               if (!isVisible()) {
+                                       dispose();
+                               }
+                       }
+               }
+               if (fDisposeTimeout > 0 && fDisposeTask == null) {
+                       fDisposeTask= new DisposeTask();
+                       display.timerExec(fDisposeTimeout, fDisposeTask);
+               }
+       }
 }
```

>     * jface has no preferences, so it should be some IDE bundle preference (workbench).

I'm not so sure about the preference, at least if integrating it in JFace. IMO whoever creates `BrowserInformationControl` should tell it to dispose itself after being closed for N seconds. Since its possible some client code relies on the `BrowserInformationControl` not getting disposed after being closed (i.e. on visibility changes). In the case of the JavaDoc hover, `AbstractReusableInformationControlCreator` ensures the recreation of the control.

I guess we can add a preference and  let `AbstractReusableInformationControlCreator` use it to make sure the control is disposed after staying closed for a long enough time?

iloveeclipse:
> I'm not so sure about the preference, at least if integrating it in JFace. IMO whoever creates `BrowserInformationControl` should tell it to dispose itself after being closed for N seconds.

Exact. As I've said, jface has no preferences. The preference can be set at some workbench bundle, and the creator(s) of the browser control can check that preference.

trancexpress:
> As I've said, jface has no preferences. The preference can be set at some workbench bundle, and the creator(s) of the browser control can check that preference.

OK, something like this?

```
diff --git a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/IPreferenceConstants.java b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/IPreferenceConstants.java
index 807f461ca3..e6c12f4d0a 100644
--- a/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/IPreferenceConstants.java
+++ b/bundles/org.eclipse.ui.workbench/Eclipse UI/org/eclipse/ui/internal/IPreferenceConstants.java
@@ -325,4 +325,15 @@ public interface IPreferenceConstants {
         */
        String SHOW_KEYS_TIME_TO_CLOSE = "showCommandKeys_timeToClose"; //$NON-NLS-1$
 
+       /**
+        * Preference for the time (in ms) after a closed browser hover is to be
+        * disposed
+        * <p>
+        * The integer default value for this preference is: <code>-1</code>.
+        * Non-positive values indicate closed hovers are not disposed automatically.
+        * </p>
+        *
+        * @since 3.129
+        */
+       String DISPOSE_CLOSED_BROWSER_HOVER_TIMEOUT = "disposeClosedBrowserHoverTimeout"; //$NON-NLS-1$
 }
```

We'll need PRs in different platform repos.

iloveeclipse:
Yes.

trancexpress:
OK.

For the JFace change see: https://github.com/eclipse-platform/eclipse.platform.text/pull/193
For the workbench UI change see: https://github.com/eclipse-platform/eclipse.platform.ui/pull/743

I'm not sure if the change is overall a good one, @cdietrich can you check if those 2 changes help your case?

cdietrich:
at a first test it seems to work for 
XtextBrowserInformationControl
unfortunately we also have a copy&paste adaptation in xtext 
XbaseInformationControl (back then / still?!? too much was nailed down for pot. subclasses)
need to check how i can test this one

cdietrich:
created https://github.com/eclipse/xtext/issues/2653 for adoption in Xtext

iloveeclipse:
I've pushed nearly final patch using (updated) constant from https://github.com/eclipse-platform/eclipse.platform.ui/pull/743 and https://github.com/eclipse-platform/eclipse.platform.text/pull/193.
I can confirm it works.

trancexpress:
How do we access the new preference? `WorkbenchPlugin.getDefault().getPreferenceStore()` seems to not be allowed in JDT UI (API restriction).

And `PlatformUI.getPreferenceStore()` doesn't seem to see the preference, when setting it via plug-in customization:
```
org.eclipse.ui.workbench/disposeClosedBrowserHoverTimeout=5000 
```
This also doesn't work:
```
org.eclipse.ui/disposeClosedBrowserHoverTimeout=5000 
```

iloveeclipse:
Simeon, please simply try this PR.

trancexpress:
> And **do not** overwrite existing code that works with your changes that don't work :-)

Sorry, I didn't see the PR was updated meanwhile.

org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/text/java/hover/JavadocHover.java
```diff
@@ -298,6 +299,15 @@ public void run() {
 		}
 	}
 
+	/**
+	 * @return timeout that we want to have on javadoc hover to avoid two browser instances hanging
+	 *         around for every opened java editor
+	 */
+	private static int getDisposeHoverTimeout() {
+		String key = IWorkbenchPreferenceConstants.DISPOSE_CLOSED_BROWSER_HOVER_TIMEOUT;
+		int timeout = Platform.getPreferencesService().getInt("org.eclipse.ui", key, 10, null);  //$NON-NLS-1$
```
iloveeclipse, line 308: "10" is left from debugging. has to be changed to -1.

iloveeclipse: `org.eclipse.ui/disposeClosedBrowserHoverTimeout=1000`

trancexpress: Maybe I had the `-pluginCustomization` argument wrong, now I see its working. Will you be changing the default or should I?

iloveeclipse:
Fail is known on master, 
JdtTextTestSuite org.eclipse.jdt.text.tests.contentassist.ContentAssistTestSuite org.eclipse.jdt.text.tests.contentassist.CodeCompletionTest.testAnonymousTypeCompletion6

https://download.eclipse.org/eclipse/downloads/drops4/I20230505-1800/testresults/html/org.eclipse.jdt.text.tests_ep428I-unit-cen64-gtk3-java17_linux.gtk.x86_64_17.html

Issue #571, opened 2023-04-04
Call Hierarchy for Record components does not work


JanecekPetr:
Using eclipse 2023-03 and Java 17 on an M1 Pro Mac with the provided JustJ 17:

1. The Call Hierarchy only works for Record components _when the record definition is in the same file as the caller_. Otherwise it does not show up at all.

    Different files:
![image](https://user-images.githubusercontent.com/1746404/229822892-b53aa604-9cb0-4ce0-ba24-d4223b832f46.png)

    Same file (it does not matter if the record is nested or a sibling, it shows up every time):
![image](https://user-images.githubusercontent.com/1746404/229823537-b87de155-a46b-4d99-b598-e69edbc5253a.png)

2. (Edit: Separated into an another ticket.) ~~Currently the Outline view does not show the record component accessor methods. I tried unticking the Synthetic members, does not make a difference. This is annoying because normally I have fields filtered out and only see methods which is now not possible for records.~~
    ~~Also I thought using Call Hierarchy on the methods might help, but they're not available at all. Maybe this is somehow related? I imagine the component accessors are special methods which circumvent the normal eclipse way of tracking things?~~

If these two things are completely unrelated, I'm happy to split this ticket into two.

trancexpress:
> 2\. the private field is final and has matching public synthetic member

I'm not sure how to do this. And which case it should cover?

PR #572, opened 2023-05-08
Call Hierarchy for Record components does not work

trancexpress:
Adjust `CallerMethodWrapper.getAccurateSearchScope()` to consider also record components, when checking if a member is private. This avoids reducing the search scope to just the compilation unit of the record defining the component, and so results in the expected search results.

Fixes: #571

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/callhierarchy/CallerMethodWrapper.java
+++ b/org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/callhierarchy/CallerMethodWrapper.java
@@ -163,7 +163,7 @@ public class CallerMethodWrapper extends MethodWrapper {
 	}
 
 	private IJavaSearchScope getAccurateSearchScope(IJavaSearchScope defaultSearchScope, IMember member) throws JavaModelException {
-		if (! JdtFlags.isPrivate(member))
+		if (!JdtFlags.isPrivate(member) || isRecordComponent(member))
 			return defaultSearchScope;
 
 		if (member.getCompilationUnit() != null) {
@@ -213,4 +213,7 @@ public class CallerMethodWrapper extends MethodWrapper {
 		return fIsExpandWithConstructorsSet;
 	}
 
+	private static boolean isRecordComponent(IMember member) throws JavaModelException {
+		return member.getElementType() == IJavaElement.FIELD && member instanceof IField f && f.isRecordComponent();
+	}
 }
--- a/org.eclipse.jdt.ui.tests/ui/org/eclipse/jdt/ui/tests/callhierarchy/CallHierarchyTestHelper.java
+++ b/org.eclipse.jdt.ui.tests/ui/org/eclipse/jdt/ui/tests/callhierarchy/CallHierarchyTestHelper.java
@@ -381,6 +381,52 @@ public class CallHierarchyTestHelper {
         assertBuildWithoutErrors(fJavaProject3);
     }
 
+	/**
+	 * Creates a record class and a type that references a field of the record.
+	 */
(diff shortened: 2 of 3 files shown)


org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/callhierarchy/CallerMethodWrapper.java
```diff
@@ -213,4 +213,15 @@ public boolean isExpandWithConstructorsSet() {
 		return fIsExpandWithConstructorsSet;
 	}
 
+	private static boolean isRecordComponent(IMember member) throws JavaModelException {
+		if (member.getElementType() == IJavaElement.FIELD && member instanceof IField f) {
+			if (f.isRecordComponent()) {
+				return true;
+			}
+			if (f.getParent() instanceof IType p && p.isRecord()) {
```
trancexpress, line 221: For whatever reason, the `SourceType` that is passed to `CallerMethodWrapper.getAccurateSearchScope()` returns `false` for `isRecordComponent()`. When using the reproducer from #571. So we need this second part here.

When selecting the record component (e.g. `wheels` in `Car.java` from the reproducer) `isRecordComponent()` returns `true` as expected.

Unfortunately I don't know how to test with the reproducer from #571. I would need to get into the method body of the method (as done in the IDE), select the method and compute the call hierarchy from there.

Possible, but I'm not sure if its worth the effort. Especially if we have a JDT core bug to fix for `isRecordComponent()`. WDYM @iloveeclipse ?

iloveeclipse: Simeon, please create JDT bug, this soukds really suspicious.

trancexpress: Created: https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1035

iloveeclipse: I have a fix for JDT core, will push soon. Just found out that RecordsElementTests were never executed...

iloveeclipse: Simeon, you can remove the workaround now. JDT core fix is merged.

iloveeclipse:
The tests will fail now (expected) and should pass after next SDK build.

trancexpress:
> The tests will fail now (expected) and should pass after next SDK build.

See my comment, I'm not sure it will. The added test is not the same as the reproducer, since the search is not started from the correct element:

> Unfortunately I don't know how to test with the reproducer from https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/571. I would need to get into the method body of the method (as done in the IDE), select the method and compute the call hierarchy from there.

iloveeclipse:
A, OK, haven't run your test, just updated one in JDT core that would fail without fix.

trancexpress:
> A, OK, haven't run your test, just updated one in JDT core that would fail without fix.

Will that JDT core test be enough then? Or should I look for some way to test with selecting the method in `Foo` (from the reproducer)?

iloveeclipse:
> > A, OK, haven't run your test, just updated one in JDT core that would fail without fix.
> 
> Will that JDT core test be enough then?

Should be enough, see https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1059

iloveeclipse (approved):

PR #593, opened 2023-05-24
Fix regression in RenameLinkedMode

### LINKED ISSUE #592: Refactor renames random java elements
Very nasty regression (random results - but reproducible):
ALT+SHIFT+R on a method sometimes also selects and refactors code that has nothing todo with it.
Example:
image

I wanted to refactor the name of "waitFor" but in the same turn it refactored "lockAcquiringRunnable"
image
When i slightly change the code then either problem is gone or other elements are refactored:
image

You can download the the example from https://github.com/jukzi/eclipse.platform/commits/refactorBug
```
git clone git@github.com:jukzi/eclipse.platform.git
git checkout refactorBug

```
Then import the `project eclipse.platform\runtime\tests\org.eclipse.core.tests.runtime` to your workspace.
It will have many compilation errors if you do not have the dependencies but that does not affect the bug.

Sorry, that i could not find a smaller example .. as soon as i reduce it i can't reproduce.

Feels like there is a comparison of a colliding hash instead of checking for equals.

jjohnstn:
- add verification that iMethod root is same as root we are using NodeFinder on
- fixes #592

## What it does

Fixes regression.  See issue.

## How to test

See issue.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.ui/ui refactoring/org/eclipse/jdt/internal/ui/refactoring/reorg/RenameLinkedMode.java
+++ b/org.eclipse.jdt.ui/ui refactoring/org/eclipse/jdt/internal/ui/refactoring/reorg/RenameLinkedMode.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2005, 2022 IBM Corporation and others.
+ * Copyright (c) 2005, 2023 IBM Corporation and others.
  *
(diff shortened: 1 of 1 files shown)


jukzi (approved):
Change LGTM, works (manualy tested), and does not influence code other then the rename usecase.

iloveeclipse (approved):
LGTM

PR #617, opened 2023-06-06
Use typed for-each loop

jukzi:
Avoiding type-casts makes code easier to read.


jjohnstn (approved):
Looks good when master is opened for 4.29 M1.

Issue #631, opened 2023-06-13
Eclipse becomes unresponsive when trying to refresh a viewer with more than 30000 elements on one level


raghucssit:
This issue is to consume the solution provided for platform ui issue https://github.com/eclipse-platform/eclipse.platform.ui/issues/818
This is still a draft stage. Platform UI fix is still in progress.

PR #632, opened 2023-06-13
Use limit based viewers #631

raghucssit:
**This PR won't compile unless you have changes from [810](https://github.com/eclipse-platform/eclipse.platform.ui/pull/810) on your workspace**

## What it does

This will limit the number of elements populated in Package Explorer, Outline view for Java editor and editor breadcrumb viewer to the common value defined by the workbench.

## How to test

See https://github.com/eclipse-platform/eclipse.platform.ui/pull/810 for detailed theory of operation and use case examples.

### DIFF
--- a/org.eclipse.jdt.junit/src/org/eclipse/jdt/internal/junit/ui/TestSessionLabelProvider.java
+++ b/org.eclipse.jdt.junit/src/org/eclipse/jdt/internal/junit/ui/TestSessionLabelProvider.java
@@ -186,9 +186,8 @@ public class TestSessionLabelProvider extends LabelProvider implements IStyledLa
 			else
 				throw new IllegalStateException(element.toString());
 
-		} else {
-			throw new IllegalArgumentException(String.valueOf(element));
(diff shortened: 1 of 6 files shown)


raghucssit:
@iloveeclipse FYI. This is a draft PR for consuming changes from [810](https://github.com/eclipse-platform/eclipse.platform.ui/pull/810)

iloveeclipse:
Raghu, I've played with the patch on Windows and I see that "Quick Method Outline" <Ctrl+O> in Java Editor simply doesn't open with 18.000 methods (it hangs, blinks and disappears immediately after opening shell). That must be also fixed with this patch. Beside this, on Windows / my notebook the limit of 1000 is better as 5000, because on Windows it also lags even with 5000 nodes limit.

I think we can set default limit to 1000 for SDK and change to 5000 or 10000 in our product customization, or may be keep with 1000.

iloveeclipse:
Thanks Raghu, Ctrl+O works now.

PR #628, opened 2023-06-12
Add name recommendation function for renaming field. 

DongChunhao:
- modify RenameResourceHandler.java to generate the new names for renaming fields and to display the new name in the wizard.
- fixes #627 

## What it does

Recommend the field that complies with constant naming conventions as the suggested new name.
## How to test

Run “Refactor-Rename”
## Author checklist

- [x] I have thoroughly tested my changes
- [x] The change is following the [coding conventions](https://wiki.eclipse.org/Platform/How_to_Contribute#Coding_Conventions)
- [x] I have signed the [Eclipse Contributor Agreement (ECA)](https://www.eclipse.org/legal/ECA.php)

Issue #627, opened 2023-06-12
Add name recommendation function for renaming field.


DongChunhao:
> **Renaming is one of the most frequently used features in IDEs (like Eclipse). However, we notice that their recommendation of new names during renames could be further improved. Here we present this issue to specify how we can improve the name recommendation for Rename Field refactorings.**

- The following code snippet comes the well-known project OpenJDK.  Line 44 of jdk/src/java.desktop/share/classes/sun/ java2d /marlin/Dasher.java. In this example, the original developers renamed the field "recLimit" to a new name "REC_LIMIT" because static and fine fields should follow the well-known ConstantNamingConvention (i.e., capitalizing all characters and connecting words with ‘_’).

![image](https://github.com/eclipse-jdt/eclipse.jdt.ui/assets/77560508/409b56c6-66f9-4d34-828d-45e41b7509be)

- When the developers select the field "recLimit" and click the refactoring menu "Rename" in Eclipse, Eclipse will present the following refactoring wizard where the old field name "recLimit" appears in the wizard as the default new name. 

![image](https://github.com/eclipse-jdt/eclipse.jdt.ui/assets/77560508/65331683-94eb-4202-827b-4e3ed451607f)

- A better choice here is to recommend the field that complies with constant naming conventions as the suggested new name. As a generic rule, we propose the following heuristics:

     If (f.modifiers CONTAINS ‘Final’ ) and ( f.modifiers CONTAINS ‘Static’ ) and (f.name NOTFOLLOW ConstantNamingConvention) :
                                f.newName= Conver2ConstantNamingConvention(f.name)

- We have validated the heuristic on 10574 field renamings actually conducted by the original developers in open-source applications. In total, the heuristic recommend 1073 names where 922 are exactly the same as what the original developers finally used. That is , the precision is 86%.

- We also submit the pull request to resolve this issue. When renaming field names that violate the constant naming convention, we modify the original field name to comply with the constant naming conventions and suggest it as the new name, and display it in the rename wizard.

liuhuigmail:
@jjohnstn  We have submitted here a minor suggestion on how to faciliate renaming refactoring in Eclipse. Please kindly have a look on the issue as well as the associated PR. 

Thanks.

jjohnstn:
@liuhuigmail will take a look at it

liuhuigmail:
Thanks @jjohnstn

PR #654, opened 2023-07-04
Add name recommendation function for renaming field

DongChunHao:
- modify RenameResourceHandler.java to generate the new names for renaming fields and to display the new name in the wizard.

- fixes：[#628 Add name recommendation function for renaming field.](https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/627)
## What it does

Recommend the field that complies with constant naming conventions as the suggested new name.
## How to test

- Run “Refactor-Rename”

## Author checklist

### DIFF
--- a/org.eclipse.jdt.ui/ui refactoring/org/eclipse/jdt/internal/ui/refactoring/reorg/RenameRefactoringWizard.java
+++ b/org.eclipse.jdt.ui/ui refactoring/org/eclipse/jdt/internal/ui/refactoring/reorg/RenameRefactoringWizard.java
@@ -14,13 +14,41 @@
 
 package org.eclipse.jdt.internal.ui.refactoring.reorg;
 
+import java.io.BufferedInputStream;
+import java.io.FileInputStream;
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.util.AbstractList;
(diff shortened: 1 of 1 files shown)


DongChunhao:
@jjohnstn Please kindly have a look on the resubmitted PR.

jjohnstn:
Hi @DongChunhao.  I don't think this is a good fix for the problem.  It creates an inconsistent rename interface.  Right now, if I choose to rename the field, I actually edit in place the current value with the current name being selected.  If I want options, I click on Options...  Your change now makes it so when I choose options I have a new name presented in its place which is inconsistent.

I think for this to be viable it needs to be separate from generic Rename...  One suggestion is a new Quick Assist (e.g. Use constant naming syntax).  This quick assist would show up for static final constant fields where the name doesn't follow syntax AND the corrected name isn't already in use.  A user can set it up so the editor will mark lines that have a quick assist.  With this, a user can simply CTRL+1 on the constant name or if marked, click on the light-bulb icon and select.  This quick assist could simply bring up the Rename dialog as you currently have it with the name changed or you could choose to do the rename without asking for further input if you find that the majority of cases wouldn't use the additional options.  If users ask for it, you could then add a second quick assist later which would also open the Rename options dialog.

Issue #655, opened 2023-07-04
Context-based Automated Name Recommendation For The Extract Local Variable Refactoring


Michaelll123:
### Context-based Automated Name Recommendation For The Extract Local Variable Refactoring
---
**Description**
---
We propose a context-based rule to automatically recommend names for the extracted local variables. We believe that this rule can make the existing recommendation strategy of Eclipse more accurate and effective. 
Here we first post an example to illustrate how this rule works. This example is taken from the commit of Ignite of Apache (914ca301b6520d49d4f06784594f4c75bf215e65, apache_ignite/modules/configuration/src/main/java/org/apache/ignite/internal/configuration/tree/NamedListNode.java)
```java
/* CS1: Real Case of Extract Local Variable Refactoring */
    public NamedListChange<N, N> delete(String key) {
        Objects.requireNonNull(key, "key");

-       if (map.containsKey(key)) {
-           map.get(key).value = null;
+       ElementDescriptor element = map.get(key);
+
+       if (element != null) {
+           element.value = null;
       }

        return this;
    }
```
The actual name developer assigned to the extracted expression, i.e., `map.get(key)` (line 7 in CS1) is "element". 
However, if we select the expression `map.get(key)` and conduct refactoring "extract local variable" in Eclipse, we will get the following code snippet:
```java
/* CS2: Code Snippet After Refactoring in Eclipse*/
    public NamedListChange<N, N> delete(String key) {
        Objects.requireNonNull(key, "key");

        ElementDescriptor elementDescriptor = map.get(key);

        if (elementDescriptor != null) {
            elementDescriptor.value = null;
        }

        return this;
    }
```

The name recommended for the extracted variable is "elementDescriptor", which is substantially different from what the developers manually coined. 

**What We Can Do**
---
For the posted example,  we notice that there are quite a few (five in total) variable declarations with the same assignment expression (i.e., "map.get(key)") in other methods of the same Java file.
We also notice that all such variables share the same variable name "element". Consequently, it is natural to recommend "element" as the name for the extracted expression "map.get(key)".

Here is the related context code:
```java
public final class NamedListNode<N> implements NamedListChange<N, N>, TraversableTreeNode, ConstructableTreeNode {
	
	……
	
    private final OrderedMap<ElementDescriptor> map;

	……
	
    /**
     * Returns {@link InnerNode} associated with the passed key.
     *
     * @param key Key string.
     * @return Requested value.
     */
    @Nullable
[...]

PR #656, opened 2023-07-04
context-based-name-recommendation

Michaelll123:
## What it does
- add a new context-based rule to recommend names for extracted variables.
- fix issue #655.

## How to test
Run test157, and test158 in ExtractTempTests.

## Author checklist


jjohnstn (changes requested):
Hi @Michaelll123, thank you for the contribution and while it does work as you intend, I don't think that the results justify the amount of work being performed.  You are looking to see if other variables initialize with the same expression being extracted and then are checking to see if over 70% use the same variable name.  However, the variable name isn't that important to warrant this work because at the time it is extracted, nothing else uses it but the one expression it was extracted from.  The user has more knowledge on what the intended use is in the future and will either accept the calculated name or will type in the name.  If the user is repeating a pattern found in other methods which your code seems to be looking for, it is highly likely the pattern code will already be copied including the temp variable name.  In one of your tests, you end up choosing "element" over "person" because it is used multiple times.  IMO, "person" was the better choice in all cases and would have been offered without searching the file.

In your test for expression equality, you are using String equals() which won't always be correct.  The toString() method is not intended to be used this way and is simply for debugging purposes.  The actual way to do it would be to use subtreeMatch() which is even more expense.

liuhuigmail:
@jjohnstn Thanks for quick response and valuable suggestions.  I agree with you that the developers often "has more knowledge on what the intended use is in the future and will either accept the calculated name or will type in the name."  

We still believe that it  is valuable to improve the accuracy in recommendation of variable names for extract local variable refactorings.  In case the recommendation is correct (that is, exactly as what the users want), users can simply accept it without any additional edition.  This is somewhat like automated code complete where users known exactly what to type next but still want automated suggestions.  

Please note that the evaluation data are real-world [extract local variable] refactorings mined from open-source applications, and thus the variable names had been actually cointed by the users.  However, I agree with you that the given example (suggesting "element" instead of "person") is bad. 

Please kindly let us known wheter it is fine to revise the code as you suggested and replace the bad example.  

Thanks.

Michaelll123:
Thanks for the valuable suggestions! 
- All the requested changes were addressed.
- I further made some modifications to make the changes only work for the Extract Local Variable Refactoring, which avoids impacting other refactoring processes.
- Apology for the bad posted example, I have changed it into a more convincing one. Here are the code:
```java
package p; // 29, 28, 29, 71

import java.util.LinkedHashMap;
import java.util.Map;

public class A {
	private static final Version VERSION_1_4_0_M3= Version.parse("1.4.0.M3");

	private static final Version VERSION_2_0_0_M1= Version.parse("2.0.0.M1");

	private static final Version VERSION_2_0_0_M6= Version.parse("2.0.0.M6");

	protected String getServletInitializrClass(ProjectRequest request) {
		Version bootVersion= Version.safeParse(request.getBootVersion());
		assert bootVersion != null;
		if (VERSION_1_4_0_M3.compareTo(bootVersion) > 0) {
			return "org.springframework.boot.context.web.SpringBootServletInitializer";
		} else if (VERSION_2_0_0_M1.compareTo(bootVersion) > 0) {
			return "org.springframework.boot.web.support.SpringBootServletInitializer";
		} else {
			return "org.springframework.boot.web.servlet.support.SpringBootServletInitializer";
		}
	}

	protected void resolveModel(ProjectRequest originalRequest) {
		ProjectRequest request= new ProjectRequest();
		// Kotlin supported as of M6
		final boolean kotlinSupport= VERSION_2_0_0_M6
				.compareTo(Version.safeParse(request.getBootVersion())) <= 0;
	}
}

class Version {
	private String version;

	public Version() {
		version= "";
	}

	public String getVersion() {
		return version;
	}

	public void setVersion(String version) {
		this.version= version;
	}

	public static Version safeParse(String version) {
		try {
			return parse(version);
		} catch (Exception ex) {
			return null;
		}
	}

	public int compareTo(Version v1) {
		if (v1.getVersion().equals(this.version))
			return 0;
		else
			return 1;
	}

	public static Version parse(String version) {
		Version v= new Version();
		v.setVersion(version);
		return v;
	}
}

class ProjectRequest {

	String bootVersion= "";

	public String getBootVersion() {
		return bootVersion;
	}

	public void setBootVersion(String bootVersion) {
		this.bootVersion= bootVersion;
	}

}
```
In this example, Eclipse recommends "safeParse" as the name for the selected expression "Version.safeParse(request.getBootVersion())". However, the developer coined "bootVersion" as the new name, which seems to be better. This test case is a modified version of a real-world extract local variable case. 
(https://github.com/spring-io/initializr/commit/faa7de99c2d80300b0e30825b2b54d6d44c43c5c, initializr-generator/src/main/java/io/spring/initializr/generator/ProjectGenerator.java )

jjohnstn:
> Thanks for the valuable suggestions!
> 
>     * All the requested changes were addressed.
> 
>     * I further made some modifications to make the changes only work for the Extract Local Variable Refactoring, which avoids impacting other refactoring processes.
> 
>     * Apology for the bad posted example, I have changed it into a more convincing one. Here are the code:
> 
> 
> ```java
> package p; // 29, 28, 29, 71
> 
> import java.util.LinkedHashMap;
> import java.util.Map;
> 
> public class A {
> 	private static final Version VERSION_1_4_0_M3= Version.parse("1.4.0.M3");
> 
> 	private static final Version VERSION_2_0_0_M1= Version.parse("2.0.0.M1");
> 
> 	private static final Version VERSION_2_0_0_M6= Version.parse("2.0.0.M6");
> 
> 	protected String getServletInitializrClass(ProjectRequest request) {
> 		Version bootVersion= Version.safeParse(request.getBootVersion());
> 		assert bootVersion != null;
> 		if (VERSION_1_4_0_M3.compareTo(bootVersion) > 0) {
> 			return "org.springframework.boot.context.web.SpringBootServletInitializer";
> 		} else if (VERSION_2_0_0_M1.compareTo(bootVersion) > 0) {
> 			return "org.springframework.boot.web.support.SpringBootServletInitializer";
> 		} else {
> 			return "org.springframework.boot.web.servlet.support.SpringBootServletInitializer";
> 		}
> 	}
> 
> 	protected void resolveModel(ProjectRequest originalRequest) {
> 		ProjectRequest request= new ProjectRequest();
> 		// Kotlin supported as of M6
> 		final boolean kotlinSupport= VERSION_2_0_0_M6
> 				.compareTo(Version.safeParse(request.getBootVersion())) <= 0;
> 	}
> }
> 
> class Version {
> 	private String version;
> 
> 	public Version() {
> 		version= "";
> 	}
> 
> 	public String getVersion() {
> 		return version;
> 	}
> 
> 	public void setVersion(String version) {
> 		this.version= version;
> 	}
> 
> 	public static Version safeParse(String version) {
> 		try {
> 			return parse(version);
> 		} catch (Exception ex) {
> 			return null;
> 		}
> 	}
> 
> 	public int compareTo(Version v1) {
> 		if (v1.getVersion().equals(this.version))
> 			return 0;
> 		else
> 			return 1;
> 	}
> 
> 	public static Version parse(String version) {
> 		Version v= new Version();
> 		v.setVersion(version);
> 		return v;
> 	}
> }
> 
> class ProjectRequest {
> 
> 	String bootVersion= "";
> 
> 	public String getBootVersion() {
> 		return bootVersion;
> 	}
> 
> 	public void setBootVersion(String bootVersion) {
> 		this.bootVersion= bootVersion;
> 	}
> 
> }
> ```
> 
> In this example, Eclipse recommends "safeParse" as the name for the selected expression "Version.safeParse(request.getBootVersion())". However, the developer coined "bootVersion" as the new name, which seems to be better. This test case is a modified version of a real-world extract local variable case. ([spring-io/initializr@faa7de9](https://github.com/spring-io/initializr/commit/faa7de99c2d80300b0e30825b2b54d6d44c43c5c), initializr-generator/src/main/java/io/spring/initializr/generator/ProjectGenerator.java )

Hi @Michaelll123, sorry for the delay (we are in test mode for 4.29 M1).  The question is whether the additional code added to Eclipse and doing the computation is worth it for the majority of scenarios and will improve the user experience.  It did in your specific example, but I am skeptical it will do so consistently in the wild after looking at the simple tests.  For example, I don't think precedence would ever apply to non-method call expressions (e.g. extracting an addition operation such as "a + b").  In such an example, a and b could have a completely different context from use to use (e.g. offset + length is a very common pattern you will see but it is likely to be the offset and length for something completely different).

I think that you could improve the name suggestion logic without looking for precedence.  In your example, the resulting type of the expression is Version and the name suggested is "safeParse" which simply regurgitated the outer method call..  Based on that, I think that the suggested name could have been generated better.  For example, instead of "safeParse", it could have generated "safeParseVersion" (since it returned a class Version which name was not part of the method call and it wasn't obvious what safeParse was doing such as we know when we have a getter method).  This would have been very reasonable.  The logic could have been even smarter and realized that the method had a single parameter that was a getter method and come up with "safeParseBootVersion" which would have been even better.  The code already recognizes getters (e.g. getBootVersion() would have come up with bootVersion as variable).  If the logic recognized commonly used coding verbs such as parse and format, it could have even come up with "safeParsedBootVersion" which would have been near perfect.

That said, I think that search logic is only worth it in the case of a method call that is not simple (e.g. multiple parameter and output class is not part of the method name).  I also think that if you find any match, that is likely a reasonable choice and you can stop there without acquiring a consensus.  This would optimize the search to stop with a match.

Michaelll123:
@jjohnstn Hi, is this pull request fine enough to be merged? Please check it out and reply to us. Thanks for your precious time!

jjohnstn:
Hi @Michaelll123 thanks for the update.  Was busy recently and will review tomorrow.

org.eclipse.jdt.core.manipulation/common/org/eclipse/jdt/internal/core/manipulation/StubUtility.java
```diff
@@ -1051,6 +1051,90 @@ public static String[] getVariableNameSuggestions(int variableKind, IJavaProject
...
+					expectedType= expectedType.getElementType();
+				}
+				if (expectedType.isParameterizedType()) {
+					expectedType= expectedType.getTypeDeclaration();
+				}
+				typeName= expectedType.getName();
+			}
+		}
+
+		String recycledName= recycleNames(typeName, assignedExpression, excluded, identicalName, namesInSameMethod);
```
jjohnstn, line 1074: I think it would make sense to not bother calling recycleNames() unless the expression is a MethodInvocation and add a comment (e.g. // if we have a method invocation, see if we can recycle some previously used variable name that is assigned the same expression.

org.eclipse.jdt.core.manipulation/common/org/eclipse/jdt/internal/core/manipulation/StubUtility.java
```diff
@@ -1051,6 +1051,90 @@ public static String[] getVariableNameSuggestions(int variableKind, IJavaProject
...
+						return recycledName;
+					}
+				}
+			}
+
+		}
+		return null;
+	}
+
+	private static String recycleNames(Collection<String> excludedNames, String identicalName, Collection<String> namesInSameMethod) {
```
jjohnstn, line 1124: no need for this method, just put the logic above.

jjohnstn (changes requested):
Hi @Michaelll123 Thanks for the update.  I added some comments.

Michaelll123:
Hi, @jjohnstn, all the requested changes have been carefully addressed. Thanks for your precious time to review the code again.

jjohnstn (approved):

jjohnstn:
Hi @Michaelll123 Thanks for the updates.  Patch merged.

Michaelll123:
Thank you and have a nice day!

Issue #653, opened 2023-07-03
Opening the dialog "Run configurations..." for the first time takes too long --> UI freeze


fedejeanne:
It sometimes takes a while to open the dialog *Run configurations...* for the first time:

![image](https://github.com/eclipse-jdt/eclipse.jdt.ui/assets/2205684/cdf1489c-e94c-4b06-9d08-6864022a35af)

This is the sum of several smaller delays, some of which can be seen in the attached snapshot:
[open-run-configurations-dialog.zip](https://github.com/eclipse-jdt/eclipse.jdt.ui/files/11937105/open-run-configurations-dialog.zip)

## Related tasks
- [x] eclipse-jdt/eclipse.jdt.core#1214 
- [x] https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/683

fedejeanne:
One possible explanation would be the changes introduced in https://github.com/eclipse-jdt/eclipse.jdt.ui/commit/815287b20b0e86d27692b953e6e275b977f8c2b5, specifically the newly introduced call to `collectMethodNames` which ends up resolving the hierarchy of several classes.

It all starts here: 
![image](https://github.com/eclipse-jdt/eclipse.jdt.ui/assets/2205684/ce2ab563-fe6c-46cc-9ade-170ec1ff07b4)

PR #675, opened 2023-07-20
Show progress when searching test methods in JUnit run configuration #653

fedejeanne:
## What it does
Move the logic to search for test methods from `JUnitLaunchConfigurationTab` to `TestSearchEngine::findTestMethods`. Extract the whole logic for caching the results into `TestMethodsCache` (new class) and do the whole searching and caching more efficiently i.e. only when necessary and showing the progress with a monitor by using `ModalContext::run`

Fixes https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/683
Contributes to https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/653

This PR shows a progress indicator when searching for test methods in a class (which may happen when one opens a JUnit run configuration). The progress indicator looks like this when you first open the _Run configurations_ and a JUnit configuration is selected:

image

... and like this when you select or edit an existing JUnit run configuration and force a new search for test methods (the progress indicator appears at the bottom of the _Run configurations_ dialog).

image

## How to test

The main goal is to open some _Run configurations_ for JUnit tests and move between them, letting them search the necessary test methods. 

All of the following must hold true while testing:
* If the UI is blocked (which is normal) then some progress report should appear, either in a separate dialog (when opening the _Run configurations_ for the first time) or in a progress bar at the bottom of the _Run configurations_ dialog.
* Previous results from **the same** run configuration (project + class + JUnit runner) should be cached _e.g._ if you already found all test methods for class `A` in project `B` and for `JUnit5` then changing to class `A2` might trigger a search (if `A2` exists) but setting the class back to `A` should **not** trigger another search.
* Selecting another configuration in the dialog (on the left-side panel) empties the cache.
* If you cancel a search then the _Test method (text field)_ and its _Search_ button should be deactivated since the methods of the class are not known. 
image
* It follows that if you cancel a search then the configuration is **invalid** _i.e._ the _Run_ button should be deactivated and an error should be displayed in the notification area of the dialog (upper area).

### A small hack to help testing
If you need more time to see the progress dialog/bar then you can modify `TestSearchEngine::collectMethodNames` like this:

```java
	private static void collectMethodNames(IType type, IJavaProject javaProject, String testKindId, Set<String> methodNames, IProgressMonitor monitor) throws JavaModelException {
		if (type == null) {
			return;
		}

		SubMonitor subMonitor= SubMonitor.convert(monitor, 3 + 3); // Add +3

		// Add this idle wait of 3 seconds and advance the progress in the subMonitor
		for (int i= 0; i < 3; i++) {
			try {
				Thread.sleep(1_000);
				subMonitor.split(1);
			} catch (InterruptedException e) {
				// ignore
			}
		}

		// leave the rest unchanged
                ...
	}
```

### How did I test?
These are all the stuff I did and tried when testing:
* I created a hierarchy of 2 test classes so that I may run tests in the super-class from the sub-class
* I added test methods both in the super-class and in the sub-class
* I created _Run configurations_ for the test methods and also for the complete classes (_i.e._ left the _Test method_ empty in the configuration dialog).
* I created a completely empty run configuration which will never be able to run but it helps to empty the cache without closing the _Run configurations_ dialog (selecting it empties the cache of the previous JUnit run configuration).
* I edited the existing JUnit configuration in every possible way I could imagine, entering both valid and invalid values for the project, test class and test method and also switching between JUnit3, 4 and 5 in the drop-down menu ("_Test runner_")
* I also used the `Browse...` and `Search` buttons to change the project and the class in the run configuration
* I sometimes canceled the search for test methods
* I changed the project / test class / test runner back and forth several times (without selecting another _Run configuration_ from the left-side panel) and purposely triggered new searches for test methods. 
* I completely emptied the cache by selecting another (an empty) run configuration in the left-side panel --> no searches should occur if the configuration is empty

## Author checklist

- ✔ I have thoroughly tested my changes
- ✔ The change is following the [coding conventions](https://wiki.eclipse.org/Platform/How_to_Contribute#Coding_Conventions)
- ✔ I have signed the [Eclipse Contributor Agreement (ECA)](https://www.eclipse.org/legal/ECA.php)

### DIFF
--- a/org.eclipse.jdt.junit/src/org/eclipse/jdt/internal/junit/ui/JUnitMessages.java
+++ b/org.eclipse.jdt.junit/src/org/eclipse/jdt/internal/junit/ui/JUnitMessages.java
@@ -118,6 +118,8 @@ public final class JUnitMessages extends NLS {
 
 	public static String JUnitLaunchConfigurationTab_error_notJavaProject;
 
+	public static String JUnitLaunchConfigurationTab_error_operation_canceled;
+
 	public static String JUnitLaunchConfigurationTab_error_projectnotdefined;
 
 	public static String JUnitLaunchConfigurationTab_error_projectnotexists;
@@ -372,4 +374,6 @@ public final class JUnitMessages extends NLS {
 	public static String TestRunnerViewPart_JUnitPasteAction_label;
 
 	public static String TestRunnerViewPart_layout_menu;
+
+	public static String TestSearchEngine_search_message_progress_monitor;
 }
--- a/org.eclipse.jdt.junit/src/org/eclipse/jdt/internal/junit/ui/JUnitMessages.properties
+++ b/org.eclipse.jdt.junit/src/org/eclipse/jdt/internal/junit/ui/JUnitMessages.properties
@@ -169,6 +169,7 @@ JUnitLaunchConfigurationTab_folderdialog_message=Choose a Project, Source Folder
 JUnitLaunchConfigurationTab_error_projectnotdefined=Project not specified
 JUnitLaunchConfigurationTab_error_projectnotexists=Project does not exist
 JUnitLaunchConfigurationTab_error_notJavaProject=Specified project is not a Java project
+JUnitLaunchConfigurationTab_error_operation_canceled=(Operation canceled by the user)
 JUnitLaunchConfigurationTab_error_testnotdefined=Test not specified
 JUnitLaunchConfigurationTab_error_testcasenotonpath=Cannot find class 'junit.framework.TestCase' on project build path.
 JUnitLaunchConfigurationTab_addtag_label=Con&figure...
@@ -276,3 +277,5 @@ JUnitViewEditorLauncher_dialog_title=Import Test Run
 JUnitViewEditorLauncher_error_occurred=An error occurred while opening a test run file.
 ClasspathVariableMarkerResolutionGenerator_use_JUnit3=Use the JUnit 3 library
 ClasspathVariableMarkerResolutionGenerator_use_JUnit3_desc=Changes the classpath variable entry to use the JUnit 3 library
+
+TestSearchEngine_search_message_progress_monitor=Searching for test methods in ''{0}''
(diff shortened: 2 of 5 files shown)


fedejeanne:
Random test failure of `SaveParticipantTest.testFormatChangeBug561164` already documented in https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/79

fedejeanne:
Hello all,
there's been no activity on this one in the past few weeks. 
Please let me know if there's anything I can do to help reviewing this PR (more details/screenshots/comments).

I just rebased and solved the conflicts to ease the review.

jjohnstn:
@fedejeanne Will take a look

jjohnstn (commented):
Tried it out.  Works well, but have the following comment.

I used your delay code to be able to cancel a search.  When a search is cancelled, it would be helpful if the error message states that it cannot find method after cancelled search.  After cancelling, I believe the search button for method should remain active and if pressed, should restart the search.  Otherwise, after cancelling a search a user has no options other than to close without saving.  I had a case where it was running all tests and after cancelling it left the Run button active because it did not have to do the verification which is fine.

fedejeanne:
Thank you for reviewing and testing!

I added a hint that says the user canceled the operation in order to let the user know _why_ can't the method be found.
Additionally, the **Search...** button now remains active even after cancellation and triggers the search again if necessary.

image

jjohnstn:
Hi @fedejeanne Thanks for the changes.  Could you take a look at the build failures?  Monday is cut-off for M3 so if we can't get this in by then, it will have to wait until 4.30 M1.

jjohnstn:
There was a fix recently in jdt.ui for Junit 4/5 failures.  Please rebase on master (don't merge) and do a force push if needed.

jjohnstn:
I made the rebase for you.  I had a similar problem with a PR and rebasing fixed the test failures.

jjohnstn:
Thanks @fedejeanne I made a couple tweaks (no logging interrupt exception and fixing error message to not spit out null when the testmethods field not filled in).  Merged for M3.

fedejeanne:
Thank you very much for the review, changes and merging @jjohnstn ! And I'm glad to see that it's a **noteworthy** change!

jjohnstn:
@fedejeanne With regards to noteworthy, would you like to create a N&N PR for this in eclipse-platform/www.eclipse.org-eclipse.git if you haven't already?  If you're not sure, I can do it for you.

fedejeanne:
>... would you like to create a N&N PR...?

My pleasure!

fedejeanne:
@jjohnstn --> https://github.com/eclipse-platform/www.eclipse.org-eclipse/pull/51

Issue #667, opened 2023-07-11
 [Coding conventions] Using brackets for blocks of code with only 1 line of code


fedejeanne:
_(Copied from https://github.com/eclipse-platform/eclipse.platform.ui/issues/928 based on suggestion from @mickaelistria)_

The issue was raised by @HeikoKlare in https://github.com/eclipse-platform/eclipse.platform.ui/pull/755#discussion_r1256321294

The easiest way to make sure this always happens is by setting it in the preferences:

![image](https://github.com/eclipse-jdt/eclipse.jdt.ui/assets/2205684/91f0ad31-128a-45fb-9bfa-90b4684ad454)

What do you guys think:

1. Are you in favor of adding this as a coding convention? (I'm in favor so 👍 from my side)
2. If yes, do you see any other way of doing it besides setting this property in each and every project?

Any hints on the second question would be highly appreciated since setting the property in each project is a load of work and it can easily be forgotten when setting new projects (also, setting only that property on 1 project added like 30+ changed lines in the properties file).

Thank you in advance for your participation!

PR #676, opened 2023-07-21
Always use brackets in blocks of code by default #667

fedejeanne:
## What it does
Change the default settings of the built-in clean-up profile and in the "_Additional actions"_ of the "_Save Actions_"; set them to always use brackets for blocks of code (even for one-liners).

These settings are just the defaults, they can be overwritten.

Fixes https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/667

### Save actions
* Open a newly created workspace
* Open _Window > Preferences > Java / Editor / Save Actions_
* Check _Perform the selected actions on save_
* Check _Additional actions_
* Click _Configure_
* Switch to the tab _Code Style_ --> _Use blocks in if/while/..._ should be checked
  * If it's not checked then you probably loaded some preferences from your user-space. Click _Restore defaults_ --> it should now be checked.

### Clean up (only the default profile)
* Open a newly created workspace
* Open _Window > Preferences > Java / Code Style / Clean Up_
* Select the default profile: _Eclipse [built-in]_
* Click _Edit..._
* Switch to the tab _Code Style_ --> _Use blocks in if/while/..._ should be checked.

## Author checklist

- ✔ I have thoroughly tested my changes
- ✔ The change is following the [coding conventions](https://wiki.eclipse.org/Platform/How_to_Contribute#Coding_Conventions)
- ✔ I have signed the [Eclipse Contributor Agreement (ECA)](https://www.eclipse.org/legal/ECA.php)

### DIFF
--- a/org.eclipse.jdt.ui/core extension/org/eclipse/jdt/internal/corext/fix/CleanUpConstantsOptions.java
+++ b/org.eclipse.jdt.ui/core extension/org/eclipse/jdt/internal/corext/fix/CleanUpConstantsOptions.java
@@ -16,10 +16,13 @@
 package org.eclipse.jdt.internal.corext.fix;
 
 import org.eclipse.core.runtime.Assert;
+
+import org.eclipse.jface.preference.IPreferenceStore;
+
+import org.eclipse.jdt.ui.cleanup.CleanUpOptions;
+
 import org.eclipse.jdt.internal.ui.JavaPlugin;
 import org.eclipse.jdt.internal.ui.fix.UnimplementedCodeCleanUp;
-import org.eclipse.jdt.ui.cleanup.CleanUpOptions;
-import org.eclipse.jface.preference.IPreferenceStore;
 
 public class CleanUpConstantsOptions extends CleanUpConstants {
 	private static void setEclipseDefaultSettings(CleanUpOptions options) {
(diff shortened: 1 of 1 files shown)


jjohnstn:
Hi @fedejeanne I think this is too contentious a change.  Users who use single line blocks on a regular basis will be very annoyed when performing a cleanup or saving when they see changes to code they weren't expecting to change.  While I agree with the practice:  I personally always use brackets in all blocks manually, it is not a good idea to impose a default that can make sweeping changes that not everyone will agree about.  Some projects store save preferences for this very reason so all users automatically follow established coding guidelines.

fedejeanne:
Thank you @jjohnstn for your input. Do you see any better approach to guarantee that at least the projects under [eclipse-platform](https://github.com/eclipse-platform), [eclipse-jdt](https://github.com/eclipse-jdt), etc use this by default without having to ship it in the Eclipse-defaults (see https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/667#issuecomment-1630476961)?

mickaelistria:
@jjohnstn on the issue, some other contributors have expressed some support for this feature. Users of the "bad practice" will have to fix their project-specific settings. Or maybe they'll just get used to the best practice and enjoy it. In any case, I support that a good development tool should encourage the good practices, even if it annoys a few users.

iloveeclipse:
I would also support that change. We should not provide defaults that are considered "bad practice".

Changing defaults doesn't mean there is no way back for users. Those who will be annoyed by the change & want keep "old good" settings, can set project specific options to old default & be happy again.

jjohnstn:
Hi @fedejeanne I think this is too contentious a change.  Users who use single line blocks on a regular basis will be very annoyed when performing a cleanup or saving when they see changes to code they weren't expecting to change.  While I agree with the practice:  I personally always use brackets in all blocks manually, it is not a good idea to impose a default that can make sweeping changes that not everyone will agree about.  Some projects store save preferences for this very reason so all users automatically follow established coding guidelines.

> I would also support that change. We should not provide defaults that are considered "bad practice".
> 
> Changing defaults doesn't mean there is no way back for users. Those who will be annoyed by the change & want keep "old good" settings, can set project specific options to old default & be happy again.

If that is the consensus, then I'm certainly not going to fight it.

fedejeanne:
> If that is the consensus, then I'm certainly not going to fight it.

Ok. Does that mean you will merge it after the freeze period? :-)

jjohnstn:
> > If that is the consensus, then I'm certainly not going to fight it.
> 
> Ok. Does that mean you will merge it after the freeze period? :-)

After review, yes, for 4.30 M1.  I'm currently ill, but I hope to get this done by next week.

fedejeanne:
> After review, yes, for 4.30 M1. I'm currently ill, but I hope to get this done by next week.

Thank you and I hope you get well soon! :-)

jjohnstn (approved):
Looks good for 4.30 M1

fedejeanne:
> Looks good for 4.30 M1

Thank you! I just rebased so it should be ready to be merged :)

jjohnstn:
Test failure appears random as local run of all JDTTextTestSuite ran successfully.

PR #677, opened 2023-07-21
Remove unused data from compilation unit annotation model util map #665

### LINKED ISSUE #665: UI freezes when showing large numbers of annotations in Java files
## Description
The Java editor uses the `AnnotationModel` to show different types of annotations, such as the occurrences of the currently selected element. In case a file is quite large, this calculation can take quite long, for example, when changing the text selection to a different method.

## Reproduction
As an example, you can open the Java file in the attachment below and change the text selection between the `oftenCalledMethod` and other elements. This is an artificial and really extreme case, just for demonstration purposes. In our actual project, we experience the same problem (with even more performance problems) in large generated files with complex type and call hierarchies. 
[Test.zip](https://github.com/eclipse-jdt/eclipse.jdt.ui/files/12004611/Test.zip)

## Expected Behavior
The Java code editor should still be responsive when showing large source code files. In particular, navigation through the file should still happen in a responsive way.

## Reasons
The reason for the behavior is an inefficient implementation of the replacement of annotations in combination with the fact that the update of the annotation model requires a lock that blocks the UI, even though it is executed in a background job.
The replacement of annotations in the `AnnotationModel` (as done when changing the selection in an editor) processes every single annotation to be removed and added on its own. The Java-specific `CompilationUnitAnnotationModel` in the `CompilationUnitDocumentProvider` uses a `ReverseMap` that relies on an inefficient lookup within an `ArrayList` for every annotation to be added, removed or retrieved. The effort grows super-linear with the number of annotations to process.

## Possible Improvement
When performing a replacement of annotations, they should not be treated as individual removals and additions but as a batch operation. Then a temporary hash-based lookup structure can be created during the annotation model update to avoid the inefficient lookups in the currently used `ArrayList`.

I will propose a fix that implements such an improved annotation replacement.

HeikoKlare:
## What it does

The `CompilationUnitDocumentProvider` uses in its Java-specific annotation model a utility map to navigate from document positions back to annotations (called `ReverseMap`). The current implementation is slow whenever a document contains plenty of annotations (such as the occurrences of the name of a marked method), as the map internally uses a list for representing the annotations which becomes slow in adding/removing when containing plenty of data. This is especially a problem on batch replacements of annotations (e.g. when selecting a different method to be marked in the document), as every single annotation is processed by the map individually. In addition, the map currently stores data that is never used, as only the specific JavaMarkerAnnotations are used from that map.

This change thus does the following:
- Specializes the utility map to only contain the required type of annotations
- Refactors the map implementation to simplify the code and make it more comprehensible

## How to test

Navigate through some large Java file and perform something that produces many annotation in the file. For example, use the test class as described and provided in #665. Without the change of this PR, the application freezes when navigating through the file, especially when selecting the `oftenCalledMethod`. With the change, the application stays responsive.

## Author checklist

Fixes #665

### DIFF
--- a/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/javaeditor/CompilationUnitDocumentProvider.java
+++ b/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/javaeditor/CompilationUnitDocumentProvider.java
@@ -20,6 +20,7 @@ import java.io.InputStreamReader;
 import java.io.Reader;
 import java.net.URI;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/javaeditor/CompilationUnitDocumentProvider.java
```diff
@@ -434,83 +435,86 @@ public boolean isQuickFixable() {
...
 
 	/**
-	 * Internal structure for mapping positions to some value.
+	 * Internal structure for mapping positions to JavaMarkerAnnotations.
 	 * The reason for this specific structure is that positions can
 	 * change over time. Thus a lookup is based on value and not
 	 * on hash value.
 	 */
-	protected static class ReverseMap {
+	protected static class ReverseJavaMarkerAnnotationsMap {
```
fedejeanne, line 443: Some hints: 

1. The visibility of `annotations` should be >= than the visibility of this class, otherwise you may end up knowing about the class but can't really use it properly (in subclasses or classes in the same package)
2. Can this class be `private` instead of `protected`?

HeikoKlare: From my point of view, the class should be `private`. I just did not change its prior visibility, but actually I would prefer your suggestion to have it `private`.

HeikoKlare: I have adapted the visibility in 576aa64.

jjohnstn (approved):
Looks good.  Testing it out with included zip file does indeed show a significant performance boost.

HeikoKlare:
Thanks for testing and merging, @jjohnstn!

PR #680, opened 2023-07-23
Extract Similar Expression in All Methods If End-Users Want

Michaelll123:
## What it does
see issue #785  

## How to test
run test159() and test160().

## Author checklist


Michaelll123:
Hi, @jjohnstn, it has been two weeks since we submit this pull request, could you please spare some time to review this? Thanks for your precious time and have a nice day~

jjohnstn:
@Michaelll123 will do today.

jjohnstn:
Hi @Michaelll123 I think my comment from your previous change applies to this as well.  You can't determine the context of an expression is the same as another one simply by looking at it's string representation.  For starters, toString() representation of an expression is not for coding usage, only debugging.  It could change at any time.

For example:

```
public class TestExtract {

	public void foo(int x, int y, int z) {
		int volume = x * y * z;
	}

	public void foo2(int x, int y, int z) {
		int total = x * y + z;
	}

}
```

In foo(), I might extract x * y to: int area = x * y;

Extracting x * y in the second case might be something completely different (int subtotal = x * y).  It isn't area.  It isn't even used in the same way (adding z vs multiplying by z).

I think your use case is very specific and doesn't apply to a general user.  Based on your given scenarios, you seem to want to have a multiple extract temp for all usages of similar expression in a file to save time.  This scenario would be rare and doesn't warrant changing the current behavior with extra code and as mentioned, it can easily pick a wrong name.  If you really want a multiple extraction, then add an extra option to the Extract Local Variable dialog to do a multiple extract (extract similar expression in all methods).  Then, after the user has selected you can do extra work to verify the expression uses similar names and equal types and you can add a suffix if the variable name is already used, etc...  It will be under control by the end-user.

Michaelll123:
Hi, @jjohnstn,  sorry for the late response, and we have implemented the functionality that allows end-users to decided whether extract all the identical expressions in the same file in these days. Please spare some time to review the code, thanks!

Michaelll123:
Hi, @jjohnstn, it has been 2 weeks since we submitted the PR as suggested, and would u please spare some time to review it? I will really appreciate that.

jjohnstn:
> Hi, @jjohnstn, it has been 2 weeks since we submitted the PR as suggested, and would u please spare some time to review it? I will really appreciate that.

Will do.  We have been frozen since 4.29 M3 and 4.30 M1 opens this week.  There have been a number of patches ahead of you in the queue that I am merging but will look at this either today or tomorrow.

jjohnstn (changes requested):
Hi @Michaelll123, thanks for the new patch.  I tried it out.  A few simple things to deal with first.

1. The original issue is no longer valid.  You are no longer using history based name selection.  I suggest you close the old issue and create a new issue that has details about what you are actually doing and change references in this patch.
2. The new option should be off by default.  I am finding it is always on.
3. Please rebase your patch on master as there are changes that are required to test the change.  There is an editor error otherwise.  Don't merge from master.
4. New files should have a single year 2023 copyright.  If they are created by modifying an existing file, add an additional comment in the copyright Contributors that it was modified from X where X is the file it was copied from.

That said, I tried a couple of tests.  The first used super method calls and worked great.  The second was a simple test which did not act as expected:

```
public class TestExtractTemp {
	
	public int getLength1(String x, int a, int b, int c) {
		int z = a + b + c;
		return z;
	}
	
	public int getLength2(String x) {
		int a = 1;
		int b = 2;
		int c = 3;
		int z = a + b + c;
		return z;
	}

}
```

If you select (a + b) in the first method and ask the code to extract that into a new variable k, the first instance is extracted as expected (k = a + b; z = k + c;).  The second instance gets: (k = a + b + c; z = k;)
I deliberately made one reference parameters while the other referenced local variables to see what would occur.

Either the second instance should be ignored or should be changed in a similar manner.  Please ping me via a comment when these are addressed and I will continue the review.

Michaelll123:
Hi, @jjohnstn, I have addressed all the requested changes. For the last question, we choose to ignore this case and only extract the selected expression into a local variable. Please review again, thanks!

jjohnstn (approved):

jjohnstn (changes requested):
Hi @Michaelll123 thanks for the fixes.  I did some debugging and figured out how to handle the infix expression properly.  What you need to do is to create a new form of your getSelectedExpression() routine that takes a start offset and a length.  In the case of other matches, the start needs to be the offset of the node plus whatever offset the original selection is within its node and the length needs to be the length of the original selection (not the length of the node).  So, you don't end up doing a createFragmentForFullSubTree() in processOtherIdenticalExpressions() but call the new getSelectedExpression() method and ultimately createFragmentForSourceRange().  I created a prototype getSelectedExpression(int, int) and for the infix expression, I passed the start offset (didn't add the offset logic) and the original selected expression length.  This worked and I was able to extract a + b from a + b + c properly across multiple methods.  As mentioned, I would have to compute an offset when the selection was further in the expression (e.g. extracting b + c from a + b + c.)  I think this technique should be used for all extra matches (not just infix expressions).  Let me know when you have this change done and I think it will work quite well.

Michaelll123:
Hi, @jjohnstn , I have made the requested changes, and it does work well. Thanks for the advice, and please review the code again!

org.eclipse.jdt.ui.tests.refactoring/test cases/org/eclipse/jdt/ui/tests/refactoring/ExtractMultipleTempTest1.java
```diff
@@ -0,0 +1,81 @@
+/*******************************************************************************
```
jjohnstn, line 1: This test has no way of running.  See other ExtractTemp tests for how to make these part of the automatic testsuite.

org.eclipse.jdt.ui.tests.refactoring/test cases/org/eclipse/jdt/ui/tests/refactoring/ExtractMultipleTempTest2.java
```diff
@@ -0,0 +1,30 @@
+/*******************************************************************************
```
jjohnstn, line 1: Same issue as test1.

jjohnstn (changes requested):
New logic works great.  Just need to add tests to the automated test suite.

org.eclipse.jdt.ui.tests.refactoring/test cases/org/eclipse/jdt/ui/tests/refactoring/ExtractMultipleTempTest1.java
Michaelll123: will do in a minute.

Michaelll123:
Hi, @jjohnstn , the test issue has been addressed.

jjohnstn (approved):

org.eclipse.jdt.ui/ui refactoring/org/eclipse/jdt/internal/ui/refactoring/ExtractTempWizard.java
```diff
@@ -102,8 +104,11 @@ public void createControl(Composite parent) {
 			layouter.perform(label, text, 1);
 
 			addReplaceAllCheckbox(result, layouter);
-			addDeclareFinalCheckbox(result, layouter);
+			addReplaceAllInFileCheckbox(result, layouter);
 			addDeclareTypeVarCheckbox(result, layouter);
+			addDeclareFinalCheckbox(result, layouter);
```
jjohnstn, line 109: Don't change the previous order of existing fields.  It should be final then var option.

org.eclipse.jdt.ui/ui refactoring/org/eclipse/jdt/internal/ui/refactoring/ExtractTempWizard.java
```diff
@@ -102,8 +104,11 @@ public void createControl(Composite parent) {
 			layouter.perform(label, text, 1);
 
 			addReplaceAllCheckbox(result, layouter);
-			addDeclareFinalCheckbox(result, layouter);
+			addReplaceAllInFileCheckbox(result, layouter);
```
jjohnstn, line 107: After testing, this checkbox should be indented and should be a sub-option of replace all (i.e. is disabled if replace all is not checked).  I tried setting the new option without the previous one and it changed one instance in each method which doesn't make sense.

jjohnstn (changes requested):
Things are working well.  Just would like a change in the option usage/presentation.  The new option should be a sub-option of Replace all.  Without replace all specified, the new option should be disabled and shouldn't apply (i.e. you cannot specify change similar expression in all methods without specifying change all instances as doing this presently changes one instance in each method then stops which doesn't really make sense).  I would suggest where you check for your new option being set, you also check that the original all instance option is enabled as well.  With that change, this can be merged.  Also, you changed the order of the remaining options when there was no reason to do so and that should be put back.  If you can, make the new checkbox indented under the all instance option.  Regardless, you should enable/disable it appropriately.  If you have issues with the indentation, I can do that later.

Michaelll123:
Hi, @jjohnstn , I have made the new option as a sub-option of "Replace all" option, and the new option is disabled if "Replace all" is not specified. But I am not sure of the meaning of "Indentation". Please check whether the current way is ok. Thanks for the review.

jjohnstn (approved):

jjohnstn:
Hi @Michaelll123 I added the indent logic and disabled the checkbox on start-up if the master setting is not enabled.  It works fine in my local Eclipse.  There is some issue with the github build currently (can't get other patches to build either).  When it is fixed, I'll check in the patch.

PR #689, opened 2023-07-28
Feature Proposal Make Static Refactoring

### LINKED ISSUE #590: Proposal: Make Static Refactoring
# Proposal: Make Static Refactoring

## Background
The Eclipse IDE provides plenty of refactorings for different use cases. We want to implement a `MakeStatic` refactoring like the one that IntelliJ provides. The following description should give you a little insight to our planning and the functions that we want to include in the refactoring.

## Goal
The `MakeStatic` refactoring can be applied to instance methods and fields to make them static. It updates accesses and calls of these methods and fields to retain syntactic correctness and the execution semantics of all code currently accessible by the JDT tooling.

## Description
`MakeStatic` refactoring can be used in different ways. At first we will implement it for methods only. Later, an implementation for fields is also possible.
By performing the refactoring, a `RefactoringWizard` will open, giving the user a preview of the changes to the source code. The user can confirm the changes or cancel the refactoring.

A static keyword will be added to the selected function when the refactoring is performed. Furthermore, the references to this method will be adjusted so that every invocation is working properly after the refactoring.

#### Before refactoring
```java
class Foo {
  public void bar() {
      int i = 0;
  }

  public void Reference() {
      Foo foo = new Foo();
      foo.bar();
  }
}
```

#### After refactoring `bar`
```java
class Foo {
   public static void bar() {
      int i = 0;
  }

   public void Reference() {
      Foo foo = new Foo();
      Foo.bar();
  }
}
```

### Special Application Conditions
There are several special cases in which the refactoring must function properly. The initial functionality we plan to include should analyze those cases and give the user information why the refactoring cannot be executed.

The initial implementation should only work under the following conditions:

* The refactored method does not access instance variable/method.
* The refactored method is not a constructor.
* The refactored method does not override a method of the parent type and also does not have a child type where the method is overridden.
* The type in which the refactored method is declared must not be a local type, an anonymous type or an annotation.
* The refactoring can be executed if a MethodInvocation is selected but not if a SuperMethodInvocation is selected.

This is the minimum we plan to bring into this refactoring for now. In the following we provide a list with additional functions this refactoring should be able to do in the future.

### Precondition checks
In the following we discuss which of the above conditions will be checked as **initial preconditions** to make sure that the refactoring can be started and the workspace is in a stable state.

#### Initial conditions
1. Check for constructor:
The refactoring should not be able to change a constructor method into a static method.

2. Check for local type, anonymous type or annotation:
The parent type of the refactored method must not be local because it is not possible to access them outside of the method they are declared in. So a static method is not possible in local types. Anonymous types do not have a class name to associate them with so they also do not allow static methods. Annotations can have methods but should be used for metadata and not behavior, so static methods will not be allowed. 

3. Check for SuperMethodInvocation:
The refactoring should not change the method of the parent type to static if a SuperMethodInvocation in the child type is selected. It would not be possible to access the method of the parent type with the "super" keyword anymore.

#### Final conditions
After the **initial conditions** are met the refactoring can be started. It is important to add **final checks** that have to be met in runtime. The user gets a error message with the specific status if a condition check fails:

1. Check for access to instance variables and methods:
The refactoring should not change a method to static if it contains instance variable access or instance method access. A static method can only access static variables or methods without having an instance of a class.

2. Check if method overrides a method of parent type:
It should not be possible to refactor a method that overrides a method of the parent type. This is because a static method cannot override a method of the parent type. It is important that the AST will be analyzed for overrides before executing the refactoring.

3. Check if method is overridden in hierarchy:
The refactoring can not be executed if the refactored method is overridden in a child type because a static method cannot be overridden.

### Extensions
When the initial scope of the refactoring is implemented, extensions can be added.
Most extensions add functionality for the various edge cases covered in the conditions.

1. Instance variables and instance method access in method body:

    Should a type contain instance variables or methods that are used in the refactored 	method, the `MakeStatic` refactoring will have the same behavior but additionally an instance of the type will be added to the parameter list of the method. Every access of an instance variable or instance method inside the refactored method will be modified with the new parameter.

 #### Before refactoring
```java
class Foo {
       int i = 0;

       public void bar() {
            this.i = 0;
       }

       public void instanceReference() {
             this.bar();
        }

        public static void staticReference() {
              Foo foo = new Foo();
              foo.bar();
         }
}
```

 #### After refactoring `bar`
```java
class Foo {

         int i = 0;

          public static void bar(Foo foo) {
               foo.i = 0;
          }

         public void instanceReference() {
               Foo.bar(this);
         }

         public static void staticReference() {
               Foo foo = new Foo();
               Foo.bar(foo);
          }
}
```

   Additionally some final conditions need to be checked to make sure that the refactoring works properly:

 - Check for duplicate method signature:
It is possible that a method with the same signature as the refactored method already exists. In this case the refactoring will be canceled and the user will receive an error message.

 - Check for duplicate parameter name:
If the new parameter that is introduced to the refactored method has the same name as an already existing parameter the refactoring will be canceled and the user will receive an error message.

2. Refactoring a method that overrides a method of the parent type:

    This extension should make it possible to refactor a method that overrides a method of the parent type. After the refactoring a potential  `Override` annotation is deleted and the “super”-keyword will be exchanged with a new parameter containing an instance of the subclass. This is only possible when there is no further override in the lower hierarchy. If there is another override the same error will be shown as in the initial implementation. This means that only the lowest method in the hierarchy can be refactored. Even though no compiler error will be thrown after refactoring, it is necessary to include a warning when refactoring an overridding method in the subtype. This is because the program semantics can be changed very quickly if the hierarchy will be changed.

 #### Before refactoring
 ```java
public class SuperClass {

        public String toRefactor(String bar) {
              String i = bar;
              return i;
        }
}
```
```java
public class SubClass extends SuperClass {

        @Override
        public String toRefactor(String bar) {
             String i = super.toRefactor(bar);
             return i;
        }

        public static void staticMethod() {
            SubClass instance = new SubClass();
            String j = instance.toRefactor("bar");
        }
}
```

 #### After refactoring `toRefactor`

 ```java
public class SuperClass {

        public String toRefactor(String bar) {
            String i = bar;
            return i;
        }
}
```
```java
public class SubClass extends SuperClass {

       public static String toRefactor(String bar, SubClass subclass) {
           String i = subclass.toRefactor(bar);
           return i;
       }

       public static void staticMethod() {
           SubClass instance = new SubClass();
           String j = SubClass.toRefactor("bar", instance);
       }
}
```

## Feedback
We look forward to your feedback on the idea for adding such a `MakeStatic` refactoring. We particularly appreciate your input on:
1. The overall idea of having such a refactoring
2. Specific pointers to preconditions or special cases that may need to be consider before or while applying the refactoring.
3. Ideas for further extensions of the refactoring.

CodeLtDave:
Fixes #590

Introducing a new Refactoring called "Make Static", a tool that transforms instance methods into static ones in the context of Eclipse's Java Development Tools. This involves adjusting method invocations appropriately to maintain logical correctness. The refactoring supports both instance and static contexts, passing 'this' for invocations within instance methods and the relevant instance variable for invocations in other classes or static methods. The refactoring utility aids in improving code modularity and reuse.

## What it does
The "Make Static Refactoring" transforms instance methods into static methods while ensuring the code remains logically correct. The refactoring process involves two key steps:

Modifying the method declaration: The instance method is changed into a static one by adjusting its declaration. If needed an additional parameter (an instance of the enclosing class) is added to the method signature to replace the usage of 'this' keyword in the method body.

Adjusting method invocations: All occurrences of the method invocations are modified to suit the new method signature. If the method invocation occurs in a static context, the relevant instance variable is passed as an argument. If the invocation is in an instance context, 'this' keyword is passed as an argument.

This refactoring utility is designed to improve code modularity and reuse by allowing methods to be invoked without the need for class instantiation.

## How to test
Utilize the provided MakeStaticRefactoringTests.java file, which includes various test cases for different scenarios. Run this file as a JUnit test to validate functionality.

## Author checklist


jukzi:
That is much code to review. As i understand that feature should not break anything. Please prove that by applying it on the jdt.core repository and submitting that dogfooding PR.

org.eclipse.jdt.ui/ui/org/eclipse/jdt/ui/actions/IJavaEditorActionDefinitionIds.java
```diff
@@ -394,6 +394,14 @@ public interface IJavaEditorActionDefinitionIds extends ITextEditorActionDefinit
 	 */
 	String EXTRACT_CLASS= "org.eclipse.jdt.ui.edit.text.java.extract.class"; //$NON-NLS-1$
 
+	/**
+	 * Action definition ID of the refactor -> make static action
+	 * (value <code>"org.eclipse.jdt.ui.edit.text.java.make.static"</code>).
+	 *
+	 * @since 3.31
```
jjohnstn, line 401: This public api change will require bumping jdt.ui to 3.32.0 and this since tag will need changing to 3.32.

org.eclipse.jdt.ui/ui/org/eclipse/jdt/ui/actions/JdtActionConstants.java
```diff
@@ -392,6 +392,14 @@ public class JdtActionConstants {
 	 */
 	public static final String EXTRACT_CONSTANT= "org.eclipse.jdt.ui.actions.ExtractConstant"; //$NON-NLS-1$
 
+	/**
+	 * Refactor menu: name of standard make static global action
+	 * (value <code>"org.eclipse.jdt.ui.actions.makeStatic"</code>).
+	 *
+	 * @since 3.31
```
jjohnstn, line 399: similar bump needed to 3.32

org.eclipse.jdt.ui.tests.refactoring/test cases/org/eclipse/jdt/ui/tests/refactoring/MakeStaticRefactoringTests.java
```diff
@@ -0,0 +1,578 @@
...
+import org.eclipse.jdt.core.dom.ASTParser;
+import org.eclipse.jdt.core.dom.CompilationUnit;
+
+import org.eclipse.jdt.internal.corext.refactoring.RefactoringCoreMessages;
+import org.eclipse.jdt.internal.corext.refactoring.code.MakeStaticRefactoring;
+
+import org.eclipse.jdt.ui.tests.refactoring.infra.TextRangeUtil;
+import org.eclipse.jdt.ui.tests.refactoring.rules.RefactoringTestSetup;
+
+public class MakeStaticRefactoringTests extends GenericRefactoringTest {
```
jjohnstn, line 45: This class needs to be added AllRefactoringTests.java

org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/refactoring/code/makestatic/ChangeCalculator.java
```diff
@@ -0,0 +1,548 @@
...
+		List<SingleVariableDeclaration> methodParams= fTargetMethodDeclaration.parameters();
+		List<String> methodParameterTypes= new ArrayList<>(methodParams.size());
+		for (SingleVariableDeclaration methodParam : methodParams) {
+			Type type= methodParam.getType();
+			//if type is parameterized, then those typeParams need to be included in TypeParamList of method
+			if (type.isParameterizedType()) {
+				ParameterizedType parameterizedType= (ParameterizedType) type;
+				List<Type> typeParamsOfMethodParam= parameterizedType.typeArguments();
+				for (Type typeParamOfMethodParam : typeParamsOfMethodParam) {
+					methodParameterTypes.add(typeParamOfMethodParam.resolveBinding().getName());
```
jjohnstn, line 366: resolveBinding() can return null.  You need to check this here and other places to ensure you don't cause an NPE.

jjohnstn (changes requested):
Overall, looks good.  There are a few comments I have in the code.  One general one is to check the result of resolveBinding() for null before using it.

HeikoKlare:
> That is much code to review. As i understand that feature should not break anything. Please prove that by applying it on the jdt.core repository and submitting that dogfooding PR.

A dogfooding PR is a good idea. Should that only be for demonstration or with some real value, i.e., to actually be merged? In the latter case, I would propose to take a look at test classes, because they often contain utility methods defined as instance methods although they do not access the actual test instance. One example could be: `org.eclipse.jdt.core.tests.compiler.regression.AbstractRegressionTest#getCompilerTestsPluginDirectoryPath()`

I already applied the refactoring to some places in jdt.core and while most cases were handled properly, I have two findings:
1. I had one place that could not be refacored but threw an error because of "overlapping text edits". This is the case for
`org.eclipse.jdt.internal.codeassist.CompletionEngine#checkCancel()`.
2. It seems like a parameter for the original `this` object is always added to the method even if the method does not contain any instance access. Is that true and intended? That can, for example, be seen when refactoring
`org.eclipse.jdt.core.tests.compiler.regression.AbstractRegressionTest#getCompilerTestsPluginDirectoryPath()`. The refactoring adds an `AbstractRegressionTest` as parameter although I see no necessity for an accepting an instance of the class (and no usages of that parameter after applying the refactoring).

One final remark to the state of this PR: The contribution is the result of an internship in which we already did several reviews of the code before it was proposed here. For the sake of transparency, the reviews are still available: [Review 1](https://github.com/CodeLtDave/eclipse.jdt.ui/pull/3), [Review 2](https://github.com/CodeLtDave/eclipse.jdt.ui/pull/4), [Review 3](https://github.com/CodeLtDave/eclipse.jdt.ui/pull/5)

jjohnstn:
@HeikoKlare Regarding dogfooding, no we don't need to merge changes.  The point was to prove that the code doesn't simply break in real-world code.  I would like to get this code into 2023-03 M1 which has a deadline of Thursday this week but I do not see any responses to my review comments and it does not look like they have been implemented (e.g. the version bump required now that this patch has to go into 4.31).  I am also fine with having a few corner problem scenarios as we have M2 and M3 to solve them and to expose more. That is why we try and get large features like this into M1.  New issues would need to be opened as soon as the code is merged.

org.eclipse.jdt.ui.tests.refactoring/test cases/org/eclipse/jdt/ui/tests/refactoring/MakeStaticRefactoringTests.java
CodeLtDave: "Thank you for pointingthat out. I've addressed this in 3eb9df5dea85910a1bdf5178386d8310b37e4140

org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/refactoring/code/makestatic/FinalConditionsChecker.java
```diff
@@ -0,0 +1,208 @@
...
+	 *
+	 * @param imethod the IMethod representing the method declaration
+	 *
+	 * @throws JavaModelException if an exception occurs while accessing the Java model
+	 */
+	public void checkMethodIsNotDuplicate(MethodDeclaration methodDeclaration, IMethod imethod) throws JavaModelException {
+		int parameterAmount= methodDeclaration.parameters().size() + 1;
+		String methodName= methodDeclaration.getName().getIdentifier();
+		IMethodBinding methodBinding= methodDeclaration.resolveBinding();
+		ITypeBinding typeBinding= methodBinding.getDeclaringClass();
```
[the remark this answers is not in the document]
jjohnstn: Resolved

org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/refactoring/refactoring.properties
```diff
@@ -1345,6 +1345,30 @@ IntroduceIndirectionRefactoring_call_warning_anonymous_cannot_qualify=The declar
...
+MakeStaticRefactoring_checking_preconditions=Checking preconditions...
+MakeStaticRefactoring_checking_conditions=Checking conditions...
+MakeStaticRefactoring_creating_changes=Creating changes...
+MakeStaticRefactoring_not_available_for_super_method_invocations=Make Static is not available for super method invocations.
+MakeStaticRefactoring_duplicate_method_signature=Method with the same signature as the refactored method already exists.
+MakeStaticRefactoring_method_already_static=Selected method is already static.
+MakeStaticRefactoring_method_is_overridden_in_subtype=Selected method is overridden in subtype. 
+MakeStaticRefactoring_hiding_method_of_parent_type=Make Static is not available. Method of parent type would be hidden after refactoring.
+MakeStaticRefactoring_not_available_for_recursive_methods=Make Static is not available for recursive methods.
+MakeStaticRefactoring_not_available_for_parametrized_methods=Make Static is not available for parametrized methods.
```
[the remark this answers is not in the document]
jjohnstn: Resolved

jjohnstn (approved):

jjohnstn:
Adding this to M1 to ensure proper time to test the feature.

jjohnstn:
@CodeLtDave Would you like to prepare a N&N entry for this new feature or would you like me to do it?

CodeLtDave:
> @CodeLtDave Would you like to prepare a N&N entry for this new feature or would you like me to do it?

I really appreciate the time you put in this pr. 

This is my first big pr to eclipse, can you elaborate what you mean with N&N entry? I think it would be a useful learning to me to do it myself if it is not too complicated. 

I was addressing the issues you found, how can I implement these?

jjohnstn:
Hi @CodeLtDave The @author tag issue can wait for M2 and requires a test.  I added resolveBinding checking and fixed the typo.  The issues that @HeikoKlare found need issues opened for M2.

The N&N entry is added to  www.eclipse.org-eclipse/news/4.31/jdt.html  You can read the instructions.html found in the news directory and see the previous N&N entries for 4.30 as an example.

org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/refactoring/code/makestatic/ChangeCalculator.java
```diff
@@ -0,0 +1,548 @@
...
+	}
+
+	private void updateJavaDocs() {
+		Javadoc javadoc= fTargetMethodDeclaration.getJavadoc();
+		if (javadoc != null) {
+			TagElement newParameterTag= fTargetMethodDeclarationAST.newTagElement();
+			newParameterTag.setTagName(TagElement.TAG_PARAM);
+			newParameterTag.fragments().add(fTargetMethodDeclarationAST.newSimpleName(fParameterName));
+			ListRewrite tagsRewrite= fTargetMethodDeclarationASTRewrite.getListRewrite(javadoc, Javadoc.TAGS_PROPERTY);
+			tagsRewrite.insertFirst(newParameterTag, null);
```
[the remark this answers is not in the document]
HeikoKlare: If I verified correctly, the parameter documentation will always be the first javadoc entry with the current implementation (even before any "general" documentation). I propose to reuse functionality from existing refactoring implementations (such as "Change Method Signature") for a reasonable placement of the parameter documentation in javadoc. I have documented the issue here: #1043

HeikoKlare:
Thank you very much for rewiewing and merging the contribution, @jjohnstn!

I have documented all findings in dedicated issues:
- #1043 for the javadoc tag issue
- #1044 for the unnecessary parameter addition
- #1045 for the "overlapping text edits" failure

We will work on these issues to achieve an improved state of the refactoring for M2.

HeikoKlare:
For the record: I have put some more effort into verifying the behavior of the "Make Static" refactoring. To this end, I have applied the refactoring to every method within the class `org.eclipse.jdt.core.tests.compiler.regression.AbstractRegressionTest` for which the preconditions are fulfilled (e.g., not being an overwritten or overwriting method).

The results are as follows:
* `Compilation`: Only a single compile error occurs. The reason is documented in #1083 and fixed in #1085. With that fix, no compile error occurs.
* `Behavior`: I have executed the regression test suite (`org.eclipse.jdt.core.tests.compiler.regression.TestAll`), in which many test classes extend the refactored `AbstractRegressionTest`. I have 70 test failures (out of more than 100k tests) in my local setup, but they are exactly the same for the code as-is and the refactored code, so no changes in behavior because of the refactoring.

jukzi:
> * I have executed the regression test suite

the good thing about dogfooding PR would be that all those checks come for free.

HeikoKlare:
> the good thing about dogfooding PR would be that all those checks come for free.

Thanks for refreshing the notice on a dogfooding PR, @jukzi! Sorry that I forgot about that one after the feature had been merged before I made that PR. I have opened the according PR for JDT Core: eclipse-jdt/eclipse.jdt.core#1878
Everything compiles and all tests (in particular the [regression tests](https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-1878/1/testReport/org.eclipse.jdt.core.tests.compiler.regression/) affected by the refactoring) succeed.

PR #702, opened 2023-08-11
Fixes broken Unit 4 & 5 support after hamcrest update from 1.x to 2.x

iloveeclipse:
- additionally avoids returning "null" classpath entries from JUnit 5 container and properly logs if a bundle was missing.

See https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1245

### DIFF
--- a/org.eclipse.jdt.junit.core/src/org/eclipse/jdt/internal/junit/buildpath/BuildPathSupport.java
+++ b/org.eclipse.jdt.junit.core/src/org/eclipse/jdt/internal/junit/buildpath/BuildPathSupport.java
@@ -232,6 +232,8 @@ public class BuildPathSupport {
 
 				return JavaCore.newLibraryEntry(bundleRootLocation, srcLocation, null, getAccessRules(), attributes, false);
 			}
+			String message = "Unable to compute bundle location for '" + bundleId + "' with range " + versionRange;  //$NON-NLS-1$//$NON-NLS-2$
+			JUnitCorePlugin.log(Status.error(message, new IllegalStateException(message)));
 			return null;
 		}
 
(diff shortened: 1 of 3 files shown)


org.eclipse.jdt.junit.core/src/org/eclipse/jdt/internal/junit/buildpath/BuildPathSupport.java
```diff
@@ -303,7 +305,7 @@ public boolean accept(File dir, String name) {
 			"org.junit", new VersionRange("[4.13.0,5.0.0)"), null, "org.junit_4.*.jar", "org.junit.source", "source-bundle/", JUnitPreferencesConstants.JUNIT4_JAVADOC); //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$ //$NON-NLS-4$ //$NON-NLS-5$
 
 	private static final JUnitPluginDescription HAMCREST_CORE_PLUGIN= new JUnitPluginDescription(
-			"org.hamcrest.core", new VersionRange("[1.1.0,2.0.0)"), null, "org.hamcrest.core_1.*.jar", "org.hamcrest.core.source", "source-bundle/", JUnitPreferencesConstants.HAMCREST_CORE_JAVADOC); //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$ //$NON-NLS-4$ //$NON-NLS-5$
+			"org.hamcrest.core", new VersionRange("[2.2.0,2.3.0)"), null, "org.hamcrest.core_2.*.jar", "org.hamcrest.core.source", "source-bundle/", JUnitPreferencesConstants.HAMCREST_CORE_JAVADOC); //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$ //$NON-NLS-4$ //$NON-NLS-5$
```
merks, line 308: You might use VersionRange("[1.1.0,3.0.0)") and "org.hamcrest.core_*.jar" so that people aren't forced to start using org.hamcrest.core 2.2.x...

iloveeclipse: I would first like to get **something** working :-) If you have an idea why test still fail ...

iloveeclipse:
I see `Caused by: java.lang.ClassNotFoundException: org.hamcrest.SelfDescribing` in the SDK but I use already broken SDK :-)

org.eclipse.jdt.junit.core/src/org/eclipse/jdt/internal/junit/buildpath/BuildPathSupport.java
merks: The failures don't give much of a clue do they?  I could try it locally, but I'm rapidly running out of time for today.  😱

merks:
That does sound like this is not on the classpath:

image

org.eclipse.jdt.junit.core/src/org/eclipse/jdt/internal/junit/buildpath/BuildPathSupport.java
iloveeclipse: Aaaa.. Looks like JUnit 4 is broken too :-)

iloveeclipse: @merks : could you please create a dedicated ticket for that? I see your point, but here I want to fix the main issue first.

merks: @iloveeclipse 

I agree that the focus should be to get something working with as few changes as possible.  That's why my phrasing was more meant more as a question than as a suggestion...

iloveeclipse: As you know more about hamcrest 1/2 etc, could you please create a ticket that explains your concerns? If I understood it right, IDE will now only support some tests requiring 2.x hamcrest, but I honestly don't know what IDE should do if there are **two** hamcrests installed in the IDE and how user is supposed to choose which one to use for particular junit container.

merks: @iloveeclipse 

I promise to open a PR with suggested change **after** the tests are working, when I have more time tomorrow.  (If I understand the design correctly, JDT will choose the highest available version that fits in the specified range, just as it does for all the other cases where there might even be multiple versions of junit 4; and then we just need to be sure the jar name glob will expand/match against that selected version.)

merks:
Yeah, it worked! You're awesome!!  And thanks @laeubi  too!   🏆

Issue #705, opened 2023-08-11
Support for either org.hamcrest.core 1.x or org.hamcrest/org.hamcrest.core 2.x


merks:
This constant was changed to accommodate org.hamcrest.core 2.x

https://github.com/eclipse-jdt/eclipse.jdt.ui/blob/7e18055320a51cb8bd946156952b610824951ffd/org.eclipse.jdt.junit.core/src/org/eclipse/jdt/internal/junit/buildpath/BuildPathSupport.java#L307-L308

but now it excludes org.hamcrest.core 1.x from the range.  But org.hamcrest.core alone 1.x is sufficient to satisfy org.junit's bundle require.

On the other hand, org.hamcrest.core 2.x is just a wrapper and has a bundle require on org.hamcrest 2.x, so this constant was added because without this bundle, the org.hamcrest.core bundle 2.x is effectively useless:

https://github.com/eclipse-jdt/eclipse.jdt.ui/blob/7e18055320a51cb8bd946156952b610824951ffd/org.eclipse.jdt.junit.core/src/org/eclipse/jdt/internal/junit/buildpath/BuildPathSupport.java#L310-L311

I.e., the org.hamcrest.core bundle 2.x provides nothing of significance so it doesn't actually need to be on the classpath...

We could change the lower bound of the range of HAMCREST_CORE_PLUGIN here to 1.x:

https://github.com/eclipse-jdt/eclipse.jdt.ui/blob/7e18055320a51cb8bd946156952b610824951ffd/org.eclipse.jdt.junit.core/src/org/eclipse/jdt/internal/junit/buildpath/BuildPathSupport.java#L307-L308

However, if 1.x, rather than 2.x, actually resolves for HAMCREST_CORE_PLUGIN's range, then the org.hamcrest plugin 2.x is no longer needed to make org.junit happy.

But the following "new" logic diagnoses an error for any range that doesn't resolve:

https://github.com/eclipse-jdt/eclipse.jdt.ui/blob/7e18055320a51cb8bd946156952b610824951ffd/org.eclipse.jdt.junit.core/src/org/eclipse/jdt/internal/junit/buildpath/BuildPathSupport.java#L235-L236

[...]

PR #706, opened 2023-08-12
Add either org.hamcrest.core 1.x or org.hamcrest 2.x to junit container

merks:
The org.hamcrest.core 2.x bundle is effectively an empty redirect to org.hamcrest 2.x, so there's no point in putting it on the classpath.

Favor org.hamcrest 2.x over org.hamcrest 1.x if both are present.

Log an error if neither org.hamcrest.core 1.x nor org.hamcrest 2.x are present.

https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/705

## Author checklist

### DIFF
--- a/org.eclipse.jdt.junit.core/src/org/eclipse/jdt/internal/junit/buildpath/BuildPathSupport.java
+++ b/org.eclipse.jdt.junit.core/src/org/eclipse/jdt/internal/junit/buildpath/BuildPathSupport.java
@@ -232,8 +232,6 @@ public class BuildPathSupport {
 
 				return JavaCore.newLibraryEntry(bundleRootLocation, srcLocation, null, getAccessRules(), attributes, false);
 			}
-			String message = "Unable to compute bundle location for '" + bundleId + "' with range " + versionRange;  //$NON-NLS-1$//$NON-NLS-2$
-			JUnitCorePlugin.log(Status.error(message, new IllegalStateException(message)));
 			return null;
 		}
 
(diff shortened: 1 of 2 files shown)


merks:
@iloveeclipse 

I think this is ready to go...

merks:
@iloveeclipse 

This is starting to grow stale and I think we are better off shipping m3 with this issue properly addressed.

iloveeclipse:
> This is starting to grow stale and I think we are better off shipping m3 with this issue properly addressed.

Ed, I do not understand which use case is here fixed. Platform ships with hamcrest 2.x. Why should we try to load 1.x version? What will happen with the code that expects 2.x?

Beside this, we will again "silently" fail if some library is not there / not at expected range.

merks:
> > This is starting to grow stale and I think we are better off shipping m3 with this issue properly addressed.
> 
> Ed, I do not understand which use case is here fixed. 

I did explain the details in the issue, so I'm not sure which parts need to be reiterated?

https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/706

> Platform ships with hamcrest 2.x. Why should we try to load 1.x version? What will happen with the code that expects 2.x?

Hamcrest 2.x is checked first and is preferred, as the commit message states, so maybe the question better asked is what happens if the code expects < 2.0?

The platform ships with junit 4 as well, so why does it support junit 3 when it doesn't ship that?  I don't understand that either...

> 
> Beside this, we will again "silently" fail if some library is not there / not at expected range.

This is not a silent failure;
```
			String message = "Either org.hamcrest.core 1.x or org.hamcrest 2.x must be present.";  //$NON-NLS-1$
				JUnitCorePlugin.log(Status.error(message, new IllegalStateException(message)));
```

In any case, if you feel it's better and less risky to stop supporting something that has always worked in the past and still works today, that's of course your choice.  You might have mentioned that in the issue and you might have mentioned it several days ago given that I do have other/better things on which to spend my time.   One word of warning though, the constraint on HAMCREST_PLUGIN is too strict <= 2.3.0 and will definitely bite you in the future.

iloveeclipse:
> Hamcrest 2.x is checked first and is preferred, as the commit message states, so maybe the question better asked is what happens if the code expects < 2.0?

Assuming user will **add** hamcrest 1.x to the IDE, 2.x will win anyway. 

So only few products that have jdt.ui will be able to configure themselves (how?) to have hamcrest 2.x be replaced by 1.x  (is this possible without side effects?).

I see hamcrest.core 2.x being a part of the org.eclipse.jdt feature:
image

So to make the product that contains `org.eclipse.jdt` feature that includes `hamcrest.core` one would need to uninstall that bundle first, which is probably not trivial because it is included in the feature?

So the product would need to manually add all jdt bundles except from `org.eclipse.jdt` feature except hamcrest?

Other way around: what would happen if we **revert** the order and let hamcrest 1.x "win" by default? If it is not there, nothing changes for most of the SDK consumers because SDK ships 2.x only. If someone needs hamcrest 1.x to get code that requires hamcrest 1.x running, it will be possible without extra product, just by installing hamcrest 1.x into existing product?

> This is not a silent failure;

For the **hamcrest only**. All other entries can now silently fail to resolve. Sure, this was always the case till my recent commit, but if you could change the code to report (not expected) fails, we could merge this PR.

PR #709, opened 2023-08-14
Build artifacts should be ignored from git

robstryker:
## What it does

Build artifacts should not be seen as dirty or new or untracked. They should be ignored

## How to test

Run the build. Then run git status. 

## Author checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,6 @@ Snap.*
 # maven
 /*/target/
 .DS_Store
+org.eclipse.jdt.ui.examples.projects/archive/junit/org.eclipse.jdt.ui.junit.sampleproject-*.jar
+org.eclipse.jdt.ui.examples.projects/jdtexamples.jar
+


robstryker:
I recognize that some of these artifacts are from me using maven wrapper, but others are not. Still, since I was told using maven wrapper is a good idea to use specific versions of maven to ensure a build, it seems worthwhile to add those to the ignore list.

rgrunber:
Would you consider the following a workaround in order to not add the "mvn wrapper" stuff into `.gitignore` ? I only suggest this because it doesn't look like JDT UI has really ever taken into account user workflows in the build process :

If you've used the Maven wrapper once and it's generated the `.mvn/` , `mvnw`, `mvnw.cmd` files, and you've run a build, the Maven distribution should be downloaded to `$HOME/.m2/wrapper/dists/apache-maven-${version}-bin/`. Then you can just create a symbolic link/shortcut from somewhere on your path to the maven instance on in that local repository location. For example I have :

`/home/rgrunber/.local/bin/mvn -> /home/rgrunber/.m2/wrapper/dists/apache-maven-3.9.3-bin/54p6kc7m0ap3os8bc4ge28ia2/apache-maven-3.9.3/bin/mvn` . With that I don't really need the wrapper data in the project folder as I've got Maven 3.9.3. Yes, it wouldn't allow you to configure individual projects with different Maven versions, but I think for practically all Eclipse projects, 3.9.x is the minimum (because of Tycho 4.x) which is satisfied.

robstryker:
I'm open to workarounds for the `mvnw` issue, but the two build artifacts definitely need to be ignored:

```
org.eclipse.jdt.ui.examples.projects/archive/junit/org.eclipse.jdt.ui.junit.sampleproject-1.0.0-SNAPSHOT-dist.jar
org.eclipse.jdt.ui.examples.projects/jdtexamples.jar
```

rgrunber:
Agreed. Clearly [org.eclipse.jdt.ui.examples.projects is part of the build](https://github.com/eclipse-jdt/eclipse.jdt.ui/blob/8234af07ce40f708b34b4d48b195f0d4e4dc8402/pom.xml#L63).

The artifacts are generated by :

https://github.com/eclipse-jdt/eclipse.jdt.ui/blob/8234af07ce40f708b34b4d48b195f0d4e4dc8402/org.eclipse.jdt.ui.examples.projects/build.properties#L22

https://github.com/eclipse-jdt/eclipse.jdt.ui/blob/8234af07ce40f708b34b4d48b195f0d4e4dc8402/org.eclipse.jdt.ui.examples.projects/pom.xml#L32-L58

I think we should merge this once 4.30 opens up (for M1).

rgrunber (approved):
I'd be fine with merging this once 4.30 opens up.

PR #713, opened 2023-08-14
Remove ITextSelection references from some refactors

robstryker:
## What it does

The refactors themselves should not require UI, whenever possible, while the actions that call them are inherently UI creatures and should have access to the UI classes. 

This patch removes some utility functions which help create SurroundWithTryCatchRefactoring.java and SurroundWithTryWithResourcesAction.java, and ask the action to provide the various required integers themselves instead of passing in a UI element. 

This will help separate core and UI more effectively in the near future. 

## Author checklist


rgrunber (approved):
Seems reasonable. There's already helper methods for offset/length so just using those allows us to avoid some code that's in the UI.

I would probably merge once 4.30 is open.

PR #714, opened 2023-08-14
Extracting RefactorSaveHelper constants into core

robstryker:
## What it does

This patch extracts some constants from RefactoringSaveHelper, specifically the return values of the user prompt. While the refactors themselves do want to know the result of a user prompt, the fact is that a user prompt may be from eclipse UI / swt, or it may be from a CLI (if one is attempting to use jdt.core.manipulations in some type of headless application). 

By extracting this interface and decoupling it from a UI class, we allow other applications (like JDT LS) to make use of these refactors without needing to clone the class. 

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.manipulation/META-INF/MANIFEST.MF
+++ b/org.eclipse.jdt.core.manipulation/META-INF/MANIFEST.MF
@@ -71,5 +71,6 @@ Export-Package: org.eclipse.jdt.core.manipulation,
  org.eclipse.jdt.internal.ui.text.correction;x-friends:="org.eclipse.jdt.ui,org.eclipse.jdt.debug.ui",
  org.eclipse.jdt.internal.ui.text.correction.proposals;x-friends:="org.eclipse.jdt.ui",
  org.eclipse.jdt.internal.ui.text.template.contentassist;x-friends:="org.eclipse.jdt.ui",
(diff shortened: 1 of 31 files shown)


jjohnstn (commented):
Minor nit: JDT uses x= y format instead of x = y.  Assuming you don't have another change that does a minor bump this will require a minor version bump to jdt.core.manipulation due to the new public interfaces.  This bump isn't done as part of the initial 4.30 version bumps.  Should be fine for 4.30 M1.

robstryker:
I've fixed the assignment formatting. I will ensure a version bump occurs but am hesitant to put it any specific PR at this moment due to the stack of commits / rebasing / etc. But if you tell me it should be part of this one, I'll do it. Otherwise I prefer to wait until the end of the week.

jjohnstn:
I'll handle the version bump for you.  I am checking in other patches and if the minor bump hasn't occurred yet, I'll just add it.

robstryker:
I had a version bump in here but then api-tools complained that it wasn't actually API and so i shouldn't have bumped versions.

robstryker:
Seems it's in an xfriends package and not an exported package?  So api-tools said 'dont upversion it'

jjohnstn (approved):
Looks good, merging.

PR #716, opened 2023-08-14
Migrate core extension classes to use the Core version of LinkedProposalModel and associated classes

robstryker:
## What it does

LinkedProposalModel and associated classes already have Core versions. They are nearly identical except that the LinkedProposalPositionGroup.Proposal class includes an option to provide an image. 

To the best of my ability, I have ensured that only callers that create proposals with no images have been migrated. UI elements that live in clearly UI-based folders and that make use of a Proposal image have been left untouched, but I appreciate a code review to ensure I didn't miss anything. 

Whenever possible, the superclasses in Core are preferred in API locations. This may cause API compatibility errors for some illegal extenders. 

## Author checklist

### DIFF
--- a/org.eclipse.jdt.ui.tests/ui/org/eclipse/jdt/ui/tests/quickfix/QuickFixTest.java
+++ b/org.eclipse.jdt.ui.tests/ui/org/eclipse/jdt/ui/tests/quickfix/QuickFixTest.java
@@ -64,9 +64,9 @@ import org.eclipse.jdt.core.dom.TypeDeclaration;
 import org.eclipse.jdt.core.dom.VariableDeclarationFragment;
 
 import org.eclipse.jdt.internal.core.manipulation.dom.ASTResolving;
-import org.eclipse.jdt.internal.corext.fix.LinkedProposalModel;
-import org.eclipse.jdt.internal.corext.fix.LinkedProposalPositionGroup;
-import org.eclipse.jdt.internal.corext.fix.LinkedProposalPositionGroup.Proposal;
+import org.eclipse.jdt.internal.corext.fix.LinkedProposalModelCore;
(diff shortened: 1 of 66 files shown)


robstryker:
Just to discuss:  the second commit in this patch fixes a bad assumption in `CompilationUnitRewriteOperation` and all of its subclasses, namely, that the linked model is of the more specific type `LinkedProposalModel` instead of the more generic `LinkedProposalModelCore`

This class, `CompilationUnitRewriteOperation`, has its main entry point of `rewriteAST(CompilationUnitRewrite cuRewrite, LinkedProposalModelCore linkedModel)` as well as an abstract function for subclasses to implement: `rewriteAST(final CompilationUnitRewrite cuRewrite, final LinkedProposalModel linkedModel)`. 

There are a few problems here, or at least things that make this more annoying. First, these two functions share almost the same signature, except that the main entry accepts `LinkedProposalModelCore` and the abstract method for subclasses to implement takes `LinkedProposalModel`.  Having these two functions share the same name isn't that bad, but having the second parameter be ambiguous, where one is a subtype of the other, is annoying for code cleanliness. 

Second, the current implementation does a naive cast from `LinkedProposalModel` to `LinkedProposalModelCore` without doing any checks.  This obviously fails all the time. 

The second commit here removes the overloaded `rewriteAST` abstract function and replaces it with one named `rewriteASTInternal`, taking a `LinkedProposalModelCore` as a parameter. While this does require changing ALL subclasses of this class (and there are MANY), I still believe it is the right solution. 

Since this class is internal, I believe all of these changes should be valid. The build for this PR has not complained about changing or breaking any API. So unless someone objects, I believe this is the right solution.

jjohnstn (approved):
Looks good.

PR #724, opened 2023-08-17
NullAnnotationsRewriteOperations: OutOfBoundsException: Index 1 out of bounds for length 1

### LINKED ISSUE #1294: Interrupted while waiting for CSS fragments cache lock, proceeding without using cache
Using I20240327-1800 build I randomly get errors without stack trace while working in Java editor:

`Interrupted while waiting for CSS fragments cache lock, proceeding without using cache`

This is a regression from https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/1074.

@RedeemerSK: please check. 
- First of all, if errors are logged, they should be logged with a stack trace.
- I assume the error is just caused by the fact that while editing or navigating inside Java file, mouse hover request may be cancelled because of some user actions that make the request obsolete. This is not an error and shouldn't be spamming the log.

stephan-herrmann:
fixes [#1294](https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/725)

## What it does
Ensure that parameters are looked up within the actual declaring list, even if that is from an outer enclosing method.


stephan-herrmann:
Not a regression, so the prepared fixes (jdt.core and jdt.ui) will have to wait until 4.30 M1 (milestone not yet available in jdt.ui).

PR #726, opened 2023-08-17
Abstract out text search to break dependency with org.eclipse.search, a ui component

robstryker:
## What it does

Abstract out access to org.eclipse.search, which is a UI component, so that jdt.manipulations can have all of the core-refactor classes and still compile. 

Anyone depending on a headless jdt without jdt.ui would have to implement their own search engine, at least until org.eclipse.search has their core nad UI separated a bit better. 

## Author checklist


robstryker:
This PR needs to be re-evaluated. A different design may be required.

PR #730, opened 2023-08-21
Version bumps for workbench & IDE related bundles

iloveeclipse:
Updated dependencies to match most recent major.minor versions from dependent bundles to make sure whatever API was added and in whichever combination p2 installs latest dependencies containing most recent API level.

See https://github.com/eclipse-platform/eclipse.platform.ui/issues/1051


iloveeclipse:
@jjohnstn : for RC1 we need an additional committer to review the change. Could you please review this PR?

The point why [Oomph fails](https://github.com/eclipse-platform/eclipse.platform.ui/issues/1051) is: if p2 based application (like Oomph) tries to compute "minimal required set of installable units" for JDT UI, it should not take "some" matching versions, but latest we test & provide in the current release. 

The original issue should have been (I believe, because of transitive dependencies) fixed via https://github.com/eclipse-platform/eclipse.platform.ui/pull/1052, this change is additionally to avoid other possible not yet detected problems in same "IDE/UI" area.

Otherwise we might install "good enough" versions from p2 of view but actually an untested combination of dependent bundles in some (not latest) versions that match requested ranges.

Note, this PR does not yet require [updated `org.eclipse.ui 3.204.0` version](https://github.com/eclipse-platform/eclipse.platform.ui/pull/1052/commits/74be66e47a3cf1e4e1b89d0b2fdde1366a5bd2c4), because it is not yet merged and so not yet visible in the platform. I plan to create a second PR that would do this once we have SDK build available.

jukzi:
lgtm

PR #640, opened 2023-06-21
Add new hover for Java hex, octal, and binary constants

jjohnstn:
- add new JavaConstantHover class
- add new hover to o.e.jdt.ui plugin.xml
- add new hover property strings
- fixes #639
- for Bug 559614

## What it does

Adds new hover to use on hex, octal, and binary constants to see value.

## How to test

See bug.

## Author checklist

- [X] I have thoroughly tested my changes
- [X] The change is following the [coding conventions](https://wiki.eclipse.org/Platform/How_to_Contribute#Coding_Conventions)
- [X] I have signed the [Eclipse Contributor Agreement (ECA)](https://www.eclipse.org/legal/ECA.php)

PR #742, opened 2023-08-31
Improve hover for Java hex, octal, and binary constants

howlger:
Show hovers also in the following cases:
```java
int bin1 = 0B11111111; // uppercase "0B"
int bin2 = 0b1111_1111; // binary with an underscore
int oct = 0_377;  // octal with an underscore
int hex = 0xff_ff_ff; // hex with underscores
```

Do not show hovers for doubles starting with "0", e.g.:
  double d= 01e6D;

This does not fix the issue that hovers are shown in the following cases where they are not make sense:
```java
double d2= 012.045;
```
- First hover for "012": "10 : [0xa]"
- Second hover for "045": "37 : [0x25]"

See also https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/640

### DIFF
--- a/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/text/java/hover/JavaConstantHover.java
+++ b/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/text/java/hover/JavaConstantHover.java
@@ -33,20 +33,23 @@ public class JavaConstantHover extends AbstractJavaEditorTextHover {
 			}
 			if (hoverSource != null && hoverSource.startsWith("0")) { //$NON-NLS-1$
 				try {
-					long longValue= hoverSource.startsWith("0b") ? //$NON-NLS-1$
-							Long.parseLong(hoverSource.substring(2), 2) : Long.decode(hoverSource).longValue();
+					long longValue= hoverSource.startsWith("0b") || hoverSource.startsWith("0B") ? //$NON-NLS-1$ //$NON-NLS-2$
+							Long.parseLong(withoutUnderscoreInfixes(hoverSource.substring(2)), 2)
+							: Long.decode(withoutUnderscoreInfixes(hoverSource)).longValue();
 					return "<body><p>" + Long.toString(longValue) + "<b> : [0x" + Long.toHexString(longValue) + "]</p></body>"; //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$
 				} catch (NumberFormatException e) {
-					try {
-						double doubleValue= Double.valueOf(hoverSource).doubleValue();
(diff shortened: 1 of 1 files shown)


howlger:
@jjohnstn Could you please look into this?

jjohnstn (approved):
Looks good @howlger.  Will merge for 4.30 M1

jjohnstn:
Will look into fixing the remaining double issue.

jjohnstn:
Hi @howlger I fixed the double issue and added back a previous feature.  My initial code handled doubles specified in hex format (e.g. 0x1.0p2) which evaluates to 4.0.  I guess with various other fixing, it lost that support.  I added fixes to the code to add more characters to the hover region which stopped on either side of the decimal point.  Let me know if you find anything else in your testing.

howlger:
I see, I was not aware of the full [integer](https://docs.oracle.com/javase/specs/jls/se17/html/jls-3.html#jls-3.10.1) and [floating-point literals](https://docs.oracle.com/javase/specs/jls/se17/html/jls-3.html#jls-3.10.2) specification.

We still have some cases missing, e.g.:
-  `0Xff00p0d` - uppercase `X`
- `0x7fff_ffff_ffff_ffffl` - `long` with underscores
- `0x10L` - hex with the type suffix `l` or `L`
- `010L` - octal with the type suffix `l` or `L`
- ...?

I have not yet tested it, but maybe there are still cases where a hover is shown by mistake:
```java
double d1= 012.045; // no hover
double d2= 010D; // no hover; = 10 (prefix '0' ignored), but 010L = 8 ('0' means octal)
double d3= 0xff.0025p-08d; // hover: 0.9960959553718567
float f1= 02.05e00f; // no hover
float f2= 02.05e02f; // no hover
float f3= 02.05e+02f; // no hover
float f4= 02.05e-02f; // no hover
```

If and only if the hover is not show for invalid literals, e.g. not for `0x_ff` or `0xff_`, then `withoutUnderscoreInfixes(...)` can be simplified to `replace("_", "")`.

jjohnstn:
Hi @howlger I fixed the upper-case/lower-case issues by lower-casing the string to begin with.  I fixed the l/L suffix by looking for it and removing prior to conversion since the Long class doesn't support it when parsing (weird if you ask me since float and double support their suffixes).  Let me know if there are any other tweaks you want made, otherwise I'll check this in for 4.30 M1.  I don't think the hover should show invalid constants so your underscore replacement logic is good.

jjohnstn:
Merging.  If there are issues, please open a separate issue.

howlger:
Thanks for fixing/improving and merging it!

I verified it in `eclipse-SDK-I20230917-1800`.

PR #763, opened 2023-09-07
Remove bold settings from dark theme CSS

### LINKED ISSUE #756: Java syntax colors and settings don't save
_Originally reported in [Bug 486992](https://bugs.eclipse.org/bugs/show_bug.cgi?id=486992) and more recently in https://github.com/eclipse-packaging/packages/issues/54_

When in dark theme (but perhaps other cases too based on comments) when unchecking Bold in `Keywords, excluding 'return'` it's not maintaining the bold font I set.

@valmirserpa kindly made a screen recording in https://github.com/eclipse-packaging/packages/issues/54#issuecomment-1709163594 which I embed here as the original was not necessarily stored in a permanent location:

https://github.com/eclipse-jdt/eclipse.jdt.ui/assets/679236/c7f8d445-6683-409e-b6f2-7df0febc5ab9

jjohnstn:
- fixes #756

## What it does

Removes bold settings from dark theme CSS so user can set them.

## How to test

See issue.

## Author checklist


jjohnstn:
@vogella Could you take a look?  The issue is that some preferences are being specified in the css file and if the user changes them, they get reset by the css file.  I changed it so only colors are specified, but don't use dark theme so can't say if the change causes any issue for end-users.

vogella:
@jjohnstn sorry I recently had to switch laptops for repair and have no development setup for JDT and therefore cannot test. If the dark theme still looks like for a new Java file in a new workspace, I think the change is OK.

jjohnstn:
Ok, thanks @vogella The real change is all the additional bells and whistles that aren't enabled by default.  The old CSS turned almost all of them on.  Any enablement settings current users have shouldn't be altered by updating to the change.  A brand new workspace just doesn't have the optional things enabled by default (such as additional colors for enum or abstract classes) but they are easily settable and the settings are preserved/stored across sessions.  I'll check it in for M1 and see if anybody has issues.

iloveeclipse:
I've removed M1 milestone, that is long overdue. Feel free to set any reasonable milestone instead.

jukzi:
ping @vogella

vogella:
> ping @vogella

Why this ping? I think @jjohnstn wanted to check.

jukzi:
> Why this ping? I think @jjohnstn wanted to check.

oh, ok -> ping @jjohnstn

akurtakov:
Jeff, what is the status of this one?

jjohnstn:
@akurtakov I initially had a change which simply removed the bold settings from the dark theme CSS and then I also removed some more non-color settings, but it changes the behavior when a user switches to dark mode the first time since some bold settings are not set by default so I wasn't sure this was going to go over with end-users.  I was hoping to eventually see if settings could be or were already separated for the themes.  It makes sense that the settings should be for a theme and not global.  I just never got around to it.

Issue #61, opened 2022-05-20
typehierarchy: UI Freezes till Indexing done


jukzi:
reproduce: 
1. press "Rebuild Index" in the Java Preferences
2. in a java editor press CTRL+T

=> Mouse icon changes to busy and UI freezes till indexing finished. Impossible to cancel. 
Expected: show a cancelable(!) "wait till indexing finished" Dialog

```
"main" #1 prio=6 os_prio=0 cpu=128937.50ms elapsed=1278.33s tid=0x000001580d59f730 nid=0xe70 runnable  [0x0000000f8a8fb000]
   java.lang.Thread.State: RUNNABLE
        at org.eclipse.swt.internal.win32.OS.WaitMessage(Native Method)
        at org.eclipse.swt.widgets.Display.sleep(Display.java:4757)
        at org.eclipse.jface.operation.ModalContext$ModalContextThread.block(ModalContext.java:167)
        at org.eclipse.jface.operation.ModalContext.run(ModalContext.java:368)
        at org.eclipse.ui.internal.WorkbenchWindow.lambda$7(WorkbenchWindow.java:2345)
        at org.eclipse.ui.internal.WorkbenchWindow$$Lambda$2406/0x0000000802052330.run(Unknown Source)
        at org.eclipse.swt.custom.BusyIndicator.showWhile(BusyIndicator.java:74)
        at org.eclipse.ui.internal.WorkbenchWindow.run(WorkbenchWindow.java:2343)
        at org.eclipse.jdt.internal.ui.typehierarchy.TypeHierarchyLifeCycle.ensureRefreshedTypeHierarchy(TypeHierarchyLifeCycle.java:212)
        at org.eclipse.jdt.internal.ui.typehierarchy.TypeHierarchyLifeCycle.ensureRefreshedTypeHierarchy(TypeHierarchyLifeCycle.java:161)
        at org.eclipse.jdt.internal.ui.typehierarchy.HierarchyInformationControl.setInput(HierarchyInformationControl.java:275)
        at org.eclipse.jface.text.AbstractInformationControlManager.internalShowInformationControl(AbstractInformationControlManager.java:1151)
        at org.eclipse.jface.text.AbstractInformationControlManager.presentInformation(AbstractInformationControlManager.java:1120)
        at org.eclipse.jface.text.AbstractInformationControlManager.setInformation(AbstractInformationControlManager.java:431)
[...]

PR #772, opened 2023-09-12
fix StatusLine progress

jukzi:
https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/61


jukzi:
see https://github.com/eclipse-platform/eclipse.platform.ui/pull/1077#issuecomment-1714948144

PR #781, opened 2023-09-13
Migrate usage to logic in core classes when applicable

### LINKED ISSUE #695: Have "jdt.ui/core refactoring" use core interfaces or core base classes whenever possible
Many times, when getting a string representation of a java element, often just for error logging, jdt.ui is using the class JavaElementLabels or BindingLabelProvider  instead of the JavaElementLabelsCore or BindingLabelProviderCore classes. 

The core classes live in jdt.core.manipulation or sometimes in jdt.core. Pushing non-UI logic down the stack, and migrating usage to favor the lower level classes, will serve to unify logic and make more of the code accessible to headless plugins or headless consumers in the long run, specifically for clients like jdt.ls. 

While migrating non-UI code to lower level plugins may be a desirable long term goal, the task should be performed slowly and carefully. Migrating usage to non-UI bundles is a good incremental step to prepare for the day when much of the main logic currently in jdt.ui can be moved down. It also is just good code maintenance.

robstryker:
- Fixes #695

## What it does

Many classes use JavaElementLabels, BindingLabelProvider, or similar. Core classes with substantially the same functionality exist and should be preferred in most cases. 

Changes like this will make it easier to move many refactoring or proposal classes into a core-only bundle in the future, which is critical for headless consumers like jdt.ls. 

## Author checklist

### DIFF
--- a/org.eclipse.jdt.ui/core refactoring/org/eclipse/jdt/internal/corext/refactoring/changes/CompilationUnitReorgChange.java
+++ b/org.eclipse.jdt.ui/core refactoring/org/eclipse/jdt/internal/corext/refactoring/changes/CompilationUnitReorgChange.java
@@ -32,7 +32,7 @@ import org.eclipse.jdt.core.JavaCore;
 import org.eclipse.jdt.internal.corext.refactoring.reorg.INewNameQuery;
 import org.eclipse.jdt.internal.corext.util.JavaElementResourceMapping;
 
-import org.eclipse.jdt.ui.JavaElementLabels;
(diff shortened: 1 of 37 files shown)


rgrunber:
As it happens, I'm doing something [very similar in JDT-LS](https://github.com/eclipse-jdtls/eclipse.jdt.ls/pull/2854/commits/d160a5f9d77397203c5fcdff99d86f9c0f0f68a3) so this is very likely the right thing to do .

rgrunber (approved):
Looked through it. Couldn't find anything wrong.

PR #782, opened 2023-09-14
JavaElementImplementationHyperlink: fix progress during newTypeHierarchy

jukzi:
newTypeHierarchy is a JavaModelOperation which uses SubMonitor.convert since https://bugs.eclipse.org/bugs/show_bug.cgi?id=477790 That is not allowed when beginTask is already called. See SubMonitor.convert javadoc.

https://github.com/eclipse-platform/eclipse.platform.ui/issues/1107

### DIFF
--- a/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/javaeditor/JavaElementImplementationHyperlink.java
+++ b/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/javaeditor/JavaElementImplementationHyperlink.java
@@ -269,15 +269,13 @@ public class JavaElementImplementationHyperlink implements IHyperlink {
 				}
 				try {
 					String typeLabel= JavaElementLabels.getElementLabel(type, JavaElementLabels.DEFAULT_QUALIFIED);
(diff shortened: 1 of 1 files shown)


org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/javaeditor/JavaElementImplementationHyperlink.java
```diff
@@ -269,15 +269,13 @@ public void acceptSearchMatch(SearchMatch match) throws CoreException {
 				}
 				try {
 					String typeLabel= JavaElementLabels.getElementLabel(type, JavaElementLabels.DEFAULT_QUALIFIED);
-					monitor.beginTask(Messages.format(JavaEditorMessages.JavaElementImplementationHyperlink_search_method_implementors, typeLabel), 10);
+					monitor.setTaskName(Messages.format(JavaEditorMessages.JavaElementImplementationHyperlink_search_method_implementors, typeLabel));
```
iloveeclipse, line 272: Do you actually **see** the task name on the status line?
I don't see it neither with nor without this patch.

jukzi: i don't see it either

PR #801, opened 2023-09-18
Move constants necessary for proposals into jdt manipulations, and small cleanup

robstryker:
## What it does

Moving on to migrating the refactors portion of jdt.ui into being usable by core-only consumers, the classes in this pull request were deemed as also requiring to be moved down into jdt.manipulations beforehand. 

Other than moving some constants down, or the creation of a utility class to convert / wrap / unwrap between the core or ui versions of an interface, the biggest change here is with IProblemLocation, which was deemed to have the exact same interface as IProblemLocationCore, a completely unnecessary split. While I haven't gone ahead and removed the duplicative interface, I have made one extend the other. 

While in the future, the complete removal of one or the other will be preferred, trying to decide which should stay or go is a debate for another time. 

## Author checklist


jjohnstn:
Hi @robstryker You cannot delete IProblemLocation.  It is public API.  This requires a major version bump which we can't do out of the blue.  I can't remember exactly but I believe we couldn't move the public interface into jdt.core.manipulation which is why we did the duplication and had both ProblemLocation and ProblemLocationCore.

robstryker:
> Hi @robstryker You cannot delete IProblemLocation. It is public API. This requires a major version bump which we can't do out of the blue. I can't remember exactly but I believe we couldn't move the public interface into jdt.core.manipulation which is why we did the duplication and had both ProblemLocation and ProblemLocationCore.

Well, the public interface is now inside manipulations as well, as of this PR here, so I think we can probably remove the Core interface in the future instead. 

This PR passes green (at least in github PR build) and so it seems that moving IProblemLocation into manipulations has worked just fine? 

If this PR gets merged, then the core version is no longer necessary and can be entirely replaced with the public API instead.

rgrunber:
As long as the API baseline doesn't complain (about the package move) it should be fine though I seem to recall a conversation I had with @mickaelistria where he mentioned such changes are considered API breaking.

mickaelistria:
AFAIK, the only way to move an API upstream in the dependency would be to re-export the dependency and/or the package from upstream. However here I don't think it's feasible because of the split package.
I would suggest to just create a new package in manipulation (it will also be cleaner), with the same API and make the legacy API become empty and just implement the newer one and mark it `@Deprecated`

PR #803, opened 2023-09-19
Support 'Show In' similar to 'Link with Editor'.

raghucssit:
'Link with Editor' tries to select visible parent in case of element being linked is not created yet on tree viewer. Similar implementation is used for 'Show In'

Fixes https://github.com/eclipse-platform/eclipse.platform.ui/issues/1014

### DIFF
--- a/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/packageview/PackageExplorerPart.java
+++ b/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/packageview/PackageExplorerPart.java
@@ -1355,6 +1355,22 @@ public class PackageExplorerPart extends ViewPart
             if (revealElementOrParent(element))
 	            return IStatus.OK;
         }
+
+        //we tried removing filters, still element was not found. we can try to select any possible element in it's parent hierarchy. Like 'Link with Editor'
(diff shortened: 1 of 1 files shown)


iloveeclipse:
@raghucssit : could you please provide detailed steps how to verify this patch? I've tried it on latest master and I don't see any differences compared to the master state.

raghucssit:
> @raghucssit : could you please provide detailed steps how to verify this patch? I've tried it on latest master and I don't see any differences compared to the master state.

In Java Editor select a Class name and then Show In->Package Explorer context menu. if it is not visible and some filters are applied  then below warning will be displayed.
show-in-warn
On select 'Yes' it will remove filters and try to reveal. If it is still not visible then this patch tries to calculated visible item in parent hierarchy which will finally an ExpandableNode if item is inside that.

raghucssit:
If we have the item to be selected hidden in 2 or more hidder hierarchy this patch will select the visible parent ExpandableNode.
Example:
limit = 5 and 10 projects
element to select is `class10` like below 
proj10
  pack10
     class10
 This patch will select ExpandableNode in the top level project.
Without this patch it returns IStatus.ERROR

iloveeclipse:
Ok, I see now.

iloveeclipse (approved):

PR #814, opened 2023-09-21
Move constants necessary for proposals into jdt manipulations, and cleanup

robstryker:
## What it does

Moves some constants into jdt manipulations from jdt ui. These constants are pre-requisites to many proposals being moved down into jdt.manip, which will help various headless consumers of jdt like jdt.ls. 

## Author checklist

- [x] I have thoroughly tested my changes
- [x] The change is following the [coding conventions](https://wiki.eclipse.org/Platform/How_to_Contribute#Coding_Conventions)
- [x] I have signed the [Eclipse Contributor Agreement (ECA)](https://www.eclipse.org/legal/ECA.php)

PR #813, opened 2023-09-21
Standardize internal code on IProblemLocationCore

robstryker:
## What it does

This standardizes all internal code on IProblemLocationCore whenever possible, while leaving the existing API and extension point interfaces unchanged. Interactions with those interfaces will now wrap all core versions before passing them on to extension points. 

This issue is a requirement to move substantially all proposals into jdt.manipulations, which will help headless consumers like jdt.ls. 

## Author checklist

- [x] I have thoroughly tested my changes
- [x] The change is following the [coding conventions](https://wiki.eclipse.org/Platform/How_to_Contribute#Coding_Conventions)
- [x] I have signed the [Eclipse Contributor Agreement (ECA)](https://www.eclipse.org/legal/ECA.php)

PR #815, opened 2023-09-21
Proposals branch

robstryker:
## What it does

Splits the proposal logic into core and UI. It moves the vast majority of the logic into the core classes, while maintaining the UI as existing wrappers for stability. 

This patch also creates an IInvocationContextCore class and makes it the default interface for use in the core objects, in the same way as #813 made the core classes use the relevant Core interface. Just as in the problem location situation, the UI's API interface extends the hidden non-API core one. 

This patch depends on #813  and #814 

## Author checklist


robstryker:
This patch is very difficult to proof-read, due to the fact that the bulk (but not the entirety) for about 37 classes have been moved into their own new unique internal core classes. This means the vast majority of the patch is removals of 90% of one file, and creation of that same code in a new file. 

I'm not sure how to improve this patch to make it more easily followable and to verify that no logic changes have occurred hidden in the patch. =/

jjohnstn:
Hi @robstryker There are new conflicts to fix for this patch.

robstryker:
Squashed and rebased. Looking for a final green build.

org.eclipse.jdt.core.manipulation/proposals/org/eclipse/jdt/internal/ui/text/correction/AddAllMissingJavadocTagsProposalCore.java
```diff
@@ -0,0 +1,180 @@
+package org.eclipse.jdt.internal.ui.text.correction;
```
jjohnstn, line 1: Missing license.  A few others are missing licenses as well.  Other than that, looks good.

jjohnstn (changes requested):

jjohnstn:
@robstryker Please also rebase.  It would be good if you can get this in before Monday where large patches are frowned upon.

jjohnstn (approved):

jjohnstn:
Jenkins appears to be off so new build is not occurring.  Last change built successfully and the only change since then was the addition of copyright headers done my myself.  This refactoring change should go into M1 as it is too big for M2 or M3 so I am merging this before the Monday deadline which requires no large changes.

jjohnstn:
@iloveeclipse please note above.

PR #875, opened 2023-10-23
Fix StringBuffer to StringBuilder cleanup

### LINKED ISSUE #869: StringBuffer to StringBuilder cleanup failure
The following scenario:

```
import java.io.StringWriter;

public class TestStringBufferCleanup {
	
	public void foo() {
		StringWriter stringWriter = new StringWriter();
		StringBuffer buffer = stringWriter.getBuffer();
		System.out.println(buffer.toString());
	}

}
```

when the StringBuffer to StringBuilder cleanup is applied creates a failure because buffer is converted to StringBuilder even though StringWriter.getBuffer() returns a StringBuffer.

jjohnstn:
- fixes #869
- add logic to handle StringWriter.getBuffer()
- add new test scenario to CleanUpTest1d8

## What it does

Fixes StringBuffer to StringBuilder clean up to handle StringWriter.getBuffer().

## How to test

See issue or new test scenario.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/fix/StringBufferToStringBuilderFixCore.java
+++ b/org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/fix/StringBufferToStringBuilderFixCore.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2021 Red Hat Inc. and others.
+ * Copyright (c) 2021, 2023 Red Hat Inc. and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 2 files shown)


HannesWell:
Would it be possible to make this more generic and check if a method is called that returns a `StringBuffer` and is not changed in the current clean-up or at least cannot be changed because it is not in the workspace?
There could also be external libraries that return a `StringBuffer`.

jjohnstn:
> Would it be possible to make this more generic and check if a method is called that returns a `StringBuffer` and is not changed in the current clean-up or at least cannot be changed because it is not in the workspace? There could also be external libraries that return a `StringBuffer`.

I like the idea.  The list of all Compilation Units involved in the cleanup is not available to the core fix (it sees just the one CU).  I'll see how to determine that the class is not in the workspace as I've not come across this yet.

jjohnstn:
@HannesWell I modified the logic so that it will wrap any method invocation that returns StringBuffer that is binary only and isn't coming from StringBuffer itself.

HannesWell:
> @HannesWell I modified the logic so that it will wrap any method invocation that returns StringBuffer that is binary only and isn't coming from StringBuffer itself.

Sounds good. Thanks:)

PR #883, opened 2023-10-29
Add API filters

fedejeanne:
From `JUnitRemoteTestRunnerClient` to `RemoteTestRunnerClient`, both in the same plugin.


fedejeanne:
There's been no activity on this one. Could someone please review it?

jjohnstn:
Looks fine.  Any reason you didn't also add a filter for JUnitTestPlugin.JUnitVersion.getJUnitTestKind()?

fedejeanne:
@jjohnstn Thank you for reviewing it! 

> Any reason you didn't also add a filter for JUnitTestPlugin.JUnitVersion.getJUnitTestKind()?

I was initially baffled about that one since `ITestKind` is part of the internal API but its containing package is "friends" with the bundle `org.eclipse.jdt.ui.unittest.junit` which should have made the warning go away. Since the warning didn't go away, I can only assume that the warning means _"giving an internal API as a result to the outside world makes the outside world rely on internal APIs"_ for which the right solution should be to move `ITestKind` to an "official" (_i.e._ **not** internal) package instead of silencing the warning.

Or am I seeing this all wrong?

jjohnstn:
> > Any reason you didn't also add a filter for JUnitTestPlugin.JUnitVersion.getJUnitTestKind()?
> 
> I was initially baffled about that one since `ITestKind` is part of the internal API but its containing package is "friends" with the bundle `org.eclipse.jdt.ui.unittest.junit` which should have made the warning go away. Since the warning didn't go away, I can only assume that the warning means _"giving an internal API as a result to the outside world makes the outside world rely on internal APIs"_ for which the right solution should be to move `ITestKind` to an "official" (_i.e._ **non**-internal) package instead of silencing the warning.
> 
> Or am I seeing this all wrong?

No.  It is a mess.  The original code should not have added the public method in the public enum and instead only returned the String field and let the callers make the call to TestKindRegistry possibly via an internal util.  All the callers that use it are inside org.eclipse.jdt.ui.unittest.junit.

I don't think ITestKind is meant to be a public interface nor should it be forced to be due to another plug-in's mistake.  Another solution to this is to mark the getJUnitTestKind() method as deprecated and eventually have the current callers use the alternate method of obtaining it.  It would require a major version bump to remove it officially but I don't think that is necessary as no-one else will ever use it and could always be modified to return null to make it unusable.

The API filter should be ok.

fedejeanne:
> No. It is a mess. The original code should not have added the public method in the public enum and instead only returned the String field and let the callers make the call to TestKindRegistry possibly via an internal util. 

That would mean making `TestKindRegistry` public. But I think not even that would have sufficed: I see pieces of code using the following methods of the interface: `getId`, `getLoaderClassName`, `getLoaderPluginId`, `getDisplayName`, `getClasspathEntries`, etc. which means that either the whole interface should have been made public or at least a wrapper that contains all the necessary data for the clients.

> All the callers that use it are inside org.eclipse.jdt.ui.unittest.junit.

PDE is also using it: https://github.com/fedejeanne/eclipse.pde/blob/master/ui/org.eclipse.pde.unittest.junit/src/org/eclipse/pde/unittest/junit/launcher/JUnitPluginLaunchConfigurationDelegate.java#L473C53-L473C67

> The API filter should be ok.

👍 Thank you again for reviewing it!

jjohnstn:
Thanks @fedejeanne - merging.

Issue #837, opened 2023-09-28
[junit] provide a pure JUnit5 classpath container


laeubi:
Currently JDT provides a JUNIT5 classpath container that uses the [junit-vintage](https://junit.org/junit5/docs/current/user-guide/#migrating-from-junit4) engine allowing to run Junit5 in parallel to JUnit 3/4:

![grafik](https://github.com/eclipse-jdt/eclipse.jdt.ui/assets/1331477/f38475ef-6cde-4402-b635-0fbb036c2da6)

for project that want to go pure JUnit5 it would be good to strip away the vintage stuff and only have the new JUnit5 items so for example the don't get suggestions from JUnit3/4 API anymore.

To make this non intrusive I would suggest to allow adding an attribute to the container like this:

```
<classpathentry kind="con" path="org.eclipse.jdt.junit.JUNIT_CONTAINER/5">
  <attributes>
	  <attribute name="vintage" value="false"/>
  </attributes>
</classpathentry>
```

If no attribute is given it should assume `true` so existing setups are not affected.

laeubi:
Tycho now has also support for this:
- https://github.com/eclipse-tycho/tycho/pull/3896

PR #886, opened 2023-10-30
Add support for disabling the vintage engine with an attribute

laeubi:
This adds the infrastructure for disabling the adding of vintage engine/junit4 to the classpath of the JUnit5 container.

See 
- https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/837

This does not add the UI part yet as I first want to get generals agreement on this before I start with further improvements.

@iloveeclipse @akurtakov can you review and/or suggest a reviewer?
FYI @HeikoKlare

### DIFF
--- a/org.eclipse.jdt.junit.core/META-INF/MANIFEST.MF
+++ b/org.eclipse.jdt.junit.core/META-INF/MANIFEST.MF
@@ -3,7 +3,7 @@ Automatic-Module-Name: org.eclipse.jdt.junit.core
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.jdt.junit.core;singleton:=true
-Bundle-Version: 3.12.200.qualifier
+Bundle-Version: 3.13.0.qualifier
 Bundle-Activator: org.eclipse.jdt.internal.junit.JUnitCorePlugin
 Bundle-ActivationPolicy: lazy
 Bundle-Vendor: %providerName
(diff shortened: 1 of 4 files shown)


akurtakov:
@jjohnstn Would you please look into this one?

jjohnstn (commented):
Looks good, but I think I would have made the user specify true to activate the behavior and have it false by default.  So, instead of specifying vintage false, I would have renamed it to be something like pureJunit5 or noVintage or omitVintage set to true.  Feel free to disagree if you feel strongly it should be a false setting.

laeubi:
> Looks good, but I think I would have made the user specify true to activate the behavior and have it false by default.

That would have been optimal, but main problem is to retain backward compatibility for existing setups, but as the user obviously should not fiddle around at all with this setting in the classpath directly (but having a checkbox on the JUnit page) I think this is acceptable here.

> something like pureJunit5 or noVintage or omitVintage

I think these names sound a bit odd and as explained ideally "no vintage" would have been the default in the first place, but as we can't change that the little unusual of a default = true seems acceptable, so I would prefer use the suggested approach here.

I just need to know if this can be merged as is and I add the UI parts as a separate change or do you prefer to have it in one PR?

jjohnstn:
@laeubi Adding the UI parts as a separate change is fine.

laeubi:
> @laeubi Adding the UI parts as a separate change is fine.

@jjohnstn alright, can you merge the change then?

Issue #897, opened 2023-11-03
CPListElement does not take existing attributes into account


laeubi:
`LibrariesWorkbookPage` calls `org.eclipse.jdt.internal.ui.wizards.buildpaths.CPListElement.CPListElement(Object, IJavaProject, IClasspathEntry, int, IPath, IModuleDescription, boolean, IResource, IPath)` but in the case of an edited classpath entry every custom attribute is not updated because `CPListElement` ignores the `IClasspathEntry` supplied extra attributes.

PR #898, opened 2023-11-03
Add UI for user to select if vintage engine should be used

laeubi:
Currently one needs to edit the classpath directly what is a bit inconvenient.

This adds a new checkbox displayed in case JUNIT5 is used that allows the user to choose if vintage engine should be used.

Fix https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/837

@jjohnstn this is the UI part, can you review/merge?

There is currently one bug preventing one from edit the setting afterwards I have created
- https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/897

to track that issue.


laeubi:
Any chance this can get in before the freeze on monday?

org.eclipse.jdt.junit/src/org/eclipse/jdt/internal/junit/ui/JUnitMessages.properties
```diff
@@ -190,6 +190,7 @@ JUnitContainerWizardPage_wizard_title=JUnit Library
 JUnitContainerWizardPage_option_junit3=JUnit 3
 JUnitContainerWizardPage_option_junit4=JUnit 4
 JUnitContainerWizardPage_option_junit5=JUnit 5
+JUnitContainerWizardPage_enableVintage=Enable Vintage Engine
```
jukzi, line 193: For a user (like me) it's not selfexplaining what "Vintage Engine" is all about. Can you please add a some explanation or link to https://junit.org/junit5/docs/current/user-guide/#migrating-from-junit4 ? 
Is it same as"JUnit 3 + JUnit 4"? That would be understandable.

laeubi: > For a user (like me) it's not selfexplaining what "Vintage Engine" is all about

I don't think there is enough space/context to explain JUnit5 concepts to users in this dialog

> Is it same as"JUnit 3 + JUnit 4"? That would be understandable.

Vintage is JUnit 5 + Support for old JUnit 3 + JUnit 4

laeubi: grafik

I made now a clickable link to the documentation and added a tooltip, I think there is not much more we can do.

org.eclipse.jdt.junit/src/org/eclipse/jdt/internal/junit/ui/JUnitMessages.properties
```diff
@@ -190,6 +190,8 @@ JUnitContainerWizardPage_wizard_title=JUnit Library
 JUnitContainerWizardPage_option_junit3=JUnit 3
 JUnitContainerWizardPage_option_junit4=JUnit 4
 JUnitContainerWizardPage_option_junit5=JUnit 5
+JUnitContainerWizardPage_enableVintage=Enable <a>Vintage Engine</a>
+JUnitContainerWizardPage_enableVintage_tooltip=JUnit Vintage provides a TestEngine for running JUnit 3 and JUnit 4 based tests on the platform.
```
iloveeclipse, line 194: > on the platform

Can we change it to "on JUnit 5"?

With "platform" everyone understands something different...

laeubi: Changed it to "JUnit Platform", because [JUnit 5 = JUnit 5 = JUnit Platform + JUnit Jupiter (+ JUnit Vintage)](https://junit.org/junit5/docs/current/user-guide/#overview-what-is-junit-5), claiming it runns Junit 3/4 on Junit 5 is confusing it think as it is not the case...

iloveeclipse: But that's the case, we are running JUnit 3/4 on some libraries provided by JUnit 5. Anyway, I personally don't care, just it will confuse anyone else.

jukzi:
i tested the PR:
when i switch the junit version the box vanishes but the text stay
image
i'l be AFKfor today.

org.eclipse.jdt.junit/src/org/eclipse/jdt/internal/junit/ui/JUnitMessages.properties
laeubi: Well technically there is `JUnit Platform` that defines generic test runs, then there is

- `JUnit Jupiter` as an Provider of a `JUnit Platform` (often refereed to as JUnit5) with new annotations and extension model
- `JUnit Vintage` what Implements a `JUnit Platform` to run Junit4/3 programming model

So basically yes we probabbly better have had not "JUNIT5" but `JUnit Platform` and then one would have ben able to dynamically select ANY `JUnit Platform` provider dynamical but thats a bit of history now and not sure if anyone is willing to change this just for the naming...

That's also why I said I'm not sure if such a configuration page is the right place to tell people JUnit concepts... I would assume that a user made the choice before open the wizard at all...

laeubi:
> when i switch the junit version the box vanishes but the text stay

Fixed.

jjohnstn (commented):
The New JUnit test case wizard still presents the user the option to create a Junit 3 or Junit 4 test which can't be run without the vintage engine support.  The wizard needs to be changed to either not offer JUnit 3 or 4 in such a scenario or to make it an error.  Likewise, the test finders would need to not even offer to run a JUnit 3 or 4 test style test case as a JUnit test in such a project.

I agree with others that we shouldn't assume the end-user knows about the vintage engine.  The option could be reworded to be something like: "Disable JUnit 3 and 4 support" or "Support JUnit 3 and 4" depending on which way you want to present it.

laeubi:
> The New JUnit test case wizard still presents the user the option to create a Junit 3 or Junit 4 test which can't be run without the vintage engine support. 

Yes that's another place that needs adjustments, but I would like to make this on a separate PR to not let changes grow to large with many aspects.

> I agree with others that we shouldn't assume the end-user knows about the vintage engine.

I'm an enduser and I know about it ;-)

If you search for junit5 vintage you find for example:
- https://www.baeldung.com/java-junit-vintage-engine-vs-junit-jupiter-engine
- https://stackoverflow.com/questions/59193282/difference-between-junit-vintage-engine-and-junit-jupiter-engine
- and many more elaborating the topic ...

because that a common term for example also the maven artifacts are named as such, so I think it is good enough for users to understand and easy enough for anyone interested to find out about it if not.

jukzi (changes requested):
term "Enable <a>Vintage Engine</a>" unclear

laeubi:
> term "Enable Vintage Engine" unclear

I already added an explanation see above screenshot ....

jukzi:
> I already added an explanation see above screenshot ....

see https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/898#pullrequestreview-1716437001 that that was not rated sufficent. please update wording, resolve conflict and ping me for review.

laeubi:
I'm sure JDT team will come with a much cooler solution than I can offer then.

PR #909, opened 2023-11-08
Remove empty if-statement

fedejeanne:
## What it does
Remove empty `if`-statement. It has no side-effects.


fedejeanne:
There's been no activity on this one. Could someone please review it?

org.eclipse.jdt.core.manipulation/refactoring/org/eclipse/jdt/internal/corext/refactoring/util/ChangedValueChecker.java
```diff
@@ -369,9 +369,7 @@ private void extend2EndOfLoop(ASTNode node) {
 						|| temp instanceof ForStatement || temp instanceof DoStatement) {
 					int offset= temp.getStartPosition();
 					int length= temp.getLength();
-					if (offset < startOffset && offset + length > startOffset)
```
jjohnstn, line 372: Let's say startOffset is 4 and endOffset is 8.  If there is something that has offset 2 and length 10, it will fall into this if and do nothing.  It will not attempt the else which would also trigger as 2 is < 8 and 12 is > 8 but the else changes the endOffset.  So, the if is not an obvious empty statement to remove.   It could be made clearer by incorporating the logic of the if into the else if expresion to form a new if where the endOffset change only occurs if all conditions are checked.

jjohnstn (changes requested):

org.eclipse.jdt.core.manipulation/refactoring/org/eclipse/jdt/internal/corext/refactoring/util/ChangedValueChecker.java
fedejeanne: Thank you for reviewing it!
Good point, I totally missed the `else` in the following line. I extracted some variables for the conditions and added some comments in an attempt to make the logic a bit clearer. I'm not completely happy with some of the names though, I couldn't come up with anything better than `cond1` and `cond2`. Do you have any suggestion on that? Or do you think the comments are enough to clarify the logic?

jjohnstn (approved):

PR #913, opened 2023-11-09
Disable some problem marker

jukzi:
`The expression of type _ is boxed into _` - not a problem 
`Empty block should be documented` - not a problem 
`The allocated object is never used` - normal for SWT 
`The switch statement should have a default case`
`_ is a raw type. References to generic type _<E> should be parameterized` - new code is reviewed anyway and nobody takes time to refactor the old raw types

Also includes some minor warning fixes


jukzi:
With this change JDT UI shows only more relevant problem markers, like Potential NPE, resource leaks, ... which may make sense to actual fix and otherwise just get lost in shuffle

PR #921, opened 2023-11-13
Small checks and behavior differences for refactorings and assists

robstryker:
Part 1

There will be 2 test failures but i forget what they are so I'll let Jenkins discover it and fix them. 

## Author checklist


org.eclipse.jdt.core.manipulation/refactoring/org/eclipse/jdt/internal/corext/refactoring/surround/SurroundWithTryCatchRefactoring.java
```diff
@@ -161,7 +161,7 @@ public RefactoringStatus checkActivationBasics(CompilationUnit rootNode) throws
 		RefactoringStatus result= new RefactoringStatus();
 		fRootNode= rootNode;
 
-		fAnalyzer= new SurroundWithTryCatchAnalyzer(fCUnit, fSelection);
+		fAnalyzer= new SurroundWithTryCatchAnalyzer(fCUnit, fSelection, true);
```
jjohnstn, line 164: I think this needs to be a preference which jdt-ls can set to true by default and false otherwise.  I think there are scenarios where a user would want to handle the exception for the method (e.g. to log something that isn't ever supposed to happen and a similar-type exception is actually expected in subsequent code).  Removing the case forces the user have to add it manually vs deleting it if it is extraneous.

jjohnstn (dismissed):

robstryker:
Reverted the test changes and behavior. Since this is delayed, I will keep making any changes necessary for jdt.ls. Marking as draft.

rgrunber (commented):
Change looks fine to me. This is mainly adding some protected classes allowing for some possible customization in refactoring behaviour.

My only concern was about the code where you retrieve only the first element of a composite change. However, looking into the (2) implementations of `RefactoringCorrectionProposalCore`, none appear to be returning a composite change. Furthermore looking at most classes that implement `Refactoring`, they seem to be returning a `CompilationUnitChange`, which is a `TextChange`.

PR #927, opened 2023-11-17
Cache the list of names whose favorite static imports we have resolved.

rgrunber:
Originally from https://github.com/redhat-developer/vscode-java/issues/3383

- Do not recompute favorite static imports for an unknown type
- Reuse compilation unit working copy in code completion logic

I tested at runtime against https://github.com/eclipse/lemminx/blob/main/org.eclipse.lemminx/src/test/java/org/eclipse/lemminx/XMLAssert.java by removing all imports and triggering the organize imports functionality. Any easy reproducer is just create some source file with **a lot** of static imports. Remove them all, and call "Organize Imports".

`ImportOrganizeTest*` were useful for testing this. If I have some time, I can investigate adding something to `OrganizeImportsPerfTest`. **Update:** The existing performance tests there don't seem to have any useful assertions as far as I can tell.

`getFavoriteStaticImports(..)` went from ~3s to ~300ms .

**Before**
organize-static-imports-before

**After**
organize-static-imports-after

### DIFF
--- a/org.eclipse.jdt.core.manipulation/common/org/eclipse/jdt/core/manipulation/OrganizeImportsOperation.java
+++ b/org.eclipse.jdt.core.manipulation/common/org/eclipse/jdt/core/manipulation/OrganizeImportsOperation.java
@@ -74,6 +74,7 @@ import org.eclipse.jdt.internal.core.manipulation.util.Strings;
 import org.eclipse.jdt.internal.corext.dom.ASTNodes;
 import org.eclipse.jdt.internal.corext.dom.Bindings;
 import org.eclipse.jdt.internal.corext.dom.ScopeAnalyzer;
+import org.eclipse.jdt.internal.corext.util.StaticImportFavoritesCompletionInvoker;
 import org.eclipse.jdt.internal.corext.util.JavaModelUtil;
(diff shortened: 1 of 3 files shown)


org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/util/CodeCompletionExecutor.java
```diff
@@ -0,0 +1,98 @@
...
+import org.eclipse.jdt.core.IType;
+import org.eclipse.jdt.core.JavaModelException;
+import org.eclipse.jdt.core.Signature;
+
+/**
+ * Copied from {@link JavaModelUtil#getStaticImportFavorites(ICompilationUnit, String, boolean, String[])}
+ * If code completion needs to be called multiple times on a given compilation unit,
+ * this class will perform the operations more efficiently by re-using a working copy.
+ */
+public class CodeCompletionExecutor {
```
rgrunber, line 32: Maybe I could call this something better. CodeCompletionInvoker ? Also, given that this is copied from `JavaModelUtil` I should keep the original license header. Although various code blocks were relocated.

PR #985, opened 2023-12-06
Make JavaUI view constants final

akurtakov:
It is a plain oversight as I can not imagine anyone wanting these to be overridden.
Added api_filters to not cause extra warnings in the code.

### DIFF
--- a/org.eclipse.jdt.ui/.settings/.api_filters
+++ b/org.eclipse.jdt.ui/.settings/.api_filters
@@ -137,6 +137,38 @@
             </message_arguments>
         </filter>
     </resource>
+    <resource path="ui/org/eclipse/jdt/ui/JavaUI.java" type="org.eclipse.jdt.ui.JavaUI">
+        <filter comment="View ID constants should be final." id="388100214">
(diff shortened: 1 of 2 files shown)


iloveeclipse:
> I can not imagine anyone wanting these to be overridden.

One can never know if some ancient IBM product relies on that "feature" :-)
But look at the commit 22e3f6a51ee8b46dca97c14668411634708b5f93...

akurtakov:
So do you ask me to abandon this one?
I see it plain wrong to allow overriding globally the ids of JDT views and only thing it could break is someone overriding the view ids. The change is even "compatible" as per https://wiki.eclipse.org/Evolving_Java-based_APIs_2 which describes:
```
Change non-final to final - Breaks compatibility (6)
(6) Pre-existing binaries that subclass the API class will fail with a link-time error. 
```
but JavaUI class is final so there is no chance to break compatibility while removes a case which should have never been there.

iloveeclipse:
> So do you ask me to abandon this one?

No. Just wanted to show where the surprising state is coming from.

PR #994, opened 2023-12-08
Jenkinsfile: fail when number of problems increased

jukzi:
see https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/941#discussion_r1420054085

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -27,7 +27,7 @@ pipeline {
 					archiveArtifacts artifacts: '*.log,*/target/work/data/.metadata/*.log,*/tests/target/work/data/.metadata/*.log,apiAnalyzer-workspace/.metadata/*.log', allowEmptyArchive: true
 					junit '**/target/surefire-reports/*.xml'
 					discoverGitReferenceBuild referenceJob: 'eclipse.jdt.ui-github/master'
-					recordIssues publishAllIssues: true, tools: [java(), mavenConsole(), javaDoc()]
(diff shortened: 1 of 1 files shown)


jukzi:
seems like it does not work since no java compiler warnings are found:
image
but there should be plenty
how to enable compiler warnings? @akurtakov

laeubi:
> seems like it does not work since no java compiler warnings are found:
> but there should be plenty

The jenkins file has `-Dmaven.compiler.failOnWarning=true` that means this build would even **fail** for any (reported) compiler warning, it of cause do not (and can of course not report any) if they are disabled by any mean for example:

https://github.com/eclipse-jdt/eclipse.jdt.ui/blob/e8754eadf7c8e6dd48f7d02b517e12363bba9329/org.eclipse.jdt.ui/pom.xml#L24

jukzi:
What is the syntax in the pom.xml to only ignore `deprecation` and `unavoidableGenericProblems`, that seems to be the intention but they disabled all warnings?

akurtakov:
IMHO, pom.xml shouldn't disable warnings at all as in that case there is no way to catch new warnings.

laeubi:
I think there was one point in time where one wants to use `-Dmaven.compiler.failOnWarning=true` but not fix some warnings ...

jukzi:
own branch: https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/995

PR #995, opened 2023-12-08
Jenkinsfile: fail when number of problems increased

jukzi:


jukzi:
@laeubi 
even with this `<code.ignoredWarnings>${tests.ignoredWarnings}</code.ignoredWarnings>` no compilers are reported what else could be the problem?:
image

akurtakov:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blame/b5384a6f123641c97ca22210940b0021097a98f1/eclipse-platform-parent/pom.xml#L114 defines it to pretty much ignore everything .

akurtakov:
The years of ignoring warnings on one level or another starts to bite us on every step.

jukzi:
@Bananeweizen do you know how to configure it so that the warnings are shown in jenkins?

laeubi:
> It's annoying that the compiler warnings are not parsed well. jenkins or whatever plugin seems to expect to find the warning in the same line. example:
> 
> ```
> 12:19:01  [WARNING] /home/jenkins/agent/workspace/eclipse.jdt.ui-github_PR-995/org.eclipse.jdt.ui/core extension/org/eclipse/jdt/internal/corext/buildpath/IBuildpathModifierListener.java:[22] 
> 12:19:01  	void buildpathChanged(BuildpathDelta delta);
> 12:19:01  	     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> 12:19:01  Javadoc: Missing comment for public declaration
> ```
> 
> image
> 
> `javac ` produces warnings in another orde. like:
> 
> ```
> .\org\eclipse\jdt\core\compiler\IProblem.java:651: warning: [dep-ann] deprecated item is not annotated with @Deprecated
>         int UnnecessaryArgumentCast = Internal + TypeRelated + 182;
>             ^
> ```

The `java()` parser is used here the documentation indicates there is also an `eclipse()` parser:
https://github.com/jenkinsci/warnings-ng-plugin/blob/main/SUPPORTED-FORMATS.md

Bananeweizen:
I tried the eclipse() parser setting already with the issue that you had opened somewhere else some weeks ago. Not sure anymore where to find it. The warnings plugin parser works fine for me at work etc., that's why I suspect something like unexpected line endings or similar.

Issue #1008, opened 2023-12-13
Quickfix Converting to Enhanced For Loop Creates Invalid Code


amartya4256:
## Current Behavior

JDT provides a quickfix for converting classic `for` loops into enhanced for loops. When using a classic `for` loop to iterate through the elements of an iterator calling `next()` within the loop body but not using its result (such as when only counting the elements), applying the quickfix for converting into an enhanced `for` loop leads to syntactically incorrect code.

### Example

Original code:
```java
List<Object> list = new ArrayList<>();
int i = 0;
for (Iterator<Object> iterator = list.iterator(); iterator.hasNext();) {
	iterator.next();
	i++;
}
```

After applying the quickfix: 
```java
List<Object> list = new ArrayList<>();
int i = 0;
for (Object object : list) {
	object;
	i++;
}
```
`object;` is an invalid statement.

### Reproduction Environment

Happens with Eclipse 2023-09 and JDK 17.

## Expected Behavior

The quick fix should always produce syntactically (and semantically) correct code. For the given example, the quick fix should either not be provided or the result should look as follows:
```java
List<Object> list = new ArrayList<>();
int i = 0;
for (Object object : list) {
	i++;
}
```

## Additional Information

The invalid result only occurs when performing the `next()` call within the loop body. When performing it within the loop header, the refactoring result is correct, e.g.:

Original code:
```java
List<Object> list = new ArrayList<>();
int i = 0;
for (Iterator<Object> iterator = list.iterator(); iterator.hasNext(); iterator.next()) {
	i++;
}
```

After applying the quickfix: 
```java
List<Object> list = new ArrayList<>();
int i = 0;
for (Object object : list) {
	i++;
}
```

PR #1009, opened 2023-12-13
Bugfix for Quickfix Converting to Enhanced For Loop Creates Invalid Code #1008

amartya4256:
Fixes https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1008

## What it does
The change fixes the conversion of an iterator based for loop to an enhanced for loop. When there is an iterator.next() in the body of the for loop and is not assigned to any variable, it becomes variable; on conversion which is an errorful statement in Java.

## How to test
Create a for loop with iterator having iterator.next() in the body in any way possible and convert the loop to enhanced for loop from the quick fixes menu.

## Details

This PR uses the block of code used by fOccurences for replacing them with comments if available. We extract the block of the code in the fix to a new method so that it can be used by ASTVisitors as well for the statements which dont have any assignments but consists of iterator calls. The extracted block throws JavaModelException. So, to deal with it I have used a try catch ball where we are extracting the comment Strings out. Just to make sure if there happens an exception while adding the comments during the conversion we dont just stop the whole operation but rather set the comments to an empty string. The other way could have been to not use try/catch but rather add throws in the method signature of the visit method of the ASTVisitor which is a no go since it is an overridden method.

I have also included some tests which support my changes for the fix.

## Author checklist

- [x ] I have thoroughly tested my changes
- [x ] I have signed the [Eclipse Contributor Agreement (ECA)](https://www.eclipse.org/legal/ECA.php)

### DIFF
--- a/org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/fix/ConvertIterableLoopOperation.java
+++ b/org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/fix/ConvertIterableLoopOperation.java
@@ -18,6 +18,7 @@ import java.util.ArrayList;
 import java.util.Enumeration;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Optional;
 
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IStatus;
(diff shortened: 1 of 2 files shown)


jjohnstn (approved):
Looks good.  Thanks.

Issue #1013, opened 2023-12-14
Quick Fix: Convert Iterable to Enhanced For Loop Omits Next Method for Any Class


amartya4256:
There's a bug currently on converting an iterable to enhanced for loop. When there is a statement like object.next(); The statement is omitted upon conversion even if the object belongs to a different class. 

The problem is that in the block there's is no check for the iterator variable before performing the operation and it must be added.

```java
public void foo() throws IOException {
	Test i = new Test();
	for (Iterator<String> iterator = c.iterator(); iterator.hasNext();) {
		i.next();
                iterator.next();
	}
}
```
```java
class Test {
	public Object next() {
		return null;
	}
}
```

Upon Conversion:

### Current Behavior
```java
public void foo() {
    Test i = new Test();
    for (String string: c) {
    }
}
```

### Expected Behavior
```java
public void foo() {
    Test i = new Test();
    for (String string: c) {
        i.next(); 
    }
}
```

PR #1014, opened 2023-12-14
Quick Fix: Convert Iterable to Enhanced For Loop Omits Next Method fo…

amartya4256:
## What it does

The fix to check if the expression containing next as a method binding is the relevant iterator.

Fixes https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1013

## How to test
The problem is that in the block there's is no check for the iterator variable before performing the operation and it must be added.

```java
public void foo() throws IOException {
	Test i = new Test();
	for (Iterator<String> iterator = c.iterator(); iterator.hasNext();) {
		i.next();
                iterator.next();
	}
}
```
```java
class Test {
	public Object next() {
		return null;
	}
}
```

Upon Conversion:

### Previous Behavior
```java
public void foo() {
    Test i = new Test();
    for (String string: c) {
    }
}
```

### Current/Expected Behavior
```java
public void foo() {
    Test i = new Test();
    for (String string: c) {
        i.next(); 
    }
}
```

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/fix/ConvertIterableLoopOperation.java
+++ b/org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/fix/ConvertIterableLoopOperation.java
@@ -298,19 +298,20 @@ public final class ConvertIterableLoopOperation extends ConvertLoopOperation {
 				public final boolean visit(final MethodInvocation node) {
 					IMethodBinding binding= node.resolveMethodBinding();
 					if (binding != null && ("next".equals(binding.getName()) || "nextElement".equals(binding.getName()))) { //$NON-NLS-1$ //$NON-NLS-2$
-						List<ASTNode> siblingNodes = Optional.ofNullable(ASTNodes.getParent(node, Statement.class))
-								.map(it -> ASTNodes.getChildren(it))
-								.orElse(new ArrayList<>());
-						if(siblingNodes.stream().allMatch(node::equals)) {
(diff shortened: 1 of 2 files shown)


robstryker:
Patch looks straightforward. A 7-line block of code was moved from the beginning of an if-statement into a nested if-statement with more specific conditions, ie, when the Expression is of type `Name` . The block appears otherwise unchanged. A test case has also been added. 

Not being familiar with the code, my fear would be that if this block of code needed to execute for the more general condition, ie if the Expression was of any type other than `Name`, it's possible this could lead to a loss of functionality. 

However, all the tests pass, which means that no previously-considered use case is failing. 

I'd be inclined to merge this if I had commit rights, but maybe @jjohnstn  can review further.

amartya4256:
> Patch looks straightforward. A 7-line block of code was moved from the beginning of an if-statement into a nested if-statement with more specific conditions, ie, when the Expression is of type `Name` . The block appears otherwise unchanged. A test case has also been added.
> 
> Not being familiar with the code, my fear would be that if this block of code needed to execute for the more general condition, ie if the Expression was of any type other than `Name`, it's possible this could lead to a loss of functionality.
> 
> However, all the tests pass, which means that no previously-considered use case is failing.
> 
> I'd be inclined to merge this if I had commit rights, but maybe @jjohnstn can review further.

Yeah this block of code was added by me yesterday in a previous PR. It's nothing but a fix for the regression that happened because of it the last day. Ideally this block must only be executed after checking if the expression of the method binding is the iterator otherwise it will also apply for any next or nextElement method. I didn't see it yesterday but realized that today.

PR #1016, opened 2023-12-14
Touch bundles affected by the changed compilation settings

iloveeclipse:
See https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1661 
See https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1659


iloveeclipse:
No **real** test fails, only quality gate warnings reported as errors

Issue #1021, opened 2023-12-15
Missing stack traces of fails/errors for @TestTemplate with only 1 executed test


trancexpress:
A regression from: https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/946

We don't show the child that has the fail trace. Also double-clicking on the test suite (with no children) doesn't navigate to the failed line.

iloveeclipse:
JUnit 5 example to reproduce:

```
import static org.junit.jupiter.api.Assertions.assertEquals;

import org.junit.Assume;
import org.junit.jupiter.api.MethodOrderer.MethodName;
import org.junit.jupiter.api.RepeatedTest;
import org.junit.jupiter.api.Test;
import org.junit.jupiter.api.TestMethodOrder;

@TestMethodOrder(MethodName.class)
public class TestTemplatesJunit5 {

	@Test
	public void testOkNoRepeats() throws Exception {
		Thread.sleep(500);
	}
	
	@RepeatedTest(value = 1)
	public void testOkOneRepeat() throws Exception {
		Thread.sleep(500);
	}

	@RepeatedTest(value = 2)
	public void testOkTwoRepeats() throws Exception {
		Thread.sleep(500);
	}
	
	@Test
	public void testWithErrorNoRepeats() throws Exception {
		doStupidThings(null);
	}
	
	@RepeatedTest(value = 1)
	public void testErrorOneRepeat() throws Exception {
		doStupidThings(null);
	}
	
	@RepeatedTest(value = 2)
	public void testErrorTwoRepeats() throws Exception {
		doStupidThings(null);
	}

	@Test
	public void testFailNoRepeats() throws Exception {
		assertEquals("expected", "observed");
	}
	
	@RepeatedTest(value = 1)
	public void testFailOneRepeat() throws Exception {
		assertEquals("expected", "observed");
	}

	@RepeatedTest(value = 2)
	public void testFailTwoRepeats() throws Exception {
		assertEquals("expected", "observed");
	}

	@RepeatedTest(value = 0)
	public void testWrongRepeat() throws Exception {
		Thread.sleep(500);
	}
	
	@Test
	public void testAssume() throws Exception {
		Assume.assumeFalse(true);
	}
	
	@RepeatedTest(value = 1)
	public void testAssumeOneRepeat() throws Exception {
		Assume.assumeFalse(true);
	}
	
	@RepeatedTest(value = 2)
[...]

Issue #945, opened 2023-11-28
Dont show children for JUnit 5 @TestTemplate tests that ran only once


trancexpress:
The JUnit view displays `@TestTemplate` JUnit 5 tests with an extra `TestSuiteElement`, that contains the template executions.

E.g. the following test code:

```
public class TestTemplatesJunit5 {

	@RepeatedTest(value = 1)
	public void test1() throws Exception {
		Thread.sleep(500);
	}

	@RepeatedTest(value = 2)
	public void test2() throws Exception {
		Thread.sleep(500);
	}

	@Test
	public void test3() throws Exception {
		Thread.sleep(500);
	}
}
```

Results in this JUnit view state: 

![junit_view_repeated_test_single_run](https://github.com/eclipse-jdt/eclipse.jdt.ui/assets/24752155/966ba0f1-d093-4c08-9f98-32f91915309a)

For `test.TestTemplatesJunit5.test1()` this extra `TestSuiteElement` results in only noise. E.g. when repeating failed tests (see https://github.com/artsok/rerunner-jupiter), often there will be many (or all) tests in a test case that have this structure single child in a test suite.

Ideally we hide such `TestSuiteElement` tree elements with only a single `TestCaseElement` child. E.g. the view could look like this:

![junit_view_repeated_test_improvement](https://github.com/eclipse-jdt/eclipse.jdt.ui/assets/24752155/a0dc3626-5719-4c18-a47c-5d8f6110bd08)

iloveeclipse:
Fix was reverted via https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/1203, reopening.

PR #1022, opened 2023-12-15
Failed TestTemplate with only one test doesn't show fail trace

trancexpress:
This change adjusts TestSuiteElement to show the trace of a failed child, if that child is the only test case it ran and its dynamic. It also adjusts navigation to the failed line in this case - so that navigation goes to the failed line and not to the test method with the @TestTemplate annotation.

Fixes: #1021

## Author checklist


trancexpress:
@iloveeclipse this change makes me think that we want to redo the original fix in some way... Since we are IMO adding two dirty hacks on top of a hack... But I'm not sure how else to fix https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/945 and this is (I think) a marginal corner case (that we happen to run into due to our retry-on-fail test policy). I'm not sure its worth bigger time investment to fix this the right way (assuming there is one)?

org.eclipse.jdt.junit/src/org/eclipse/jdt/internal/junit/ui/TestViewer.java
```diff
@@ -448,10 +448,15 @@ private OpenTestAction getOpenTestAction(TestSuiteElement testSuite) {
 		String testName= testSuite.getTestName();
 		ITestElement[] children= testSuite.getChildren();
 
-		if (testName.startsWith("[") && testName.endsWith("]") && children.length > 0 && children[0] instanceof TestCaseElement) { //$NON-NLS-1$ //$NON-NLS-2$
+		if (children.length > 0 && children[0] instanceof TestCaseElement tce && tce.isDynamicTest()) {
```
trancexpress, line 451: This doesn't take into account e.g. `@RepeatedTest` that makes effort to pretty-print the test executions as e.g. `repeat 1 of 4` and not just `[1]`, `[2]`, etc..

iloveeclipse (approved):

iloveeclipse:
There are no **real** test fails on jenkins, only "Quality gate" maven warnings reported as errors. Hate that.

Issue #1024, opened 2023-12-20
Entries of JAR description files saved while exporting a JAR have non-deterministic order


HeikoKlare:
## Current Behavior

The JAR Exporter allows to export JAR files from a selection of resources in the workspace. The description for such an export can be saved in JAR description files (*.jardesc). From these descriptions, an export can be started:

![Image](https://github.com/vi-eclipse/Eclipse-JDT/assets/755472/753d6150-7649-49fd-abd4-2f42dbf08b5a)

The export allows to save the description file, e.g., at the original location to update it with a modified selection of elements to be exported into the JAR:

![Image](https://github.com/vi-eclipse/Eclipse-JDT/assets/755472/68c4a269-1834-4616-9c43-824076081ee5)

The generated *.jardesc file is an XML file containing the selected elements, e.g.:
```xml
<selectedElements exportClassFiles="true" exportJavaFiles="false" exportOutputFolder="false">
    <javaElement handleIdentifier="=JarDescriptionDemo/test=/test=/true=/"/>
    <javaElement handleIdentifier="=JarDescriptionDemo/src"/>
</selectedElements>
```

These elements are in arbitrary order, which can change upon every export.

## Expected Behavior

In order to ease merging of these *.jardesc files, the order of the selected elements should be deterministic, e.g., by simply sorting them by their name.

## Additional Information

The `org.eclipse.jdt.internal.ui.jarpackager.JarPackageWizard` sets the elements to be exported via `JarPackageData` from 
`org.eclipse.jdt.internal.ui.jarpackager.JarPackageWizardPage#getSelectedElementsWithoutContainedChildren()`. This method converts a (hash) set of elements to an array, leading to different orders based on the element's hash values: 

https://github.com/eclipse-jdt/eclipse.jdt.ui/blob/5c7d1a2b2690f9b7b34a7cf3a724c541c5c6c79d/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/jarpackager/JarPackageWizardPage.java#L663-L667

Sorting the array of returned elements would solve the issue. I will provide a PR for that.

PR #1025, opened 2023-12-20
Make order of elements in saved JAR description file deterministic #1024

HeikoKlare:
## What it does
When exporting JAR description files from the JarPackageWizard, the elements to be exported into the JAR file are currently stored in a non-deterministic order. This makes it particularly difficult to merge the description files in case they are put under version control. For a detailed description, see #1024.

This change makes a JAR description file exported from the JarPackageWizard always list the exported elements in the same order by sorting them according to their names. It also adds an according regression test.

Fixes https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1024

## How to test
The behavior can be tested by creating and updating a *.jardesc file from the JAR export wizard. Before this change, the order of elements in the *.jardesc file will vary upon each export, while after this change the order will be deterministic.

The added test case `WizardTests#testJarPackageWizard_sameElementOrderInJarDescriptionExports` simulates these multiple exports and can also be used for verification.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.ui.tests/ui/org/eclipse/jdt/ui/tests/dialogs/WizardsTest.java
+++ b/org.eclipse.jdt.ui.tests/ui/org/eclipse/jdt/ui/tests/dialogs/WizardsTest.java
@@ -14,10 +14,25 @@
  *******************************************************************************/
 package org.eclipse.jdt.ui.tests.dialogs;
 
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
+
+import java.io.InputStream;
+import java.nio.file.Files;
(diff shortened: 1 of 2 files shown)


iloveeclipse (approved):

iloveeclipse:
/rebase 

Note: currently the PR fails because of stupid quality gate => https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1026
However if rebased, it will fail due https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1028...

Still rebasing to see if we only get one test fail.

HeikoKlare:
Thanks for processing this PR so quickly, @iloveeclipse!

PR #1048, opened 2024-01-06
refactor method-ref fixes to be used in headless mode.

gayanper:
## What it does
Refactor AddMissingMethodReferenceDeclaration logic to be ready to consume by both Eclipse UI and headless consumers such as jdt.ls

## How to test
Existing test suite

## Author checklist

### DIFF
--- a/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/text/correction/QuickAssistProcessor.java
+++ b/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/text/correction/QuickAssistProcessor.java
@@ -155,7 +155,6 @@ import org.eclipse.jdt.internal.corext.fix.AddVarLambdaParameterTypesFixCore;
 import org.eclipse.jdt.internal.corext.fix.ChangeLambdaBodyToBlockFixCore;
 import org.eclipse.jdt.internal.corext.fix.ChangeLambdaBodyToExpressionFixCore;
 import org.eclipse.jdt.internal.corext.fix.CleanUpConstants;
-import org.eclipse.jdt.internal.corext.fix.CompilationUnitRewriteOperationsFixCore;
 import org.eclipse.jdt.internal.corext.fix.ControlStatementsFix;
 import org.eclipse.jdt.internal.corext.fix.ConvertLambdaToMethodReferenceFixCore;
 import org.eclipse.jdt.internal.corext.fix.ConvertLoopFixCore;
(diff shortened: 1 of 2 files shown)


org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/text/correction/QuickAssistProcessor.java
```diff
@@ -981,252 +979,19 @@ public static boolean getRemoveVarOrInferredLambdaParameterTypes(IInvocationCont
...
-		if (QuickAssistProcessorUtil.isTypeReferenceToInstanceMethod(methodReferenceNode)) {
-			String methodReferenceQualifiedName= ((Name) methodReferenceNode.getExpression()).getFullyQualifiedName();
-			String typeDeclarationName= astRoot.getPackage().getName().getFullyQualifiedName() + '.' + typeDeclaration.getName().getFullyQualifiedName();
-			if (!methodReferenceQualifiedName.equals(typeDeclarationName)
-					&& !methodReferenceQualifiedName.equals(typeDeclaration.getName().getFullyQualifiedName())) {
-				// only propose for references in same class
-				return false;
-			}
-			addStaticModifier= true;
+		if (resultingCollections == null) {
```
jjohnstn, line 983: You should do some form of verification before returning true.  A number of quick assists actually do the work and check if the fix is null or not in concert with the the resultingCollections check for null.  If you don't want to create the fix, you should at least add a method that will do some form of pre-verification.  For example, the original code starts by setting the methodReferenceNode field and returning false if it is null.  There are other returns of false below that that could be done quickly.

jjohnstn (changes requested):
Hi @gayanper Just one comment regarding verification for the hasAssists() call.

org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/text/correction/QuickAssistProcessor.java
gayanper: Agree, i think i found similar cases while looking at some previously changed quick fixes as well. But my bad I didn't consider here when i did this change. 

I think i could extract code in the CoreFix class and use that here and also in createMethod. Will do that changes tomorrow and push a fix.

jjohnstn: @gayanper Any progress?  I can merge once the mentioned code change is done.

gayanper: @jjohnstn i was busy last few days. I will try to fix this today or tomorrow.

org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/text/correction/QuickAssistProcessor.java
```diff
@@ -981,252 +978,19 @@ public static boolean getRemoveVarOrInferredLambdaParameterTypes(IInvocationCont
...
-				for (int i= 0; i < typeArguments.length; i++) {
-					ITypeBinding iTypeBinding= typeArguments[i];
-					SingleVariableDeclaration newSingleVariableDeclaration= ast.newSingleVariableDeclaration();
-					newSingleVariableDeclaration.setName(ast.newSimpleName(iTypeBinding.getErasure().getName().toLowerCase() + (i + 1)));
-					newSingleVariableDeclaration.setType(importRewrite.addImport(iTypeBinding.getErasure(), ast));
-					newMethodDeclaration.parameters().add(newSingleVariableDeclaration);
-				}
+		AddMissingMethodDeclarationFixCore fix= AddMissingMethodDeclarationFixCore.createAddMissingMethodDeclaration(context.getASTRoot(), covering);
+		if (fix != null) {
+			if (resultingCollections == null) {
```
gayanper, line 984: @jjohnstn looking at the code `createAddMissingMethodDeclaration` I feel that the overhead of creating the instance `AddMissingMethodDeclarationFixCore` is less, therefore I used the instance nullability to check if the fix is valid or not. Let me know if you think otherwise

jjohnstn: That is fine.  Thanks.

jjohnstn (approved):

Issue #1045, opened 2024-01-05
"Make Static" refactoring failing because of overlapping text edits


HeikoKlare:
The recently introduced "Make Static" refactoring (#689) fails to be applied because of "overlapping text edits" in some cases.

This is a finding that has been made during verification of the original refactoring contribution: https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/689#issuecomment-1866869206

### Expected Behavior

The refactoring should not fail because of producing overlapping text edits.

### How to Reproduce

The failure can, for example, be reproduced when applying the refactoring to `org.eclipse.jdt.internal.codeassist.CompletionEngine#checkCancel()`

![image](https://github.com/eclipse-jdt/eclipse.jdt.ui/assets/755472/49cb952a-ef99-469f-bec3-f57495e739ce)

PR #1070, opened 2024-01-15
Avoid "overlapping text edits" in "Make Static" refactoring #1045

HeikoKlare:
## What it does
The "Make Static" refactoring currently produces overlapping text edits in case there are changes in the source code file containing the refactored method before and after that method. The reason is that the calls to the refactored method are wrapped into a single MultiTextEdit whose scope is then overlapping with the TextEdit for the refactored method itself.

With this fix, the changes to callers of the refactored method are treated as individual TextEdits rather than one MultiTextEdit, so that none of the generated text edits have pairwise overlaps. An according regression test for the behavior is added.

Fixes https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1045

## How to test
Take any class, in which you want to refactor an instance method to a static method which has callers in that class above and below the method to be refactored. Before this change, the refactoring will fail with an "overlapping text edit". Afterwards it will succeed.
You can, for example, use the method `org.eclipse.jdt.internal.codeassist.CompletionEngine#checkCancel()`, as described in #1045, or the scenario of the regression test added with this PR.

## Author checklist


jjohnstn:
I messed this up by editing the conflicts using the web editor and it did a merge commit.  I will fix and do a force push.

Issue #1044, opened 2024-01-05
"Make Static" refactoring adds unnecessary parameter


HeikoKlare:
The recently introduced "Make Static" refactoring (#689) adds a parameter for the `this` object in some cases in which it is not unnecessary because no `this` reference is used within the method body.

This is a finding that has been made during verification of the original refactoring contribution: https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/689#issuecomment-1866869206

### Expected Behavior

The refactoring should only introduce a parameter for the original `this` object in case `this` is actually accessed in the original method body.

### How to Reproduce

The failure can, for example, be reproduced when applying the refactoring to `org.eclipse.jdt.core.tests.compiler.regression.AbstractRegressionTest#getCompilerTestsPluginDirectoryPath()`.
The refactoring introduces an unnecessary `AbstractRegressionTest` parameter.

#### Before:
```java
protected String getCompilerTestsPluginDirectoryPath() {
	try {
		URL platformURL = Platform.getBundle("org.eclipse.jdt.core.tests.compiler").getEntry("/");
		return new File(FileLocator.toFileURL(platformURL).getFile()).getAbsolutePath();
	} catch (IOException e) {
		e.printStackTrace();
	}
	return null;
}
```

#### After:
```java
protected static String getCompilerTestsPluginDirectoryPath(AbstractRegressionTest abstractRegressionTest) {
	try {
		URL platformURL = Platform.getBundle("org.eclipse.jdt.core.tests.compiler").getEntry("/");
		return new File(FileLocator.toFileURL(platformURL).getFile()).getAbsolutePath();
	} catch (IOException e) {
		e.printStackTrace();
	}
	return null;
}
```

PR #1072, opened 2024-01-15
Ensure "Make Static" refactoring only adds parameter if necessary #1044

HeikoKlare:
## What it does

The "Make Static" refactoring currently adds a "this" parameter to the method in cases in which it is actually not necessary. This is particularly the case the body of the method to be refactored contains invocations of methods on a newly instantiated object. The according logic erroneously considers such a method invocation as a reason to add a parameter.

With this fix, the identification of whether a parameter needs to be added to the method depends on whether the logic calculating an actual change to the method body is actually applied. An according regression test for the behavior is added.

Fixes https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1044

## How to test

It is easiest to test with either the example provided as a regression test within this PR or by applying the "Make Static" refactoring to `org.eclipse.jdt.core.tests.compiler.regression.AbstractRegressionTest#getCompilerTestsPluginDirectoryPath()`, as documented in the according issue #1044.

## Author checklist


jjohnstn (approved):
Looks good.

PR #1077, opened 2024-01-16
Move additional methods in SuppressWarningsSubprocessorCore to jdt.core.manipulation

hopehadfield:
## Author checklist


org.eclipse.jdt.core.manipulation/proposals/org/eclipse/jdt/internal/ui/text/correction/SuppressWarningsSubProcessorCore.java
```diff
@@ -38,4 +64,133 @@ public static final boolean hasSuppressWarningsProposal(IJavaProject javaProject
 		return false;
 	}
 
+	public static void addSuppressWarningsProposals(IInvocationContextCore context, IProblemLocationCore problem, Collection<ChangeCorrectionProposalCore> proposals) {
```
rgrunber, line 67: It looks to me like you've taken https://github.com/eclipse-jdt/eclipse.jdt.ui/blob/8e28ae0210672ac240683b778206bed85263fa66/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/text/correction/SuppressWarningsSubProcessor.java#L67 and copied it over here. To have it be a "move" you'd need to remove the method and its helpers from that location. Then, everywhere in the codebase that called that method, would get replaced with the new location here. (eg. `SuppressWarningsSubProcessor.addSuppressWarningsProposals` -> `SuppressWarningsSubProcessorCore.addSuppressWarningsProposals`.

Also, what about `addUnknownSuppressWarningProposals` & `addRemoveUnusedSuppressWarningProposals` ?

hopehadfield: I'll update that.

I didn't move `addUnknownSuppressWarningProposals` & `addRemoveUnusedSuppressWarningProposals` because they weren't needed for the user request in https://github.com/eclipse-jdtls/eclipse.jdt.ls/issues/2698. I can certainly move these as well to be used for additional suggestions in quickfix.

rgrunber: Fair. If it's not too much extra effort to move them as well, I would go for it.

.project
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
akurtakov, line 1: Please remove this file.

org.eclipse.jdt.core.manipulation/proposals/org/eclipse/jdt/internal/ui/text/correction/SuppressWarningsSubProcessorCore.java
```diff
@@ -38,4 +72,204 @@ public static final boolean hasSuppressWarningsProposal(IJavaProject javaProject
...
+	 *
+	 * @param cu the compilation unit
+	 * @param node the node on which to add a SuppressWarning token
+	 * @param warningToken the warning token to add
+	 * @param relevance the proposal's relevance
+	 * @param proposals collector to which the proposal should be added
+	 * @return <code>0</code> if no further proposals should be added to parent nodes, or the
+	 *         relevance of the next proposal
+	 *
+	 * @since 3.6
```
akurtakov, line 130: FYI - since tags on internal classes and methods are not required. Furthermore as it's copied from another bundle with different version it's plain wrong.

org.eclipse.jdt.ui.tests.refactoring/resources/InlineMethodWorkspace/TestCases/binary/.gitignore
```diff
@@ -0,0 +1 @@
+/classes/
```
akurtakov, line 1: Is this one really needed?

akurtakov (commented):
Some formal review from me as I haven't looked into the details.

org.eclipse.jdt.core.manipulation/proposals/org/eclipse/jdt/internal/ui/text/correction/SuppressWarningsSubProcessorCore.java
hopehadfield: Thanks for letting me know. Removed it.

robstryker:
Just to note, many of the other sub processors that are being converted have been switched to a base subprocessor which is abstract, and has abstract methods to create the proposals (or more accurately, whatever the generic required is). 

One of the reasons for this is that is how jdt.ls needs to handle their proposals. They eventually wrap all of their proposals into a proposal-kind wrapper structure called `ProposalKindWrapper`. The problem is that jdt.ls might have no way to know which of the various returned proposals are quick fixes or hints or refactors or whatever. 

For example, let's say that one of these static methods might return anywhere from 1 to 5 proposals (based on the situation), how would jdt.ls consume it? Let's say that jdt.ls would clasify proposals 1 and 3 as `CodeActionKind.QuickFix` and the other 3 as `CodeActionKind.Refactor`.   

What if the base implementation returns proposal 1 and 2, but not 3, 4, or 5.  The jdt.ls subprocessor would need to investigate proposals 1 and 2 to discover what code-action type to classify it as. How would it do this? What if both were `ASTRewriteCorrectionProposalCore` objects with no clear way to identify which one was which? 

Again, I'm not opposed to having a simple non-abstract core version of a subprocessor for simple cases. I do think it's not great if 6 of our subprocessors are using a class heirarchy with generics and 1 of them is using a standalone core version, but, I can ignore that. My real concern here is how jdt.ls will properly consume it in the case that it needs to classify some of these differently and has no real way to distinguish which ones are classified in which way? 

With the already-migrated versions, we have the abstract base class, the UI class that overrides all the creation of the various proposals, and other extenders (like jdt.ls) that might need to add ANY custom data of their own to the creation of a given proposal, at the time that proposal is created.  The simple standalone core subprocessor doesn't give extenders the ability to wrap a given proposal with their own custom metadata at the time of creation. They can only try to do it after the fact, which may be difficult.

org.eclipse.jdt.core.manipulation/proposals/org/eclipse/jdt/internal/ui/text/correction/SuppressWarningsBaseSubProcessor.java
```diff
@@ -0,0 +1,254 @@
...
+import org.eclipse.jdt.core.dom.rewrite.ASTRewrite;
+
+import org.eclipse.jdt.internal.core.manipulation.util.BasicElementLabels;
+import org.eclipse.jdt.internal.corext.dom.ASTNodes;
+import org.eclipse.jdt.internal.corext.util.JavaModelUtil;
+import org.eclipse.jdt.internal.corext.util.Messages;
+
+public abstract class SuppressWarningsBaseSubProcessor<T> {
+
+	protected SuppressWarningsBaseSubProcessor() {
```
rgrunber, line 56: Is this constructor actually needed ? If I remove it everything seems to work.

rgrunber (commented):
Overall, looks good to me.

rgrunber (approved):
I've tested this out and it works well. It follows the same approach established by other simplifications to other subprocessors. I think this is ready to be merged for the M3 time frame, so I'll do that once the branch opens up.

@hopehadfield last thing I can think of is to just update the copyrights "${YEAR}" -> "${YEAR}, 2024" for any touched file.

rgrunber (approved):
LGTM, so I'll be merging soon. The errors seem to be related to a new task that's not related to this change.

Issue #1043, opened 2024-01-05
"Make Static" refactoring adds new parameter documentation as first javdoc entry


HeikoKlare:
The recently introduced "Make Static" refactoring (#689) adds the documentation of a new parameter for the `this` object at the beginning of the javadoc comment, although this is usually not the expected position of a parameter documentation.

This is a finding that has been made during verification of the original refactoring contribution: https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/689#discussion_r1428535735

### Expected Behavior

The refactoring should add the parameter documentation in javadoc at a "reasonable" position, such as between the general documentation and potentially existing other parameter and return value documentations.

I propose to reuse the behavior already implemented for the "Change Method Signature" refactoring, as it already places the documentation for added parameters at reasonable positions.

### How to Reproduce

The behavior can be reproduced when refactoring any method with an existing javadoc comment. 
The results when applying it to 
`org.eclipse.jdt.core.tests.compiler.regression.AbstractRegressionTest#getCompilerTestsPluginDirectoryPath()` with some extended documentation look as follows.

#### Before:
```java
/**
* Returns the OS path to the directory that contains this plugin.
*
* @return OS path of the plugin directory
* @author unknown
*/
```

#### After:
```java
/**
* @param abstractRegressionTest
* Returns the OS path to the directory that contains this plugin.
*
* @return OS path of the plugin directory
* @author unknown
*/
```

#### Result when a adding parameter with "Change Method Signature" refactoring:

```java
/**
* Returns the OS path to the directory that contains this plugin.
* @param newParameter TODO
*
* @return OS path of the plugin directory
* @author unknown
*/
```

PR #1084, opened 2024-01-17
Improve Javadoc generated by "Make Static" refactoring #1043

HeikoKlare:
## What it does

This change improves the Javadoc generation behavior of the "Make Static" refactoring and aligns it with other refactorings such as "Change Method Signature". It ensures that generated parameter tags are inserted at proper positions within an existing Javadoc comment.
<s>This change improves the Javadoc generation behavior of the "Make Static" refactoring and aligns it with other refactorings such as "Change Method Signature". This includes (1) adding a "TODO" to added parameter tags and (2) inserting parameter tags at proper positions within an existing Javadoc comment.</s>
The tag for the parameter itself is added as the first parameter tag, following any author or version tags. The tags for potential type parameters are added as the last parameter tags, preceding any other tags than author or version.

The reused functionality from `ChangeMethodSignatureRefactoringContribution` is factored out to the `JavadocUtil` class.

Fixes https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1043

## How to test

The behavior can be reproduced by applying the "Make Static" refactoring to any method that when refactored requires a `this` parameter and has Javadoc into which the documentation for the added parameter has to be inserted. To see different flavors of the behavior, it's easiest to consider the regression test cases provided with this PR.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/refactoring/code/makestatic/ChangeCalculator.java
+++ b/org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/refactoring/code/makestatic/ChangeCalculator.java
@@ -62,6 +62,7 @@ import org.eclipse.jdt.internal.corext.dom.ModifierRewrite;
 import org.eclipse.jdt.internal.corext.refactoring.RefactoringCoreMessages;
 import org.eclipse.jdt.internal.corext.refactoring.base.ReferencesInBinaryContext;
 import org.eclipse.jdt.internal.corext.refactoring.code.TargetProvider;
+import org.eclipse.jdt.internal.corext.refactoring.util.JavadocUtil;
 import org.eclipse.jdt.internal.corext.refactoring.util.TextEditBasedChangeManager;
 import org.eclipse.jdt.internal.corext.util.Messages;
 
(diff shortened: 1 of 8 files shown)


HeikoKlare:
Thank you for reviewing and merging all the polishing PRs for the "Make Static" refactoring, @jjohnstn!
This one will fix the Javadoc issue you have pointed to in the original feature PR: https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/689#discussion_r1428535735
Together with #1085, every issue with the refactoring that I currently know about should be fixed (see also https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/689#issuecomment-1895849551).

jjohnstn (changes requested):
Do not add the TODO description for any added parameter.  This won'tt flag anyone via warnings/etc. to fix it later.  If one creates a Javadoc comment on a method, it will create the parameter tags without any comment.  You should be consistent and do the same.

jjohnstn:
Hi @HeikoKlare I don't know about the other refactorings/proposals, but I do know if I create Javadoc for a method with parameters, it does not add TODO tags.  Similarly, if I add a new parameter and then hover over it to add documentation for it, it again does not create the TODO tag.  JavadocUtil.createParamTag() has a TODO note to eventually change to add a {@todo} tag which might make sense for some of the other cases.

jjohnstn (approved):

HeikoKlare:
Yes, almost every actions that introduced Javadoc tags does not add TODO tags. That's why I totally agree with you that this refactoring should not do it either and adapted the code accordingly.

> JavadocUtil.createParamTag() has a TODO note to eventually change to add a {@todo} tag which might make sense for some of the other cases.

Note that this TODO note is from 2004 and will probably not be addressed. The code does not as if there is any systematics behind using this utility method and adding the TODO description, but I guess it was an early-days idea of an implementor of a single refactoring to provide some kind of "TODO" for added javadoc tags but that has never been implemented properly. And now there are few places (exactly the two refactorings I mentioned) where this "TODO" description is added. Since no one complained so far, there is no real demand to change it. I just wanted to offer to unify this to have a more consistent behavior of the javadoc generation of all refactorings.

Issue #1083, opened 2024-01-17
"Make Static" refactoring erroneously changes method call


HeikoKlare:
The recently introduced "Make Static" refactoring (#689) erroneously interprets a method call within an anonymous class inside the refactored method as a call on `this` and thus changes the method invocation to be performed on the added input parameter of the method. This is (1) semantically incorrect and (2) leads to compile errors if the type of the added parameter does not provide the method.

### Expected Behavior

The refactoring should only replace call to methods that are actually performed on `this`.
To this end, the implementation of `InstanceUsageRewriter#isInsideAnonymousClass(...)` needs to be corrected.

### How to Reproduce

The behavior was found when refactoring the following method:
`org.eclipse.jdt.core.tests.compiler.regression.AbstractRegressionTest#findReferences(String)`

It can be reproduced with the following minimal examples:
### Incorrect Call
When calling a method inside an anonymous class that is provided by the encapsulating class, there will be a wrong method call.

```java
public class Foo {
	
	void toBeRefactored() {
		new Foo() {
			void toImplement() {
				toCall();
			}
		};
	}
	
	void toCall() { }
	
	void toImplement() { }
	
}
```

Refactoring `toBeRefactored()` produces:
```java
public class Foo {
	
	static void toBeRefactored(Foo foo) {
		new Foo() {
			void toImplement() {
				foo.toCall();
			}
		};
	}
	
	void toCall() { }
	
	void toImplement() { }
	
}
```
But actually the call to `toBeCalled()` was made on an instance of the anonymous inner class rather than the outer instance in the `toBeRefactored` method. So in this case the refactoring should only add the `static` keyword and nothing else.

### Compile Error
When calling a method inside an anonymous class that is not provided by the encapsulating class, there will be a compile error (for the same reason), also in case the anonymous class is nested:

```java
public class Foo {
	int j = 0;
	
	void toBeRefactored() {
		this.j = 1;
		new Other() {
			@Override
			void toImplement() {
				new Object() {
					@Override
					public boolean equals(Object obj) {
						toCall();
						return false;
					}
				};
			}
		};
	}
}

public abstract class Other {
	abstract void toImplement();
	void toCall() {
	}
}
```

Refactoring `toBeRefactored()` produces:
```java
public class Foo {
	int j = 0;
	
	void toBeRefactored(Foo foo) {
		foo.j = 1;
		new Other() {
			@Override
			void toImplement() {
				new Object() {
					@Override
[...]

PR #1085, opened 2024-01-17
Correct "Make Static" refactoring behavior for anonymous classes #1083

HeikoKlare:
## What it does

The "Make Static" refactoring erroneously interprets a method call within an anonymous class inside the refactored method as a call on `this` and thus changes the method invocation to be performed on the added input parameter of the method. This is (1) semantically incorrect and (2) leads to compile errors if the type of the added parameter does not provide the method.

With this fix, the identification of whether a method invocation is performed on the `this` object of the method to be refactored and not some instance of an anonymous inner class inside that method is corrected. For three different scenarios (anonymous class extending outer class, anonymous class extending other class, and nested anonymous classes), regression tests are added.

Fixes https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1083

## How to test
As explained in #1083, the issue can be seen when applying the refactoring to `org.eclipse.jdt.core.tests.compiler.regression.AbstractRegressionTest#findReferences(String)`. Without this fix, a faulty method call will be introduced, whereas the refactoring produces a correct result with this fix. Different other scenarios, in which the issue occurs, can be seen in the regression tests added with this PR.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/refactoring/code/makestatic/InstanceUsageRewriter.java
+++ b/org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/refactoring/code/makestatic/InstanceUsageRewriter.java
@@ -15,6 +15,7 @@ package org.eclipse.jdt.internal.corext.refactoring.code.makestatic;
 import org.eclipse.jdt.core.dom.AST;
 import org.eclipse.jdt.core.dom.ASTNode;
 import org.eclipse.jdt.core.dom.ASTVisitor;
+import org.eclipse.jdt.core.dom.AbstractTypeDeclaration;
 import org.eclipse.jdt.core.dom.AnonymousClassDeclaration;
 import org.eclipse.jdt.core.dom.ClassInstanceCreation;
 import org.eclipse.jdt.core.dom.Expression;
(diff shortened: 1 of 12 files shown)


org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/refactoring/code/makestatic/InstanceUsageRewriter.java
```diff
@@ -236,15 +230,22 @@ private void modifyInstanceMethodUsage(SimpleName node, IBinding binding) {
...
-		if (declaringClass != null) {
-			boolean isLocal= declaringClass.isLocal();
-			boolean isAnonymous= declaringClass.isAnonymous();
-			boolean isMember= declaringClass.isMember();
-			boolean isNested= declaringClass.isNested();
-			boolean isTopLevel= declaringClass.isTopLevel();
-			if (!isTopLevel || isNested || isAnonymous || isLocal || isMember) {
-				return true;
+	private boolean isAccessToAnonymousClass(ASTNode node, ITypeBinding declaringClass) {
+		ITypeBinding typeOfRefactoredMethod = fTargetMethodDeclaration.resolveBinding().getDeclaringClass();
```
jjohnstn, line 234: You have to verify that resolveBinding() does not return null before calling getDeclaringClass()

jjohnstn (changes requested):
Need a null-check for resolveBinding(), otherwise looks good.

org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/refactoring/code/makestatic/InstanceUsageRewriter.java
HeikoKlare: Fixed with a690054.

jjohnstn: Thanks.   Marking this as resolved.

jjohnstn (approved):

Issue #1082, opened 2024-01-17
[win] [21] New Java Project Dialog too small


jukzi:
"New Java Project Dialog" shows scroll bars now
eclipse.buildId=4.31.0.I20240115-1800
java.version=21.0.1
![image](https://github.com/eclipse-jdt/eclipse.jdt.ui/assets/51790620/5ffe4e8c-7eaf-4761-8df6-6c776ef4d523)

should be big enough by default:
![image](https://github.com/eclipse-jdt/eclipse.jdt.ui/assets/51790620/aba1d71b-ddff-4bdf-aa29-e37b587bf2ec)

iloveeclipse:
Same problem on Linux:
![image](https://github.com/eclipse-jdt/eclipse.jdt.ui/assets/964108/64471693-e688-4411-8955-76d68043f8b5)

iloveeclipse:
This looks like a regression in Eclipse 4.31.

iloveeclipse:
I believe the regression is coming from https://github.com/eclipse-platform/eclipse.platform.ui/pull/1283, at least if I return `null` from `org.eclipse.jface.wizard.WizardDialog.getDialogBoundsSettings()` the dialog size changes to somewhat more acceptable.

marcushoepfner:
I will have a look.

marcushoepfner:
That does only happen if you select the exact wizard directly correct?
I can reproduce it with e.g. File-> new Java Project.
If I select File -> New -> Other (CMD/CTRL + N) and select a wizard, the dialog is resized to the correct size.

The dialog settings identifier is always "wizard.dialogbounds", no matter which wizard. I guess this needs to be changed.

jukzi:
Hard to say since the dialog seems to keep it't size after i used it once. i initially had the problem when i opened it from context menu in package explorer.

marcushoepfner:
Issue in o.e.platform.ui has been created.
Refer to https://github.com/eclipse-platform/eclipse.platform.ui/issues/1511.

I think we can close this one.

iloveeclipse:
@marcushoepfner : thanks for looking into.

PR #1087, opened 2024-01-17
fix "New Java Project" Dialog too small [#1082]

jukzi:
https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1082

### DIFF
--- a/org.eclipse.jdt.ui/ui/org/eclipse/jdt/ui/wizards/NewJavaProjectWizardPageOne.java
+++ b/org.eclipse.jdt.ui/ui/org/eclipse/jdt/ui/wizards/NewJavaProjectWizardPageOne.java
@@ -1831,4 +1831,9 @@ public class NewJavaProjectWizardPageOne extends WizardPage {
 
 		return true;
 	}
+
+	@Override
+	public  org.eclipse.swt.graphics.Point getMinimumPageSize() {
+		 return new org.eclipse.swt.graphics.Point(635,845);
+	}
 }


jukzi:
@iloveeclipse can you test on linux, please

org.eclipse.jdt.ui/ui/org/eclipse/jdt/ui/wizards/NewJavaProjectWizardPageOne.java
```diff
@@ -1831,4 +1831,9 @@ private static boolean isValidWorkingSet(IWorkingSet workingSet) {
 
 		return true;
 	}
+
+	@Override
+	public  org.eclipse.swt.graphics.Point getMinimumPageSize() {
+		 return new org.eclipse.swt.graphics.Point(635,845);
```
akurtakov, line 1837: Such hardcoding makes the change impossible to change as minimum sizes are calculated based on native widgets and that is affected even by changing a UI theme. I strongly believe this is wrong workaround and it will bite us back soon enough if applied.

iloveeclipse:
> can you test on linux, please

The absolute dialog dimensions are not cross-platform and will not work with HiDPI anyway.
I believe the regression is coming from https://github.com/eclipse-platform/eclipse.platform.ui/pull/1283 , at least if I return null from `org.eclipse.jface.wizard.WizardDialog.getDialogBoundsSettings()` the dialog size changes to somewhat more acceptable.

iloveeclipse (changes requested):
The fix should be in platfrom, where regression is coming from.

PR #1090, opened 2024-01-18
Use the ECJ generated compiler logs to parse compiler problems

laeubi:
It is more reliable to pare a structured file than the log output.


jukzi:
thanks @laeubi  , that seems to be a working workaround for
https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1717

The info markers are now rendered correctly:
image
We might even consider to reduce the ecj qualityGate to LOW if this keeps working.

PR #1092, opened 2024-01-18
Jenkinsfile: archive compilelogs

jukzi:


jukzi:
was only experiment, which showed multiple compiler log files named @dot.xml
image
image

PR #1101, opened 2024-01-18
remove Exceptions not actually thrown

jukzi:
eliminates problem markers and dead code


org.eclipse.jdt.core.manipulation/common/org/eclipse/jdt/core/manipulation/internal/javadoc/CoreJavadocAccess.java
```diff
@@ -304,11 +304,11 @@ protected StringBuffer createSuperMethodReferences(final IMethod method) throws
 		return createSuperMethodReferencesHTML(data.superInterfaceMethods(), data.superClassMethod());
 	}
 
-	protected StringBuffer createSuperMethodReferencesHTML(ArrayList<IMethod> superInterfaceMethods, IMethod superClassMethod) throws JavaModelException {
```
rgrunber, line 307: I think this would break https://github.com/eclipse-jdtls/eclipse.jdt.ls/blob/8f0f822a89bf1f6109a9d95d048e7b53110df19f/org.eclipse.jdt.ls.core/src/org/eclipse/jdt/ls/core/internal/javadoc/JavadocContentAccess2.java#L117-L121 . @robstryker same as before, we should probably document these.

jukzi:
@robstryker can we agree on a timeline, for example a merge conflict until 2024-01-26 ?

jukzi:
may i submit?

Issue #1093, opened 2024-01-18
TypeFilter: java.lang.IllegalArgumentException at shutdown


jukzi:
logged at the end of QuickFixTestSuite:
```
java.lang.IllegalArgumentException
	at org.eclipse.core.internal.preferences.AbstractScope.getNode(AbstractScope.java:36)
	at org.eclipse.jdt.internal.corext.util.TypeFilter.getPreferenceStore(TypeFilter.java:107)
	at org.eclipse.jdt.internal.corext.util.TypeFilter.dispose(TypeFilter.java:140)
	at org.eclipse.jdt.internal.core.manipulation.JavaManipulationPlugin.stop(JavaManipulationPlugin.java:81)
	at org.eclipse.osgi.internal.framework.BundleContextImpl$3.run(BundleContextImpl.java:893)
	at org.eclipse.osgi.internal.framework.BundleContextImpl$3.run(BundleContextImpl.java:1)
	at java.base/java.security.AccessController.doPrivileged(AccessController.java:569)
	at org.eclipse.osgi.internal.framework.BundleContextImpl.stop(BundleContextImpl.java:885)
	at org.eclipse.osgi.internal.framework.EquinoxBundle.stopWorker0(EquinoxBundle.java:1071)
	at org.eclipse.osgi.internal.framework.EquinoxBundle$EquinoxModule.stopWorker(EquinoxBundle.java:392)
	at org.eclipse.osgi.container.Module.doStop(Module.java:695)
	at org.eclipse.osgi.container.Module.stop(Module.java:554)
	at org.eclipse.osgi.container.ModuleContainer$ContainerStartLevel.decStartLevel(ModuleContainer.java:2111)
	at org.eclipse.osgi.container.ModuleContainer$ContainerStartLevel.doContainerStartLevel(ModuleContainer.java:1981)
	at org.eclipse.osgi.container.SystemModule.stopWorker(SystemModule.java:278)
	at org.eclipse.osgi.internal.framework.EquinoxBundle$SystemBundle$EquinoxSystemModule.stopWorker(EquinoxBundle.java:222)
	at org.eclipse.osgi.container.Module.doStop(Module.java:695)
	at org.eclipse.osgi.container.Module.stop(Module.java:554)
	at org.eclipse.osgi.container.SystemModule.stop(SystemModule.java:209)
[...]

PR #1122, opened 2024-01-24
Don't set preference key of other bundle on shutdown

iloveeclipse:
The other bundle might not be shutdown yet and would still want that key.

Fixes https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1093


iloveeclipse:
@rgrunber , @robstryker : please take a look.

rgrunber:
As per https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1093#issuecomment-1908430879 , I think it's something I built into the logic upon request, and I can't see it breaking JDT-LS, but let me quickly try and see.

Issue #1080, opened 2024-01-17
Allow Batch Field Encapsulation for Multiple Fields


amartya4256:
## Current Situation
JDT provides two functionalities for the generation of getters and setters.
1. Apply the _Encapsulate Field_ refactoring, which generates getters and setters and, in particular, allows to replace references to the field with calls of the generated accessors.
![Image](https://github.com/vi-eclipse/Eclipse-JDT/assets/755472/ecc34de3-151f-4769-a323-de10a841b2ca)
2. Apply the _Generate Getters/Setters_ source modification, which allows to add getters and setters to the file.
![Image](https://github.com/vi-eclipse/Eclipse-JDT/assets/755472/82c72fe6-ba70-4669-ac2b-d084fbb6eda6)

While getter/setter generation can be applied in a batch operation for multiple fields of a class, the refactoring can only be applied to a single field at once.

## Expected Behavior
The user shall be able to perform a getter/setter generation together with field encapsulation (i.e., replacement of field access with calls of accessors) in batch mode to multiple fields of a class.

## Proposed Solution
Starting from the two existing functionalities, either the getter/setter generation may be extended by functionality to encapsulate the fields, or the field encapsulation refactoring may be enhanced with support for being applied to multiple fields as once.

I propose to follow to second approach, as performing field encapsulation actually is a refactoring that goes beyond source manipulation. I.e., in addition (or as a replacement for the) to the existing _Encapsulate Field_ refactoring, an _Encapsulate Fields_ refactoring should be provided. As a result, the options provided by getter/setter generation and field encapsulation should be aligned. The new refactoring should be capable of everything provided by the existing refactoring but along with a selection of all fiels to apply it to, i.e. :
- Selecting the fields to encapsulate (in case of a text selection or when triggered as a quick fix, may be restricted to the fields in the text selection)
- Per field:
  - Defining wheter a getter and/or setter shall be created along with their names
  - A selection where the method(s) shall be placed

PR #1124, opened 2024-01-24
Implemented Batch SelfEncapsulateField for multiple fields at once #1080

amartya4256:
Issue: https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1080

## What it does

Currently with Refactoring -> Encapsulate, encapsulating only one field at a time is possible. With this PR, it allows to select multiple fields and perform encapsulation on them in a batch.

## How to test

Select multiple fields at a time and click on Refactor -> Encapsulate fields in the menu. A wizard would appear to select the fields and specify how to perform the operation. Once done, click okay. This should encapsulate all the selected fields.

image

## How it works
I tried many ways to implement this. The first approach was to select all the fields needed to be refactored and then roll out the operation in a loop. However, multiple change operations were registered in the undo manager and thus it required multiple undoing to undo the result. Another approach was to create a composite refactoring which consists the information of all the single field refactoring and create a composite change consisting of all the changes from each refactoring. However, the changes were still rolled out one by one since their textChange object were different for each refactoring. And thus changes made by one refactoring would change something in the file and the next refactoring's change object will lose the context of the field which has already been registered in its TextChange object. Thus, it would make faulty changes.

I tried a few more approaches until I realized, that all the refactoring use their ASTRewrite per compilationUnit to register their changes and then create a TextEdit from all the changes registered in the ASTRewrite. My idea was to share the ASTRewrite, ImportRewrite, ASTs and several resources which are needed to create changes among all the SelfEncapsulateFieldRefactorings. These resources are managed by the composite refactoring and the SelfEncapsulateFieldRefactorings can use the relevant resources and write their changes there. Once every refactoring has finished their changes, we use these resources in the composite refactoring to create a single Change Object per compilation unit to roll out the refactoring changes. Thus, it also registers only one Undo Operation.

## Limitations
**Update: This limitation no longer exists. At first we collect all the getter setter changes and then we change the access modifiers for all the fields.**
Since we pass the insertion index of the getter and setter to the rewriter, in a case where we have multiple fields in a declaration like int i, j; together, they are split in 2 lines like private int i; private int j;. It may change the insertion index of the getter and setter. However, this behaviour doesn't end to any failure but only the getter and setter are inserted at a wrong index (one index before the one where they are supposed to be inserted).

## Author checklist


org.eclipse.jdt.ui/ui/org/eclipse/jdt/ui/actions/SelfEncapsulateFieldAction.java
```diff
@@ -171,4 +195,12 @@ public void run(IField field) {
 			return;
 		RefactoringExecutionStarter.startSelfEncapsulateRefactoring(field, getShell());
 	}
+
+	/**
+	 * @since 3.32
+	 */
+	public void run(List<IField> fields, List<IField> preselected) {
```
jukzi, line 202: can this be private - non API?

```
	/*
	 * Should be private. But got shipped in this state in 2.0 so changing this is a
	 * breaking API change.
	 */
	public void run(IField field) {
```

amartya4256: Yes, just did it.

jukzi:
please amend commit message to contain "#1080" in the title and "https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1080" in the description

jukzi:
@jjohnstn can you please review or suggest a reviewer? i like the idea of the PR

jjohnstn:
@jukzi I'll take a look

jjohnstn (changes requested):
Trying it out first.  Overall, good functionality.  Initial comments:

1.  The information about a setter being disabled due to it being a final field is missing.  I am wondering if you could add the disabled checkbox or add a message to the field node.
2.  Hitting back after preview doesn't bring me back to selection dialog.  First back has the file name specified and second back causes an error to occur.  This occurs in current code as well.  A separate issue could be opened if the dialog isn't at fault.
3. When selecting multiple fields in the file, only the last one is default selected in the dialog.  It would be nice to have all selected already checked in the dialog.
4. I have a package visibility field and public is selected for getter/setter in the dialog.  When I choose it, the field gets changed to private.  This happens in current code as well but should be looked at as something is off.  Possibly a separate issue opened.
5. Current behavior on single field narrows down the possible visibility setting so the getter/setter cannot have less accessibility than the field itself.  I understand your logic in supplying all options but you should warn when fields are selected and the getter/setters become less visible due to the setting or else disable/remove fields from the tree.
6. Name of refactoring should be changed to Encapsulate Fields

amartya4256:
> Trying it out first. Overall, good functionality. Initial comments:
> 
> 1. The information about a setter being disabled due to it being a final field is missing.  I am wondering if you could add the disabled checkbox or add a message to the field node.
> 2. Hitting back after preview doesn't bring me back to selection dialog.  First back has the file name specified and second back causes an error to occur.  This occurs in current code as well.  A separate issue could be opened if the dialog isn't at fault.
> 3. When selecting multiple fields in the file, only the last one is default selected in the dialog.  It would be nice to have all selected already checked in the dialog.
> 4. I have a package visibility field and public is selected for getter/setter in the dialog.  When I choose it, the field gets changed to private.  This happens in current code as well but should be looked at as something is off.  Possibly a separate issue opened.
> 5. Current behavior on single field narrows down the possible visibility setting so the getter/setter cannot have less accessibility than the field itself.  I understand your logic in supplying all options but you should warn when fields are selected and the getter/setters become less visible due to the setting or else disable/remove fields from the tree.
> 6. Name of refactoring should be changed to Encapsulate Fields

Regarding the second point you made, I can't reproduce the behaviour. When I hit back, it goes back to the selection page.

jjohnstn:
> > Trying it out first. Overall, good functionality. Initial comments:
> > 
> > 1. The information about a setter being disabled due to it being a final field is missing.  I am wondering if you could add the disabled checkbox or add a message to the field node.
> > 2. Hitting back after preview doesn't bring me back to selection dialog.  First back has the file name specified and second back causes an error to occur.  This occurs in current code as well.  A separate issue could be opened if the dialog isn't at fault.
> > 3. When selecting multiple fields in the file, only the last one is default selected in the dialog.  It would be nice to have all selected already checked in the dialog.
> > 4. I have a package visibility field and public is selected for getter/setter in the dialog.  When I choose it, the field gets changed to private.  This happens in current code as well but should be looked at as something is off.  Possibly a separate issue opened.
> > 5. Current behavior on single field narrows down the possible visibility setting so the getter/setter cannot have less accessibility than the field itself.  I understand your logic in supplying all options but you should warn when fields are selected and the getter/setters become less visible due to the setting or else disable/remove fields from the tree.
> > 6. Name of refactoring should be changed to Encapsulate Fields
> 
> @jjohnstn
> 
>     * Regarding the second point you made, I can't reproduce the behaviour. When I hit back, it goes back to the selection page.
> 

I am still seeing it.  I am using a child Eclipse using latest sources and using 4.31-I-builds as a target.  Perhaps it is something that has been introduced due to a platform change.

>     * Regarding the third point, I reused the selection dispatch action as it is used in generate getter and setter. Similar behaviour can also be found there. Do you think we can create a separate issue for this?

Sure.

amartya4256:
> > > Trying it out first. Overall, good functionality. Initial comments:
> > > 
> > > 1. The information about a setter being disabled due to it being a final field is missing.  I am wondering if you could add the disabled checkbox or add a message to the field node.
> > > 2. Hitting back after preview doesn't bring me back to selection dialog.  First back has the file name specified and second back causes an error to occur.  This occurs in current code as well.  A separate issue could be opened if the dialog isn't at fault.
> > > 3. When selecting multiple fields in the file, only the last one is default selected in the dialog.  It would be nice to have all selected already checked in the dialog.
> > > 4. I have a package visibility field and public is selected for getter/setter in the dialog.  When I choose it, the field gets changed to private.  This happens in current code as well but should be looked at as something is off.  Possibly a separate issue opened.
> > > 5. Current behavior on single field narrows down the possible visibility setting so the getter/setter cannot have less accessibility than the field itself.  I understand your logic in supplying all options but you should warn when fields are selected and the getter/setters become less visible due to the setting or else disable/remove fields from the tree.
> > > 6. Name of refactoring should be changed to Encapsulate Fields
> > 
> > 
> > @jjohnstn
> > ```
> > * Regarding the second point you made, I can't reproduce the behaviour. When I hit back, it goes back to the selection page.
> > ```
> 
> I am still seeing it. I am using a child Eclipse using latest sources and using 4.31-I-builds as a target. Perhaps it is something that has been introduced due to a platform change.
> 
> > ```
> > * Regarding the third point, I reused the selection dispatch action as it is used in generate getter and setter. Similar behaviour can also be found there. Do you think we can create a separate issue for this?
> > ```
> 
> Sure.

I have made the changes and for point 2, 3 and 4, i would create issues and link it here shortly.

org.eclipse.jdt.ui/ui refactoring/org/eclipse/jdt/internal/ui/refactoring/refactoringui.properties
```diff
@@ -109,11 +109,13 @@ SelfEncapsulateFieldInputPage_public=&public
 SelfEncapsulateFieldInputPage_default=p&ackage
 SelfEncapsulateFieldInputPage_protected=pro&tected
 SelfEncapsulateFieldInputPage_private=pri&vate
+SelfEncapsulateFieldInputPage_same=&same as field
 SelfEncapsulateFieldInputPage_generateJavadocComment=Generate method &comments
 SelfEncapsulateFieldInputPage_useexistingsetter_label=(using existing setter)
 SelfEncapsulateFieldInputPage_useexistinggetter_label=(using existing getter)
 SelfEncapsulateFieldInputPage_usenewgetter_label=(new setter created)
 SelfEncapsulateFieldInputPage_usenewsetter_label=(new getter created)
+SelfEncapsulateFieldInputPage_access_warning=Some getters/setters are less visible than their field
```
jjohnstn, line 118: Change this to be ... less visible than their original field

since we are resetting the field to be private, it will make more sense to the end-user

jjohnstn (changes requested):
Hi @amartya4256 Thanks for the fixes.  I like the (same as field setting) and the warning for setting a field less than its current visibility.  I get a formatting problem, however where I am setting the text entry for the setter of the last field showing up at the top of the tree area in the center when I start (e.g. my last field is d2 and I see a text entry with setD2 specified to the right of the first field which hasn't been expanded).   This vanishes when I click on something in the tree and expand it and everything works great from there.

amartya4256:
https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1173

Here's an issue for selecting all the highlighted fields

jjohnstn:
Formatting problem


Screenshot from 2024-02-06 13-47-05

jjohnstn:
Hi @amartya4256 I am on RHEL 7.9.  From the picture, you'll see that it shows the setD2 text field at the top of the tree area.

amartya4256:
Hi @jjohnstn, I had a look into it. I have Windows 11 and I can't reproduce it there. Have you tried it on windows? I'll set up a WSL and try to reproduce it for Unix. 
From the code perspective, since I am inserting items straight forward on the basis of their hierarchy, this kinda view is bizzare. I debugged and the tree structure seems fine. Could also be some rendering problem?

jjohnstn:
Hi @amartya4256 Unfortunately I don't have access to Windows or Mac.

amartya4256:
@jjohnstn @HeikoKlare 
The editor in the tree seemed a little vague since there was no information about which text editor is for setter or getter. So, I propose following design with the new commit. 
image

You can edit the getter or setter by selecting the field or its child as shown in the image and the names are updated on key release while typing in the text box.

jjohnstn (changes requested):
A number of issues for me with the new design.

1.The tree view is too small and doesn't expand when I make the dialog wider (just the text boxes expand)
2. On opening, it selects one field but the text boxes have no selection and show nothing
3. Opening any field in the tree view shows the getter/setter names but doesn't update the text box section

Screenshot from 2024-02-12 17-43-56

In my opinion, I prefer the previous design if the quirks were fixed as the current design seems cumbersome with the way it is, but I think that it is not possible due to some unreported SWT issues.

That said, I looked at existing preference dialogs and couldn't find any that had a tree view with editable items in the tree.  However, I did note a number of preference pages that use a table and then have the user hit an Edit... button to change column entries.  For example, see the Template Editor.  It has a table list with enablement of individual items.  In your case columns would be Field, Getter Name, Setter Name.  Then the user picks a selection and hits the Edit button which brings up a separate dialog.  This would be simple with the getter and setter name text boxes and for invalid setters you could disable the setter box and have an message in the dialog that explains why.  It also is consistent with existing pages and a user would certainly be comfortable using it.  You could copy the logic and save yourself the cross-platform headaches and SWT bugs.

jjohnstn (changes requested):
Looks much better, thanks.  A few things:

1. there is a quick assist for an individual field that is failing with a cast exception (select a field and do a CTRL+1...you'll see Create getter and setter for field
2. I would like to see the field that is selected also selected in the tree view and the names filled in upon opening the dialog to avoid having to select the field again

jjohnstn:
Hi @amartya4256 Thanks for the latest changes.  It is now working just how I expect it to and the quick assist is working well with just the one field being shown and selected.  We are too late to get this into 4.31 (M3 has passed).  Now that it is working well, I will do another review of the resultant code and a bit more testing.  It should easily be ready for merging for 4.32 M1 when the master branch opens after RC2.

jjohnstn (approved):

org.eclipse.jdt.core.manipulation/refactoring/org/eclipse/jdt/internal/corext/refactoring/sef/SelfEncapsulateFieldCompositeRefactoring.java
```diff
@@ -0,0 +1,229 @@
...
+	public RefactoringStatus checkFinalConditions(IProgressMonitor pm) throws CoreException, OperationCanceledException {
+		clearResources();
+		RefactoringStatus fFinalConditions = new RefactoringStatus();
+		for(SelfEncapsulateFieldRefactoring selfEncapsulateFieldRefactoring : getSelectedRefactorings()) {
+			fFinalConditions.merge(selfEncapsulateFieldRefactoring.checkFinalConditions(pm));
+		}
+		makeDeclarationsPrivate();
+		TextChangeManager fChangeManager = new TextChangeManager();
+		for(Entry<ICompilationUnit, ASTRewrite> entry : fRewritersMap.entrySet()) {
+			TextChange change= fChangeManager.get(entry.getKey());
```
jjohnstn, line 153: Can fChangeManager ever find an existing entry?  If so, what happens to the MultiTextEdit that was already created for it?  If there can't be an existing entry, simply use the TextChangeManager.get() logic when an entry is not found.

jjohnstn (changes requested):
Hi @amartya4256 this is the final code review I mentioned.  Some minor issues and some questions.  Once done or questions answered, the patch is ready to go for 4.32 M1

org.eclipse.jdt.core.manipulation/refactoring/org/eclipse/jdt/internal/corext/refactoring/sef/SelfEncapsulateFieldCompositeRefactoring.java
amartya4256: Yes there can be an existing entry. The fChangeManager.get method checks for if the key doesnt exist then it creates a new entry and returns it. And this entry can be of a compilation unit where all the fields are defined or some other compilation unit where they are referenced. We want to work on the same AST for each CU found. So it makes more sense to keep a local object which is specific to a set of change made by the refactoring. Hence, we should not make any changes here.

amartya4256:
> Hi @amartya4256 this is the final code review I mentioned. Some minor issues and some questions. Once done or questions answered, the patch is ready to go for 4.32 M1

Hi @jjohnstn I have made the required changes and answered your questions. Let me know if this is sufficient or if there's something else needed.

org.eclipse.jdt.core.manipulation/refactoring/org/eclipse/jdt/internal/corext/refactoring/sef/SelfEncapsulateFieldCompositeRefactoring.java
```diff
@@ -0,0 +1,229 @@
...
+		for(SelfEncapsulateFieldRefactoring selfEncapsulateFieldRefactoring : fRefactorings) {
+			fInitialConditions.merge(selfEncapsulateFieldRefactoring.checkInitialConditions(pm));
+		}
+		return fInitialConditions;
+	}
+
+	@Override
+	public Change createChange(IProgressMonitor pm) throws CoreException, OperationCanceledException {
+		pm.beginTask(NO_NAME, changes.size());
+		pm.setTaskName(RefactoringCoreMessages.SelfEncapsulateField_create_changes);
```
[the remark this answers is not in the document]
jjohnstn: Actually, I meant to remove the setTaskName().  According to IProgressMonitor docs, normally there is no need to call this method.  If you want the SelfEncapsulateField_create_changes message, just specify it in the beginTask since you never change it nor pass the pm anywhere.

jjohnstn:
> > Hi @amartya4256 this is the final code review I mentioned. Some minor issues and some questions. Once done or questions answered, the patch is ready to go for 4.32 M1
> 
> Hi @jjohnstn I have made the required changes and answered your questions. Let me know if this is sufficient or if there's something else needed.

Thanks @amartya4256 just a comment on the beginTask() review comment.  You don't need to use setTaskName().  See comment.

org.eclipse.jdt.core.manipulation/refactoring/org/eclipse/jdt/internal/corext/refactoring/sef/SelfEncapsulateFieldCompositeRefactoring.java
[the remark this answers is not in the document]
jjohnstn: Done

jjohnstn (approved):

jjohnstn:
@amartya4256 Thanks for the patch.  Would you like to contribute a N&N entry or would you like me to do it?

amartya4256:
@jjohnstn Sorry for the late response. Here is the N&N PR: https://github.com/eclipse-platform/www.eclipse.org-eclipse/pull/149

Issue #230, opened 2022-08-23
Java Code Assist: javaCompletionProposal throws runtime exception


nimo23:
Within the java editor, I get this error  when using **eclipse code assist**:

![Screenshot at Aug 23 16-08-12](https://user-images.githubusercontent.com/3045549/186180727-872ff036-800f-4f41-ad11-72c99cbb0ed1.png)

```
The 'org.eclipse.jdt.ui.org.eclipse.jdt.ui.javaCompletionProposalComputer.chain' proposal computer from the 'org.eclipse.jdt.ui' plug-in did not complete normally. The extension has thrown a runtime exception.

java.lang.NumberFormatException: Cannot parse null string
	at java.base/java.lang.Integer.parseInt(Integer.java:630)
	at java.base/java.lang.Integer.parseInt(Integer.java:786)
	at org.eclipse.jdt.internal.ui.text.java.ChainCompletionProposalComputer.executeCallChainSearch(ChainCompletionProposalComputer.java:152)
	at org.eclipse.jdt.internal.ui.text.java.ChainCompletionProposalComputer.computeCompletionProposals(ChainCompletionProposalComputer.java:77)
	at org.eclipse.jdt.internal.ui.text.java.CompletionProposalComputerDescriptor.computeCompletionProposals(CompletionProposalComputerDescriptor.java:348)
	at org.eclipse.jdt.internal.ui.text.java.CompletionProposalCategory.computeCompletionProposals(CompletionProposalCategory.java:340)
	at org.eclipse.jdt.internal.ui.text.java.ContentAssistProcessor.collectProposals(ContentAssistProcessor.java:333)
	at org.eclipse.jdt.internal.ui.text.java.ContentAssistProcessor.computeCompletionProposals(ContentAssistProcessor.java:289)
	at org.eclipse.jface.text.contentassist.ContentAssistant$2.lambda$0(ContentAssistant.java:2048)
	at java.base/java.util.Collections$SingletonSet.forEach(Collections.java:4905)
	at org.eclipse.jface.text.contentassist.ContentAssistant$2.run(ContentAssistant.java:2047)
	at org.eclipse.core.runtime.SafeRunner.run(SafeRunner.java:45)
	at org.eclipse.jface.text.contentassist.ContentAssistant.computeCompletionProposals(ContentAssistant.java:2044)
	at org.eclipse.jface.text.contentassist.CompletionProposalPopup.computeProposals(CompletionProposalPopup.java:577)
[...]

PR #1129, opened 2024-01-25
CompletionProposal: log error and keep going with other providers

jukzi:
for example
https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/230


jukzi:
Example output: Keeps showing (other) proposals, but logs error:

image

jukzi:
@rgrunber any objection?

rgrunber:
Yeah I'm fine with this. Making the completion proposal computers more fault tolerant is a great.

PR #1141, opened 2024-01-30
Cleanup in all -BaseSubProcessor classes and extenders

robstryker:
The recent pushdown of many SubProcessor / quickfix processors could benefit from more clarity and uniformity in some of the variable names. Several of the abstract methods have parameters like `boolean b` or `Object object`, and several of the extenders are choosing to ignore those parameters and instead pass in their own hard-coded values when constructing the proposals. 

This should be cleaned up ASAP before consumers get a chance to extend or interact with these (admittedly, internal) classes. 

## Author checklist


robstryker:
While this change should be safe, it is possible that something could break, so I will pay attention to the test suite to verify. The reason this could theoretically break something is that the primary subclasses of these base subprocessors previously ignored all parameters and selected the correct ones that made the test suite pass. 

By trusting the superclass to pass the correct parameters now, it is possible a regression could occur. I'll keep an eye on it. I did not see anything offhand, but I'd like to be sure.

robstryker:
It appears all tests pass, and the quality gate has "4 new warnings" which are the same 4 as in previous PRs. It appears it is not checking new warnings from commit to commit or PR to PR, but rather with some prior baseline? 

@jjohnstn  this patch is ready for review.

jjohnstn (approved):
Approved.

PR #1156, opened 2024-02-01
suppress "illegally extends CleanUpOptions"

jukzi:
CleanUpOptions @noextend is not intended to be subclassed by CLIENTS.


jukzi:
image

PR #1161, opened 2024-02-01
Incremental delete of duplicate interfaces part2 (Delete ICleanUpCore)

robstryker:
## What it does

Removes internal interface ICleanupCore as a duplicate. 

This PR depends on https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/1138 as it touches the same code

## Why this is needed ##
While pushing code related to the completion subprocessors from jdt.ui to jdt.core.manipulations, one specific (and hefty) subprocessor, LocalCorrectionsSubProcessor, uses a lot of proposals that wrap `CleanUp` objects, using either the `ICleanUp` or `ICleanUpCore` interfaces, as well as the `AbstractCleanUp` or `AbstractCleanUpCore` abstract classes. 

While identical interfaces are not much of a problem, having two parallel-but-slightly-different abstract classes leads to two parallel heirarchies, and clients can not simply extend both or support both very easily. While trying to untangle these specific issues, the myriad of other duplicate (and otherwise harmless) interfaces that are in the affected code also begin to add complexity to the solution. 

With that in mind, removing the most unoffensive interfaces first becomes the safest course of action, while at the same time building our way up to tackling the larger parallel-heirarchy issue via small and manageable changes to the heirarchy as the situation becomes less cluttered. 

## Author checklist


org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/util/CleanUpCoreWrapper.java
```diff
@@ -48,209 +25,16 @@
  * @since 3.31
  */
 public class CleanUpCoreWrapper {
```
rgrunber, line 27: I couldn't find these static `wrap(..)` calls used anywhere, so let me know if they're needed, or this entire class can be removed.

rgrunber (approved):
I've tested this out by playing with cleanup actions (and cleanup on save), and it all seems to be working as expected. That combined with the various cleanup tests makes me inclined to merge.

PR #1166, opened 2024-02-02
Delete internal abstract classes AbstractCleanUpCore and AbstractMultiFixCore

robstryker:
## What it does

Continue deletion and cleanup of unnecessary / duplicate classes, this time removing AbstractCleanUpCore and AbstractMultiFixCore

This PR depends on https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/1161  and https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/1138 being committed first

## Why this is needed ##
While pushing code related to the completion subprocessors from jdt.ui to jdt.core.manipulations, one specific (and hefty) subprocessor, LocalCorrectionsSubProcessor, uses a lot of proposals that wrap `CleanUp` objects, using either the `ICleanUp` or `ICleanUpCore` interfaces, as well as the `AbstractCleanUp` or `AbstractCleanUpCore` abstract classes. 

While identical interfaces are not much of a problem, having two parallel-but-slightly-different abstract classes leads to two parallel heirarchies, and clients can not simply extend both or support both very easily. While trying to untangle these specific issues, the myriad of other duplicate (and otherwise harmless) interfaces that are in the affected code also begin to add complexity to the solution. 

With that in mind, removing the most unoffensive interfaces first becomes the safest course of action, while at the same time building our way up to tackling the larger parallel-heirarchy issue via small and manageable changes to the heirarchy as the situation becomes less cluttered. 

## Author checklist


org.eclipse.jdt.core.manipulation/common/org/eclipse/jdt/internal/ui/fix/CleanUpFixWrapper.java
```diff
@@ -28,13 +27,13 @@
 import org.eclipse.jdt.internal.ui.text.correction.ProblemLocation;
 
 /**
- * Class to wrap an ICleanUpFixCore and be used as an ICleanUpFix
+ * Class to wrap an ICleanUpFix and be used as an ICleanUpFix
```
akurtakov, line 30: The comment doesn't make sense that now.

robstryker: I had wanted to wait until later, but I'm afraid I'll forget, honestly, so I'll add another commit now.  There's only like 15 references to the class and they are easily removed.

robstryker: Um, oops, seems I made it part of the part-2 commit instead of the part 3.   See https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/1161

akurtakov:
Thanks Rob!

PR #1187, opened 2024-02-12
ChangedValueChecker: multithreading fixes

jukzi:
* use concurrent datastructures, volatile fConflict
* do not spinloop burning cpu


rgrunber:
This pull request may break the eclipse-jdtls/eclipse.jdt.ls project. CC'ing @rgrunber for awareness.

Issue #1192, opened 2024-02-14
Refactor rename of method named 'copy' freezes Eclipse


jukzi:
To reproduce:
- Define a method: `void copy() {}`.
- Rename it through refactoring.

Result:
- Eclipse freezes.

It does not happen when the method has arguments.
It happens in both version 2023-03 and 2023-12.

_Originally posted by @WJSchakel in https://github.com/eclipse-platform/eclipse.platform.ui/issues/1675#issuecomment-1941559886_

some sampling:
![image](https://github.com/eclipse-jdt/eclipse.jdt.ui/assets/51790620/e24a6c57-c59b-45b8-b876-886ab1f18096)

jukzi:
I can reproduce it in some projects, for example in the package org.eclipse.jdt.unittest.junit - but not in a trival project or package org.eclipse.jdt.ui.tests.refactoring.actions.

```
public class A {
	// refactor the name:
	void copy() {}
}
```

jukzi:
stacktrace:

```
Thread [ModalContext] (Suspended (breakpoint at line 686 in org.eclipse.core.runtime.Path))	
	org.eclipse.core.runtime.Path.equals(java.lang.Object) line: 686	
	org.eclipse.jdt.internal.core.JrtPackageFragmentRoot.equals(java.lang.Object) line: 118	
	org.eclipse.jdt.internal.core.JarPackageFragment(org.eclipse.jdt.internal.core.PackageFragment).equals(java.lang.Object) line: 188	
	org.eclipse.jdt.internal.core.ClassFile(org.eclipse.jdt.internal.core.AbstractClassFile).equals(java.lang.Object) line: 132	
	org.eclipse.jdt.internal.core.ResolvedBinaryType(org.eclipse.jdt.internal.core.JavaElement).equals(java.lang.Object) line: 189	
	org.eclipse.jdt.internal.core.ResolvedBinaryType(org.eclipse.jdt.internal.core.SourceRefElement).equals(java.lang.Object) line: 98	
	org.eclipse.jdt.internal.core.ResolvedBinaryType(org.eclipse.jdt.internal.core.BinaryType).equals(java.lang.Object) line: 174	
	java.util.HashMap<K,V>.getNode(java.lang.Object) line: 577	
	java.util.HashMap<K,V>.get(java.lang.Object) line: 556	
[...]

PR #1193, opened 2024-02-14
fix endless loop in RippleMethodFinder2.UnionFind.find(IType) #1192

jukzi:
https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1192

### DIFF
--- a/org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/refactoring/rename/RippleMethodFinder2.java
+++ b/org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/refactoring/rename/RippleMethodFinder2.java
@@ -98,7 +98,8 @@ public class RippleMethodFinder2 {
 		public IType find(IType element) {
 			IType root= element;
 			IType rep= fElementToRepresentative.get(root);
-			while (rep != null && ! rep.equals(root)) {
+			Set<IType> visited=new HashSet<>();
+			while (rep != null && visited.add(root)) {
 				root= rep;
 				rep= fElementToRepresentative.get(root);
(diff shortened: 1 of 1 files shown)


jukzi:
@jjohnstn i have no idea what that method is all about, but this seems to fix an endless loop. just don't know what that method is supposed to return in such case - if it matters at all

org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/refactoring/rename/RippleMethodFinder2.java
```diff
@@ -98,7 +98,8 @@ public void init(IType type) {
 		public IType find(IType element) {
 			IType root= element;
 			IType rep= fElementToRepresentative.get(root);
-			while (rep != null && ! rep.equals(root)) {
+			Set<IType> visited=new HashSet<>();
```
jjohnstn, line 101: I'm not familiar with the code.  The change is fine for the meantime and much better than an infinite loop.  The real error is somehow the fElementToRepresentative map is being formed incorrectly.  Whether designating the first representative found as the root causes other problems or not is a future question.

jjohnstn (approved):

org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/refactoring/rename/RippleMethodFinder2.java
jukzi: > The change is fine for the meantime and much better than an infinite loop

i like that attitude

PR #1203, opened 2024-02-16
Revert workaround for #945, as it causes problems in the JUnit view

trancexpress:
We revert the changes from:

#946
#1022

Since there are other regressions we see, such as showing the string comparison dialog not working.

## Author checklist

- [ ] I have thoroughly tested my changes
- [ ] The change is following the [coding conventions](https://wiki.eclipse.org/Platform/How_to_Contribute#Coding_Conventions)
- [ ] I have signed the [Eclipse Contributor Agreement (ECA)](https://www.eclipse.org/legal/ECA.php)

PR #1217, opened 2024-02-19
Don't show children for JUnit 5 @TestTemplate tests that ran only once

iloveeclipse:
Don't show children for JUnit 5 `@TestTemplate` tests that ran only once

When running JUnit 5 `@TestTemplate` tests, extra test suites are shown in the JUnit view to represent the tests in the template. Those suites introduce unnecessary noise when the template contains only one test - e.g. when repeating tests that failed.

This change adjusts TestSuiteElement.getChildren(), pruning each TestSuiteElement that has a single TestCaseElement child that is also a dynamic test.

```
import static org.junit.jupiter.api.Assertions.assertEquals;

import org.junit.Assume;
import org.junit.jupiter.api.MethodOrderer.MethodName;
import org.junit.jupiter.api.RepeatedTest;
import org.junit.jupiter.api.Test;
import org.junit.jupiter.api.TestMethodOrder;

@TestMethodOrder(MethodName.class)
public class TestTemplatesJunit5 {

	@Test
	public void testOkNoRepeats() throws Exception {
		Thread.sleep(500);
	}
	
	@RepeatedTest(value = 1)
	public void testOkOneRepeat() throws Exception {
		Thread.sleep(500);
	}

	@RepeatedTest(value = 2)
	public void testOkTwoRepeats() throws Exception {
		Thread.sleep(500);
	}
	
	@Test
	public void testWithErrorNoRepeats() throws Exception {
		doStupidThings(null);
	}
	
	@RepeatedTest(value = 1)
	public void testErrorOneRepeat() throws Exception {
		doStupidThings(null);
	}
	
	@RepeatedTest(value = 2)
	public void testErrorTwoRepeats() throws Exception {
		doStupidThings(null);
	}

	@Test
	public void testFailNoRepeats() throws Exception {
		assertEquals("expected", "observed");
	}
	
	@RepeatedTest(value = 1)
	public void testFailOneRepeat() throws Exception {
		assertEquals("expected", "observed");
	}

	@RepeatedTest(value = 2)
	public void testFailTwoRepeats() throws Exception {
		assertEquals("expected", "observed");
	}


	@RepeatedTest(value = 0)
	public void testWrongRepeat() throws Exception {
		Thread.sleep(500);
	}
	
	@Test
	public void testAssume() throws Exception {
		Assume.assumeFalse(true);
	}
	
	@RepeatedTest(value = 1)
	public void testAssumeOneRepeat() throws Exception {
		Assume.assumeFalse(true);
	}
	
	@RepeatedTest(value = 2)
	public void testAssumeTwoRepeats() throws Exception {
		Assume.assumeFalse(true);
	}
	
	private void doStupidThings(String arg) {
		arg.toString();
	}
	
}
```

Fixes https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/945

### DIFF
--- a/org.eclipse.jdt.junit.core/src/org/eclipse/jdt/internal/junit/model/TestSuiteElement.java
+++ b/org.eclipse.jdt.junit.core/src/org/eclipse/jdt/internal/junit/model/TestSuiteElement.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2017 IBM Corporation and others.
+ * Copyright (c) 2000, 2024 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -33,6 +33,10 @@ public class TestSuiteElement extends TestElement implements ITestSuiteElement {
 
 	@Override
 	public Result getTestResult(boolean includeChildren) {
+		TestCaseElement child= getSingleDynamicChild();
+		if (child != null) {
+			return child.getStatus().convertToResult();
+		}
 		if (includeChildren) {
 			return getStatus().convertToResult();
 		} else {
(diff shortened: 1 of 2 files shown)


trancexpress:
@iloveeclipse I would rather not have changes like this all over the view and model code. Better to revert and we'll adapt our use case to not have the problem in the first place.

iloveeclipse:
After merging revert in https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/1203 for RC1 the plan is to "revert the revert" & add this patch on top for 4.31 M1.

PR #1282, opened 2024-03-24
Add support to filter chain by passing the in the token

gayanper:
## What it does
Add a new constructor to pass in the token so that the token will be used when matching the chains. This helps to better control how much chains are produced by using the text that is already typed in an editor as the token. This helps to provide better chains selection without increasing the number of chains to produce for JDT.LS. The existing constructor will preserve the current behavior uneffected.

## How to test
This new behavior can be only tested with JDT.LS PR https://github.com/eclipse-jdtls/eclipse.jdt.ls/pull/2835

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.manipulation/common/org/eclipse/jdt/internal/ui/text/ChainFinder.java
+++ b/org.eclipse.jdt.core.manipulation/common/org/eclipse/jdt/internal/ui/text/ChainFinder.java
@@ -19,7 +19,7 @@ import java.util.Map;
 
 import org.eclipse.jdt.core.IJavaElement;
 import org.eclipse.jdt.core.IType;
-
+import org.eclipse.jdt.core.compiler.CharOperation;
 import org.eclipse.jdt.internal.ui.text.ChainElement.ElementType;
 
 public class ChainFinder {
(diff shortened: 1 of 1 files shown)


gayanper:
@rgrunber can you review this and approve ?

gayanper:
@jjohnstn @rgrunber can one of you give me a go ahead for this. This is to actually improve the jdtls chain completion functionality

jjohnstn (approved):
I have no issues with the changes, but I don't work with jdt.ls so have no way of verifying this; therefore I'll have to defer to @rgrunber to at least test and report back.  I have moved to 4.32 M3 as 4.32 M2 has sailed.  Is this something that JDT UI users would also want to utilize in the future?

gayanper:
> I have no issues with the changes, but I don't work with jdt.ls so have no way of verifying this; therefore I'll have to defer to @rgrunber to at least test and report back. 

We tried the out in the following MR in jdt.ls https://github.com/eclipse-jdtls/eclipse.jdt.ls/pull/2835. But I will wait for @rgrunber 


> I have moved to 4.32 M3 as 4.32 M2 has sailed. Is this something that JDT UI users would also want to utilize in the future?

Yes I think JDT UI can use this as well if it is a good idea.

rgrunber (approved):
Looks fine to me. I'm fine with merging for M3. If it's an improvement, we can write testcases once/if JDT UI adopts it.

PR #1283, opened 2024-03-25
Add fAddFinal to insert 'final' modifier to some proposals.

rgrunber:
- "Create local variable/field" & "Assign to new local variable/field"

Based on a request from https://github.com/eclipse-jdtls/eclipse.jdt.ls/pull/2371 .


rgrunber:
The api tools errors/warnings don't look at all related to this.

PR #1074, opened 2024-01-15
Add styling enhancements to signatures (JavaElementLinks) in Javadoc. Fixes #1073

RedeemerSK:
## What it does
Adds various enhancements to styling / formatting of signatures in Javadoc presentation (HTML-based) views:
- togglable text wrapping (method parameters, generics type parameters used in method)
- togglable text formatting (generics type parameters used, method parameter names, method name, method class)
- togglable generic type parameters references text coloring
- togglable generic type parameters nesting levels text coloring
- togglable enhancement states: OFF / always ON / ON only on mouse hover over new styling button
- toggable enhancements state preferences saved separately for Javadoc hover view and Javadoc view, color preferences are shared
- styling button also acts as a menu where all aspects of enhancements can be configured : togglable states, color preferences
- defines 4 default colors for type parameters references / levels coloring, further references / levels re-use them repratedly by default
- supports Dark mode (theme) by defining separate default colors and gracefully handling switching themes

Examples:
Default enhancements:
javadoc_new
When hovered on styling button
javadoc_new_colored

Defaults for colors:
lightMode-colors-withFormatting
darkMode-colors-withFormatting

Changing settings example:
preview_overlay

colorPreferencesSubmenu-6colors
colorPreferencesSubmenu-0colors

See [Eclipse JDT forums topic](https://www.eclipse.org/forums/index.php?t=msg&th=1114002&goto=1862222&#msg_1862222) I created to communicate goals, approach, decisions taken and progress.

## How to test
At this stage of WIP prototype:
Have Javadoc for various Java source code elements displayed in both Javadoc hover (viewer) and Javadoc view. Interact with new styling menu button (added to toolbars) to form assumptions about what new visual items do & see if things then behave as one expects them to behave. 

## Author checklist

- [x] I have thoroughly tested my changes
- [x] The change is following the [coding conventions](https://wiki.eclipse.org/Platform/How_to_Contribute#Coding_Conventions)
- [x] I have signed the [Eclipse Contributor Agreement (ECA)](https://www.eclipse.org/legal/ECA.php)

Changes are currently WIP and thus at this stage not meant for thorough code-review, but rather for other people to be able to try them and provide feedback & future code-reviewer to have a look at how changes are  implemented.

PR #1295, opened 2024-03-29
Do not log locking CSS fragments cache for read was interrupted #1294

RedeemerSK:
## What it does
Removes unnecessary log message when interrupted while JavaElementLinks code was waiting for CSS fragments lock (performed for purposes of applying new Javadoc styling enhancements #1074 ).

## How to test
I was not able to reproduce the reported issue so I can't give exact steps/flow to test the fix. Issue description is however right in saying:
> while editing or navigating inside Java file, mouse hover request may be cancelled because of some user actions that make the request obsolete.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/viewsupport/JavaElementLinks.java
+++ b/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/viewsupport/JavaElementLinks.java
@@ -562,8 +562,8 @@ public class JavaElementLinks {
 						CSS_FRAGMENTS_CACHE_LOCK.unlock();
 					}
 				}
-			} catch (InterruptedException e1) {
-				JavaPlugin.logErrorMessage("Interrupted while waiting for CSS fragments cache lock, cache reset unsuccessful"); //$NON-NLS-1$
+			} catch (InterruptedException e) {
+				JavaPlugin.log(new RuntimeException("Interrupted while waiting for CSS fragments cache lock, cache reset unsuccessful", e)); //$NON-NLS-1$
 			}
 		}
 	}
@@ -866,7 +866,8 @@ public class JavaElementLinks {
 		try {
 			locked= CSS_FRAGMENTS_CACHE_LOCK.tryLock(100, TimeUnit.MILLISECONDS);
 		} catch (InterruptedException e) {
-			JavaPlugin.logErrorMessage("Interrupted while waiting for CSS fragments cache lock, proceeding without using cache"); //$NON-NLS-1$
+			Thread.currentThread().interrupt();
+			return css;
 		}
 		try {
 			if (locked) {


iloveeclipse:
I haven't analyzed the code yet - could you explain what this lock is doing first?

RedeemerSK:
> I haven't analyzed the code yet - could you explain what this lock is doing first?

To apply coloring to the type parameters (generics) references inside Javadoc hover / viewer based on current color preferences (added with #1074 ) , CSS that is inserted into Javadoc HTML (being built to be displayed) has to be enriched with CSS rules for each color used inside Javadoc HTML. Since texts for these color rules are constants (until user changes or resets above mentioned color preferences) these CSS color rules texts = fragments are cached.

Access to this cache is controlled by lock, since it may be the case that HTML needs to use more colors than what is currently cached and that's when cache needs to be expanded with new entries for additional colors.

iloveeclipse:
Does it need to be a lock at all? Can't the cache be updated "lock free"?

All what ai did today was using a nightly IDE build for usual work, added new code, refactoring old code a bit, diffing changes compared to head etc. After 2-3 hours I've seen ~ 5 errors in the log. Assuming not every cache change created an error, there must be a lot of changes in the cache. Why that?

What I would also assume based on my previous experience with hovers (sorry, still not analyzed your new code) is that there is a thread started to compute hover for one method or type, and while it was running, user already moved mouse so the hover data isn't needed at all and new one is requested. From the nature of UI requests, the UI can only request hover from a single thread, and there shouldn't be **multiple** threads calculating hovers at all. So I don't understand why a lock is needed. 

Instead of locking to guard some mutable data, I would assume the cached data should be immutable, and if update is needed, it could be cloned, updated and set after update (if not interrupted), but not modified **in place**. 

So if hover thread is interrupted because next hover will be requested, nothing should change at all, no locking, no exceptions.

RedeemerSK:
The cache is very rarely modified - only when more colors than what has been used until that moment need to be used ... so in it's modified somewhere in between once and like (I guess) 3 times during lifetime of the IDE and that's it.

https://github.com/eclipse-jdt/eclipse.jdt.ui/blob/master/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/infoviews/AbstractInfoView.java#L711 is the guy doing Javadoc content preparation for Javadoc view. Definitely not an UI (SWT) thread and not a singleton thread either.

https://github.com/eclipse-platform/eclipse.platform.ui/blob/master/bundles/org.eclipse.jface.text/src/org/eclipse/jface/text/TextViewerHoverManager.java#L145 is one doing Javadoc content preparation for Javadoc hover. This is most probably singleton.

Having breakpoint in method where cache is used I can easily reproduce having 2+ non UI threads arriving there at the same time.

That's why I guarded cache with locking. And I made it plain stupid - just a simple String[] modified in place. I still think, considering how often it is modified (basically never), it's "good-enough" solution.

> So if hover thread is interrupted because next hover will be requested, nothing should change at all, no locking, no exceptions.

That's the state after my fix. If thread is interrupted, it just returns from method right away without any enrichments, since the whole Javadoc being prepared is thrown away anyway.

PR #1325, opened 2024-04-14
Make remaining JavaPlugin lazy init fields thread safe

iloveeclipse:
Initialization of some of the fields in JavaPlugin was not thread safe.

This change tries to add thread safe init without holding extra locks in critical paths to prevent possible deadlocks while loading 3rd party code.

This may result in possible double creation of some objects but should avoid errors on multiple threads initializing same JavaPlugin fields in parallel.

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2276


iloveeclipse:
Quality gate issues are either false positive or known PDE issue https://github.com/eclipse-pde/eclipse.pde/issues/1228

PR #1375, opened 2024-04-27
Also handle RuntimeException when performing ClasspathFixProposal

laeubi:
Currently when a RuntimeException occurs inside ClasspathFixProposal an obscure error dialog as the wizard already handles errors if they are InvocationTargetException this now wraps RuntimeException as well.


org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/wizards/buildpaths/ClasspathFixSelectionDialog.java
```diff
@@ -109,6 +109,8 @@ public static boolean openClasspathFixSelectionDialog(Shell parent, final IJavaP
 						throw new InterruptedException();
 					} catch (CoreException e2) {
 						throw new InvocationTargetException(e2);
+					} catch(RuntimeException e3) {
```
iloveeclipse, line 112: Makes sense, but could we combine this with the upper catch block?

PR #1381, opened 2024-05-01
Fix inline method to check access modifiers of inlined code

### LINKED ISSUE #1358: Inline method refactoring leads to syntax errors
Eclipse version：2023-9

Repeat the steps：

1. select “staticMethod()”
2 . click “Refactor-inlinemethod”

class MyClass {
    private static int value = 10;
    public static void staticMethod() {
        System.out.println("Value: " + value);
    }
}

class AnotherClass {
    private int value = 20;
    public void callStaticMethod() {
        new MyClass().staticMethod(); 
    }
}

Refactoring result:

image

jjohnstn:
- add new checkAccessCompatible() method to SourceProvider that checks if code from declaration can be moved to target without causing an access violation
- add new AccessVisitor to SourceAnalyzer to log non-public accesses (private, package-private, and protected)
- add call to checkAccessCompatible() in CallInliner
- add new tests to InlineMethodTests
- fixes #1358

## What it does

Fixes inline method to consider the various access modifiers of members being accessed by the code to be inlined and to fail if there would be access violations from performing the inline.

## How to test

See issue and new tests.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/dom/ASTNodes.java
+++ b/org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/dom/ASTNodes.java
@@ -2526,15 +2526,15 @@ public class ASTNodes {
 	 * @param node the node
 	 * @return the top-level AbstractTypeDeclaration node for the file that node is in
 	 */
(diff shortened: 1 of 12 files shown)


jjohnstn:
Test failure is known unstable test and nothing to do with this change.

PR #1351, opened 2024-04-23
Javadoc sections misaligned

iloveeclipse:
Regression from cbc03c25f6059c774d8766bd7b02489c9df53acf / https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/941 

The code moved from JavadocContentAccess2 to CoreJavadocAccessImpl changed dt/dd tags to ul/li tag pairs, and that broke javadoc alignment.

This commit restores original tag pairs used for javadoc formatting.

Fixes https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1346

PR #1386, opened 2024-05-01
Make CoreJavadocAccessImpl more extendable for jdt.ls

robstryker:
## What it does

https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/1351 seems to have fixed an inadvertent change to the value of some constants in a file that was being abstracted out for subclassing. 

It's clear that the abstraction process was incomplete and only appeared complete because of these changed constants. The reversion of those constants to their historical value has revealed that jdt.ls cannot override the behavior as expected. 

In an attempt to allow jdt.ls to override these values, this PR will make getters for the 6 values, which subclasses may override. All references to those constants in this file can be changed to make use of the getters instead. 

This should allow jdt.ls more flexibility in overriding these values and using different formatting as they require. 

## How to test

This commit should not change current behavior. The test as performed in https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/1351 should be sufficient to see that no changes in behavior are observed. 

## Author checklist


robstryker:
@iloveeclipse Hi Andrey: Your feedback on this issue is welcome / appreciated.

iloveeclipse:
> Hi Andrey: Your feedback on this issue is welcome / appreciated.

Looks OK, please just make sure search/replace hit the right places :)

robstryker:
If I comment out the declaration of the constants, the only errors are in those getters, so, looks like I didn't miss any uses. 

Verified line by line that the proper replacements were made, and that the getters return the correct constant.

iloveeclipse (approved):

Issue #1414, opened 2024-05-21
Pasting snippet to Package Explorer does not work anymore


iloveeclipse:
Just stumbled over this bug: copy text below into clipboard
```
	public static void main(String[] args) throws Exception {
		while(true) {
			Thread.sleep(500);
		}
	}
```
and try to paste it (`Ctrl+V`) into Package Explorer.

- In 4.31 a new Snippet.java class is created with this snippet.
- In 4.32 nothing happens, except an error is reported to the error log:
```
eclipse.buildId=4.32.0.I20240520-1800
java.version=21.0.2-13
java.vendor=N/A
BootLoader constants: OS=linux, ARCH=x86_64, WS=gtk, NL=en_US
Command-line arguments:  -os linux -ws gtk -arch x86_64

org.eclipse.e4.ui.workbench
Error
Tue May 21 17:46:48 CEST 2024
Execution exception for: ParameterizedCommand(Command(org.eclipse.ui.edit.paste,Paste,
		Paste from the clipboard,
		Category(org.eclipse.ui.category.edit,Edit,null,true),
		WorkbenchHandlerServiceHandler("org.eclipse.ui.edit.paste"),
		,,true),null) in 
	context chain: WorkbenchContext -> TrimmedWindowImpl (IDEWindow) Context -> PerspectiveImpl (org.eclipse.jdt.ui.JavaPerspective) Context -> PartImpl (org.eclipse.jdt.ui.PackageExplorer) Context

org.eclipse.core.commands.ExecutionException: While executing the action, an exception occurred
	at org.eclipse.jface.commands.ActionHandler.execute(ActionHandler.java:129)
	at org.eclipse.ui.internal.handlers.E4HandlerProxy.execute(E4HandlerProxy.java:98)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
	at org.eclipse.e4.core.internal.di.MethodRequestor.execute(MethodRequestor.java:58)
	at org.eclipse.e4.core.internal.di.InjectorImpl.invokeUsingClass(InjectorImpl.java:299)
	at org.eclipse.e4.core.internal.di.InjectorImpl.invoke(InjectorImpl.java:233)
	at org.eclipse.e4.core.contexts.ContextInjectionFactory.invoke(ContextInjectionFactory.java:174)
[...]

PR #1416, opened 2024-05-22
Fix for pasting snippet with main method to Package Explorer

iloveeclipse:
After introducing implicit types via
https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2066 the snippet that contains only main() method is recognized as
ImplicitTypeDeclaration that does not inherit AbstractTypeDeclaration but common super class AbstractUnnamedTypeDeclaration.

Make sure we don't fail with CCE and don't generate extra main method that wraps around pasted main method.

Fixes https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1414

### DIFF
--- a/org.eclipse.jdt.ui/ui refactoring/org/eclipse/jdt/internal/ui/refactoring/reorg/PasteAction.java
+++ b/org.eclipse.jdt.ui/ui refactoring/org/eclipse/jdt/internal/ui/refactoring/reorg/PasteAction.java
@@ -108,6 +108,7 @@ import org.eclipse.jdt.core.dom.AST;
 import org.eclipse.jdt.core.dom.ASTNode;
 import org.eclipse.jdt.core.dom.ASTParser;
 import org.eclipse.jdt.core.dom.AbstractTypeDeclaration;
+import org.eclipse.jdt.core.dom.AbstractUnnamedTypeDeclaration;
 import org.eclipse.jdt.core.dom.Block;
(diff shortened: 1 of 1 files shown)


org.eclipse.jdt.ui/ui refactoring/org/eclipse/jdt/internal/ui/refactoring/reorg/PasteAction.java
```diff
@@ -428,18 +430,19 @@ private static List<ParsedCu> parseAsTypes(String text, CompilationUnit unit) {
...
+				List<AbstractUnnamedTypeDeclaration> types= unit.types();
 
 				int startOffset= 0;
 				String typeName= null;
 				int maxVisibility= JdtFlags.VISIBILITY_CODE_INVALID;
 
 				// Public types must be in separate CUs:
-				for (AbstractTypeDeclaration type : types) {
+				for (AbstractUnnamedTypeDeclaration type : types) {
+					Supplier<String> name = () -> type instanceof AbstractTypeDeclaration named? named.getName().getIdentifier() : ""; //$NON-NLS-1$
```
trancexpress, line 441: I would prefer a static method to a lambda here.

trancexpress:
Seems to work. Personally I don't miss randomly pasting code into the Package Explorer (its always a mistake when I do this, I forgot to select a package to paste into), but I guess the exception is bad and should be fixed.

trancexpress (approved):

PR #1433, opened 2024-05-31
Add UI to select if JUnit3/4 should be enabled for JUnit Platform

laeubi:
Currently one needs to edit the classpath directly what is a bit inconvenient.

This adds a new checkbox displayed in case Junit 5 is used that allows the user to choose if Junit3/4 compatibility layer (aka vintage engine) should be used.

Contributes to:
- https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/837

This is how the dialog looks like:
grafik


carstenartur:
Keep in mind that currently the dependency to junit 3 is hardcoded in jdt ui. Without the junit 3 class "junit.framework.Test" in the classpath some extentions are not activated.
See https://bugs.eclipse.org/bugs/show_bug.cgi?id=570024

Lookup in plugin.xml of org.eclipse.jdt.junit

laeubi:
@carstenartur would you be fine if I remove the JUnit3 from the Text? It feels not so relevant for me in this case as likley no one will write *new* code using JUnit3 using JUnit5 vintage engine.

carstenartur:
> @carstenartur would you be fine if I remove the JUnit3 from the Text? It feels not so relevant for me in this case as likley no one will write *new* code using JUnit3 using JUnit5 vintage engine.

I personally feel the same regarding Junit3, but there is jdt.core where Junit3 still is standard 😅

You better ask @noopur2507 or @jjohnstn. I'm not a commiter and they are and know much more than me about the implementation of the junit part in jdt.ui and the plan foreward.

akurtakov:
Even if one writes JUnit 3 *style* tests they will be executed by JUnit 4 library which contains a copy of what JUnit 3.x used to be. JUnit 3 itself is not shipped by eclipse for 12 years https://bugs.eclipse.org/bugs/show_bug.cgi?id=391433 .
With all of the above I believe it's time to let JUnit 3 RIP.

laeubi:
Just to prevent confusion, I don't want to remove JUnit 3 from JDT!

I just was about to ask, if the UI would be fine if we remove the `3` from the String "Enable JUnit 3/4 support", so people don't think it will work fine with JUnit3 (for whatever reasons).

laeubi:
@jjohnstn can you help out here?

JUnit 6 has even [deprecated the vintage engine](https://docs.junit.org/6.0.0-M1/release-notes/#release-notes-6.0.0-M1-junit-vintage)

> The JUnit Vintage engine is now deprecated and will report an INFO level discovery issue when it finds at least one JUnit 4 test class. For now, the intent of the deprecation is not to signal removal in the next major version but to clarify the intended purpose of the engine. It should only be used temporarily while migrating tests to JUnit Jupiter or another testing framework with native JUnit Platform support.

So I think it is really time that JDT allows users to disable that part in the JUnit Container ...

carstenartur:
Maybe time to think again about if a contribution of a junit update migration cleanup could be supported? I think it could be the reason for some users to gibe eclipse another chance if it is useful.
https://github.com/carstenartur/sandbox?tab=readme-ov-file#8-sandbox_junit

Issue #736, opened 2023-08-24
AnnotateAssistTest1d8 test failed 


noopur2507:
AnnotateAssistTest1d8.testAnnotateParameter_Varargs2 test failed on mac in I20230824-0400.

```java
java.lang.NullPointerException
at java.base/java.util.Objects.requireNonNull(Objects.java:208)
at java.base/java.util.Arrays$ArrayList.<init>(Arrays.java:4137)
at java.base/java.util.Arrays.asList(Arrays.java:4122)
at org.eclipse.jdt.ui.tests.quickfix.AbstractAnnotateAssistTests.collectAnnotateProposals(AbstractAnnotateAssistTests.java:82)
at org.eclipse.jdt.ui.tests.quickfix.AnnotateAssistTest1d8.testAnnotateParameter_Varargs2(AnnotateAssistTest1d8.java:895)
at java.base/java.lang.reflect.Method.invoke(Method.java:568)
```

noopur2507:
org.eclipse.jdt.ui.tests.quickfix.AnnotateAssistTest1d8.testAnnotateParameter_WildcardBound failed in I20231115-1800, macosx.cocoa.x86_64 java17.

```
java.lang.NullPointerException
at java.base/java.util.Objects.requireNonNull(Objects.java:208)
at java.base/java.util.Arrays$ArrayList.<init>(Arrays.java:4137)
at java.base/java.util.Arrays.asList(Arrays.java:4122)
at org.eclipse.jdt.ui.tests.quickfix.AbstractAnnotateAssistTests.collectAnnotateProposals(AbstractAnnotateAssistTests.java:82)
at org.eclipse.jdt.ui.tests.quickfix.AnnotateAssistTest1d8.testAnnotateParameter_WildcardBound(AnnotateAssistTest1d8.java:303)
at java.base/java.lang.reflect.Method.invoke(Method.java:568)
```

jukzi:
AnnotateAssistTest1d8 frequentl fails on I-Build I20240123-1800 on all platforms with various exceptions for example NPE or "Wrong number of proposals"
https://download.eclipse.org/eclipse/downloads/drops4/I20240123-1800/testresults/html/org.eclipse.jdt.ui.tests_ep431I-unit-mac64-java17_macosx.cocoa.x86_64_17.html

jukzi:
also frequently fails on CI jenkins

noopur2507:
[...]

PR #1, opened 2022-04-08
https://github.com/eclipse-platform/eclipse.platform.releng.aggregato…

ktatavarthi:
…r/issues/201 adding JenkinsFile

Signed-off-by: Kalyan Prasad Tatavarthi <kalyan_prasad@in.ibm.com>

PR #1478, opened 2024-06-26
AnnotateAssistTest1d8 test failed

stephan-herrmann:
addresses the ClassCastException

fixes https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/736

### DIFF
--- a/org.eclipse.jdt.ui.tests/ui/org/eclipse/jdt/ui/tests/quickfix/AbstractAnnotateAssistTests.java
+++ b/org.eclipse.jdt.ui.tests/ui/org/eclipse/jdt/ui/tests/quickfix/AbstractAnnotateAssistTests.java
@@ -35,6 +35,8 @@ import org.eclipse.core.resources.IFolder;
 import org.eclipse.core.resources.IProject;
 import org.eclipse.core.resources.IResource;
 
+import org.eclipse.jface.preference.IPreferenceStore;
+
 import org.eclipse.jface.text.TextSelection;
 import org.eclipse.jface.text.contentassist.ICompletionProposal;
 import org.eclipse.jface.text.quickassist.IQuickAssistInvocationContext;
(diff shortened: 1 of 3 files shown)


stephan-herrmann:
@iloveeclipse @jjohnstn  From 3 attempts I got 2 timeouts (and one unrelated intermittent failure).

Who decides about the timeout? It seems that the current limit of 40 min. is not realistic.

iloveeclipse:
> Who decides about the timeout? It seems that the current limit of 40 min. is not realistic.

See 
https://github.com/eclipse-jdt/eclipse.jdt.ui/blob/690340fdd53a50e5050c77a423dcc857c2b6f7ba/Jenkinsfile#L3

stephan-herrmann:
> > Who decides about the timeout? It seems that the current limit of 40 min. is not realistic.
> 
> See
> 
> https://github.com/eclipse-jdt/eclipse.jdt.ui/blob/690340fdd53a50e5050c77a423dcc857c2b6f7ba/Jenkinsfile#L3

Thanks, so the timeout has been at 40 since @ktatavarthi created that file 2 years ago (https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/1)

Asked differently: does everybody think this is the right threshold?

akurtakov:
Looking at https://ci.eclipse.org/jdt/job/eclipse.jdt.ui-github/job/master/buildTimeTrend - normal build is usually around ~25 minutes. What we see is probably related to https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/4063#note_2472456 . Meanwhile it should be fine to bump it to e.g. 60 minutes (twice slower than average time) so people don't have to restart builds like that.

stephan-herrmann:
> Looking at https://ci.eclipse.org/jdt/job/eclipse.jdt.ui-github/job/master/buildTimeTrend - normal build is usually around ~25 minutes. What we see is probably related to https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/4063#note_2472456 . Meanwhile it should be fine to bump it to e.g. 60 minutes (twice slower than average time) so people don't have to restart builds like that.

Thanks, will do if/when I run into this problem again in the near future :)

Issue #1443, opened 2024-06-10
ClassFileEditorInputFactory: "Its not allow to modify JavaModel during ReadOnly action. delta=Java Model[*]: {CHILDREN}"


jukzi:
Happend when i started eclipse today:

```
eclipse.buildId=4.32.0.I20240529-1800
java.version=21.0.2
java.vendor=Eclipse Adoptium
BootLoader constants: OS=win32, ARCH=x86_64, WS=win32, NL=de_DE
Command-line arguments:  -os win32 -ws win32 -arch x86_64

org.eclipse.ui
Error
Mon Jun 10 07:37:20 CEST 2024
Its not allow to modify JavaModel during ReadOnly action. delta=Java Model[*]: {CHILDREN}
	org.eclipse.jdt.core[*]: {CHILDREN | RESOLVED CLASSPATH CHANGED}
		C:\Users\jkubitz\.p2\pool\plugins\org.osgi.annotation.versioning_1.1.2.202109301733.jar[*]: {REMOVED FROM CLASSPATH}
		C:\Users\jkubitz\.p2\pool\plugins\org.osgi.annotation.bundle_2.0.0.202202082230.jar[*]: {REMOVED FROM CLASSPATH}
		C:\Users\jkubitz\.p2\pool\plugins\org.osgi.service.component.annotations_1.5.1.202212101352.jar[*]: {REMOVED FROM CLASSPATH}
		C:\Users\jkubitz\.p2\pool\plugins\org.osgi.service.metatype.annotations_1.4.1.202109301733.jar[*]: {REMOVED FROM CLASSPATH}
		C:\Users\jkubitz\.p2\pool\plugins\org.osgi.service.prefs_1.1.2.202109301733.jar[*]: {REMOVED FROM CLASSPATH}
		C:\Users\jkubitz\.p2\pool\plugins\org.apache.ant_1.10.14.v20230922-1200\lib\ant.jar[*]: {REMOVED FROM CLASSPATH}
		antadapter[*]: {REORDERED}
		codeassist[*]: {REORDERED}
		dom[*]: {REORDERED}
		eval[*]: {REORDERED}
		formatter[*]: {REORDERED}
		model[*]: {REORDERED}
		search[*]: {REORDERED}
		C:\Users\jkubitz\platform-2023-05-29\eclipse\plugins\org.osgi.annotation.versioning_1.1.2.202109301733.jar[*]: {ADDED TO CLASSPATH}
[...]

PR #1483, opened 2024-07-01
add Test for "Show Tooltip Description" #1443

jukzi:
https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1443


jukzi:
should have caused a failing test, as local :.-(

jukzi:
i was not able to write a test that fails when other jd.tui tests have run before. The tests fails only when its the first test run.

PR #1486, opened 2024-07-02
AnnotateAssistTest1d8 test failed

stephan-herrmann:
explicitly request a ClassfileEditor

2nd attempt at https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/736


stephan-herrmann:
Failure captured for posterity:
```
Creating placeholder flownodes because failed loading originals.
ERROR: Cannot resume build because FlowNode 36 for FlowHead 1 could not be loaded. This is expected to happen when using the PERFORMANCE_OPTIMIZED durability setting and Jenkins is not shut down cleanly. Consider investigating to understand if Jenkins was not shut down cleanly or switching to the MAX_SURVIVABILITY durability setting which should prevent this issue in most cases.
```
???

iloveeclipse:
> ERROR: Cannot resume build because FlowNode 36

For whatever reason our Jenkins instance restarted today, you are not alone with strange Jenkins errors :-) But it seem to run now, so you could try to retrigger the build.

PR #1487, opened 2024-07-02
make "AllTests" more specific

jukzi:
Avoid duplicate class names to make it easier to find the right one

### DIFF
--- a/org.eclipse.jdt.ui.tests.refactoring/pom.xml
+++ b/org.eclipse.jdt.ui.tests.refactoring/pom.xml
@@ -23,7 +23,7 @@
   <packaging>eclipse-test-plugin</packaging>
   <properties>
     <testSuite>${project.artifactId}</testSuite>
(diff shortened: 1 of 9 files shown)


jjohnstn (commented):
I don't have a strong opinion on this unless if affects reporting (do the checks account for tests being renamed?).  I notice that jdt.ui.tests use TestSuite in the name.  This is done here for the NLSTestSuite, so, for example, maybe using CCPTestSuite vs AllRefactoringCCPTests might be more compact and to the point and be more consistent with jdt.ui.tests.

akurtakov:
Is this still of interest for anyone?

jukzi:
> maybe using CCPTestSuite vs AllRefactoringCCPTests might be more compact and to the point and be more consistent with jdt.ui.tests.

updated to using "TestSuite" suffix

akurtakov (approved):
LGTM

PR #1514, opened 2024-07-16
String support for if/else to switch cleanup

carstenartur:
## What it does
Since Java 7 there is support for "switch" statement making use of "String" objects. The cleanup for that in eclipse jdt ui so far only supports primitive types like char, int and so on. 
I just want to see what is needed to add support for the cleanup to refactor if/else chains with String constant comparisons to switch statements as well.
It does not check the projects java version. Not sure if that makes sense any longer. Do we support java 6 today?
In an ideal world there should be at least to tests in a java 6 test configuration and a java 7 test configuration. The currently used "CleanUpTest12" seems to be somehow random.

## How to test

There is a single junit test so far. You can add testcases or use e.g. the jdt codebase for testing.

## Author checklist


jjohnstn:
Thanks @carstenartur .  We just removed support for < Java 1.8 so JVM check is not needed.  Not sure about the history of why all of this is in CleanUpTest12, but it doesn't matter.

org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/fix/SwitchFixCore.java
```diff
@@ -263,11 +265,30 @@ private Variable extractVariableAndValues(final Expression expression) {
 
 				if (infixExpression != null) {
 					return extractVariableAndValuesFromInfixExpression(infixExpression);
+				} else if (expression instanceof MethodInvocation) {
+//					if (!JavaModelUtil.is1d7OrHigher((IJavaProject)null))
+//						return null;
+					MethodInvocation method= (MethodInvocation)expression;
+					if (method.resolveMethodBinding() != null) {
+						if(!"equals".equals(method.getName().getIdentifier())) return null; //$NON-NLS-1$
```
jjohnstn, line 273: Minor nit.  Please stick with if-space-(

org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/fix/SwitchFixCore.java
```diff
@@ -263,11 +265,30 @@ private Variable extractVariableAndValues(final Expression expression) {
...
+						List<?> arguments= method.arguments();
+						if(arguments.size()!=1) return null;
```
jjohnstn, line 275: same comment about if statement.  Please also add spaces around !=

org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/fix/SwitchFixCore.java
```diff
@@ -263,11 +265,30 @@ private Variable extractVariableAndValues(final Expression expression) {
 
 				if (infixExpression != null) {
 					return extractVariableAndValuesFromInfixExpression(infixExpression);
+				} else if (expression instanceof MethodInvocation) {
+//					if (!JavaModelUtil.is1d7OrHigher((IJavaProject)null))
```
jjohnstn, line 269: Remove these two commented lines.

jjohnstn (changes requested):
Looks good.  Just some minor nits.  I will merge when done.  BTW: I have marked this as Noteworthy.  Are you interested in adding a N&N entry or would you like me to do it?

carstenartur:
Hi Jeff,
please go ahead and create a N&N entry if you feel like doing it.
Btw, should support for "enum" be added in this cleanup one day? Maybe it is pointless because everybody prefers "switch" beforehand when using enum anyway and so there is are no cases to cleanup but I'm not sure.

carstenartur:
I do not see how to check for an "enum". The approach used in the other cases does not work.

jjohnstn (approved):

jjohnstn:
Hi @carstenartur Thanks for the changes.  Enum support would be fine but as you mentioned, it is likely a user would have used a switch to begin with.  To find an enum, the ITypeBinding has an isEnum() method.  For an enum value, doing a resolveBinding() of the name will give an IVariableBinding which has isEnumConstant() method.

jjohnstn:
@carstenartur I'll do the N&N for 4.33 M2.

PR #1522, opened 2024-07-19
Fix rename local to avoid field collision

### LINKED ISSUE #1508: Bug: Renaming local variables causes code behavior to change
When renaming 'variable' to 'newField', Line 11 causes the variable assignment field statement to change to the field assignment itself：
image

Refactoring result:
image

jjohnstn:
- add logic to RenameLocalVariableProcessor to detect when a new name will cause an unexpected reference to a field with same name
- add new test to RenameTempTests
- fixes #1508

## What it does

Fixes rename local to avoid a rename that will cause an unexpected reference to an existing field with same name.

## How to test

See issue or new test.

## Author checklist


jjohnstn:
Test failure is a known inconsistent test.

PR #1536, opened 2024-07-23
dogfooding - Convert String concatenation to Text Block

carstenartur:
This is a run of the "Convert String concatenation to Text Block" cleanup without the option for "Stringbuffer" on jdt.ui (only runtime modules, no test code). With the "Stringbuffer" option we run into some errors unfortunately, there is a bug.
Since the JDT UI modules now are higher than Java 15 we can make use of it.
## What it does
By using text blocks  the code gets more readable and the formating characters like return get translated.

## How to test
Nothing should be changed because of this. Tests should run through as before and everywhere you should see Strings in the same way as before. Just check an arbitrary text block where it appears in the resulting ice.

## Author checklist

- [x ] I have signed the [Eclipse Contributor Agreement (ECA)](https://www.eclipse.org/legal/ECA.php)


jjohnstn:
Looks good.

Issue #1525, opened 2024-07-21
cleanup "use local variable type interference" refactors generics and removes needed import


carstenartur:
eclipse.buildId=4.33.0.20240711-1200
java.version=22.0.1
java.vendor=Eclipse Adoptium
BootLoader constants: OS=macosx, ARCH=x86_64, WS=cocoa, NL=de_DE

Code like this:

```java
HashSet<IJavaProject> selectedElements= new HashSet<>();
```

is refactored to 

```java
var selectedElements= new HashSet<IJavaProject>();
```

Unfortunately the resulting code has the error "IJavaProject cannot be resolved to a type" because the import is removed.

carstenartur:
Can it be that the method org.eclipse.jdt.internal.corext.refactoring.structure.ImportRemover.divideTypeRefs() simply is not smart enough to handle parameterized types?

jjohnstn:
Closed with #1538 merged.

PR #1538, opened 2024-07-24
Test and preliminary fix for issue with "USE_VAR" cleanup

carstenartur:
Reproducer for
https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1525
The "fix" included seems to be a kind of bold hack. A good fix might need a changes to some underlying classes that can not really handle parametrized types for import removal in the case of type interference.
## What it does
For the case of using diamond operater the underlying class to remove the imports of the instance expression is fed with the expression without the type so arraylist<> instead of arraylist<File>. The junit tests are fine but I do not trust this approach so much. And it seems to me there is really a lack of support for paramerized types in the classes in charge to handle the import removal. Maybe it is fine because within the scope of the code changes of this cleanup it is good enough?

## How to test
Try to run it over jdtui code sources. There are several cases where this error pops up.

## Author checklist

- [ x] The change is following the [coding conventions](https://wiki.eclipse.org/Platform/How_to_Contribute#Coding_Conventions)


carstenartur:
@jjohnstn if you feel like proposing or implementing a better fix you are welcome. I still think that this fix is a kind of hack.

jjohnstn (approved):
Hi @carstenartur I think the fix is fine.

Issue #1539, opened 2024-07-24
[Bug] [Folding mechanism] Method declaration on the Same Line as the Closing Bracket


jakub-suliga:
When a method starts on the same line as the closing bracket of the previous method, the folding mechanism does not work correctly. The method declaration disappears along with the closing bracket. Ideally, only the closing bracket should disappear, not the method declaration.

Before the folding:
![image](https://github.com/user-attachments/assets/4cd79d7b-9f76-4e83-92a3-c36a72024f0b)

After the folding the first method:
![image](https://github.com/user-attachments/assets/eb59178f-9cb7-4e73-b9ca-057f6a40db50)

How it should be:
![image](https://github.com/user-attachments/assets/4b11e8af-6a83-4854-b37f-fc63bc78923a)

The code in the example:
```java
public class b {
	void a() {
		
	}void b() {
		
	}
}
```

jakub-suliga:
This PR #1562 should fix it.

Issue #1426, opened 2024-05-27
Folding block code


XelaNimed:
Eclipse IDE for Enterprise Java and Web Developers (includes Incubating components)
Version: 2024-03 (4.31.0)
Build id: 20240307-1437
Windows 11 23H2 x64

I was very surprised that code folding in Eclipse is poor compared to other Java IDEs, and in some cases it is non-existent.
Please add/implement collapsing code blocks in all parentheses, curly/square braces.
There is also no way to define a custom region for collapsing, as is implemented in VisualStudio.

jukzi:
This project is open source - feel free to suggest a change with a Pull Request.

XelaNimed:
@jukzi I do not know how. On the main page of the project, in the [Developer resources](https://github.com/eclipse-jdt/eclipse.jdt.ui#developer-resources) section, the link leads to a page without information (or I was looking in the wrong place). In any case, I'm not sure that I can understand the codebase and come up with a quality solution.

jukzi:
please see https://wiki.eclipse.org/JDT_UI/How_to_Contribute

danthe1st:
I think this issue can be closed now since #1562 has been merged?

jjohnstn:
@XelaNimed In 4.35 (2025-03) which is about to be released, a number of control statements were added for folding (if, for, enhanced for, try, catch, while, do while).  Switch statements are not yet implemented).    In addition, 4.36 M1 (up-coming milestone for 2025-06) will add custom folding regions.  I am going to mark this closed.  If you want to create a separate issue for the missing switch statement that is fine although it is known by the contributor of the patch.

PR #1550, opened 2024-07-26
Folding mechanism for while/for/if/switch-case

jakub-suliga:
Problem:
Currently, there is no folding mechanism for while/for/if/switch-case statements in Eclipse. VSCode and IntelliJ support a folding mechanism for these statements. Therefore, I have created a method that implements this folding mechanism. This is also an open issue: #1426.

Before:
image

After:
image

How to test:
1. Start an Eclipse workspace and create a new class.
2. Add while/for/if/switch-case statements to this class.
3. Verify that the folding mechanism works as expected.

Here is a small program you can use to test:
public class SimpleProgram {
    public void checkNumber(int number) {
        if (number > 0) {
            System.out.println(number + " ist positiv.");
        } else {
            System.out.println(number + " ist null.");
        }
    }

        SimpleProgram program = new SimpleProgram();
        int count = 1;
        while (count <= 5) {
            System.out.println("Count ist: " + count);
            count++;
        }
    }
}

Bugs:
- [ ] #1539
- [ ] When there are multiple nested while/for/if/switch-case statements, not all foldable regions are recognized.

### DIFF
--- a/org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/fix/StringConcatToTextBlockFixCore.java
+++ b/org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/fix/StringConcatToTextBlockFixCore.java
@@ -50,6 +50,7 @@ import org.eclipse.jdt.core.dom.Expression;
 import org.eclipse.jdt.core.dom.ExpressionStatement;
 import org.eclipse.jdt.core.dom.FieldDeclaration;
 import org.eclipse.jdt.core.dom.IBinding;
+import org.eclipse.jdt.core.dom.IMethodBinding;
 import org.eclipse.jdt.core.dom.ITypeBinding;
 import org.eclipse.jdt.core.dom.InfixExpression;
 import org.eclipse.jdt.core.dom.InfixExpression.Operator;
@@ -502,6 +503,7 @@ public class StringConcatToTextBlockFixCore extends CompilationUnitRewriteOperat
 			private int lastStatementEnd;
 			private IBinding varBinding;
 			private List<MethodInvocation> toStringList= new ArrayList<>();
+			private List<SimpleName> argList= new ArrayList<>();
 
 			public CheckValidityVisitor(int lastStatementEnd, IBinding varBinding) {
 				this.lastStatementEnd= lastStatementEnd;
@@ -518,6 +520,10 @@ public class StringConcatToTextBlockFixCore extends CompilationUnitRewriteOperat
 
 			public List<MethodInvocation> getToStringList() {
 				return toStringList;
+			}
+
+			public List<SimpleName> getArgList() {
+				return argList;
 			}
 
 			@Override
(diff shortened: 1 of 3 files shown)


Kuba3105:
@fedejeanne

iloveeclipse:
FYI: the 23 test fails should be unrelated, I've just fixed last test that was failing on master after recent bigger changes on JDT core, so please rebase your branch on master if you want see whether your change breaks anything.

iloveeclipse:
And btw, the feature looks cool.

org.eclipse.jdt.ui/ui/org/eclipse/jdt/ui/text/folding/DefaultJavaFoldingStructureProvider.java
```diff
@@ -979,10 +980,53 @@ private void computeFoldingStructure(FoldingStructureComputationContext ctx) {
 
 			ctx.getScanner().setSource(source.toCharArray());
 			computeFoldingStructure(parent.getChildren(), ctx);
+			addFoldingRegions(source, ctx);
 		} catch (JavaModelException x) {
 		}
 	}
 
+	private void addFoldingRegions(String source, FoldingStructureComputationContext ctx) {
```
fedejeanne, line 988: This name is not very descriptive. Rename to something like:
```suggestion
	private void computeFoldingStructureOfControlStatements(String source, FoldingStructureComputationContext ctx) {
```

fedejeanne (changes requested):
Thank you for contributing this, @jakub-suliga !

I tested it and it looks pretty good. I only have one improvement that I would like to add: the `do-while` loop and rename the new method to something more expressive.

One thing about the `do-while` loop is that it also suffers from https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1539.

This is how it would look like if you include the `do-while` loop to this PR ...

image

... and fold it:
image

But that's a topic for another PR, I just wanted to point it out so it does not get forgotten.

org.eclipse.jdt.ui/ui/org/eclipse/jdt/ui/text/folding/DefaultJavaFoldingStructureProvider.java
```diff
@@ -979,10 +980,53 @@ private void computeFoldingStructure(FoldingStructureComputationContext ctx) {
...
+	    int token;
+	    Stack<Integer> blockStarts = new Stack<>();
+
+	    try {
+	        while ((token = scanner.getNextToken()) != ITerminalSymbols.TokenNameEOF) {
+	            switch (token) {
+	                case ITerminalSymbols.TokenNameif:
+	                case ITerminalSymbols.TokenNamewhile:
+	                case ITerminalSymbols.TokenNamefor:
+	                case ITerminalSymbols.TokenNameswitch:
```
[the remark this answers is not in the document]
jakub-suliga: Done

org.eclipse.jdt.ui/ui/org/eclipse/jdt/ui/text/folding/DefaultJavaFoldingStructureProvider.java

PR #1559, opened 2024-07-30
Fix "fix return type" quickfix for type param return type

### LINKED ISSUE #1558: Quickfix for return type that's a wildcard doesn't work post 1.4
The quickfix to fix the return type in the following code doesn't work.

```java
import java.lang.reflect.AccessibleObject;

public class WeirdAnonymousThingy {
	
	@Override
	public void m() {
		new AccessibleObject() {
			public void getAnnotation(Class annotationClass) {
				
			}
		};
	}
}
```

It's a particular nasty edge case: this anonymous class inherits a method that has a method-level type parameter with a bound set, and the return type is that type variable. (see [the javadoc for AccessibleObject.getAnnotation](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/reflect/AccessibleObject.html#getAnnotation(java.lang.Class)))

Pre-1.5, the erasure of the type variable is calculated as the value that `void` gets replaced with. Post 1.5, the type variable is inserted in place of `void`, which results in invalid code, since the type variable is declared in the override implementation.

I think that ideally this quickfix should provide two options:
1. Returning a type variable (and properly declaring it)
2. Returning the erasure of the type variable

datho7561:
Fixes #1558

## What it does
The old implementation would propose replacing the return type with erasure of the type parameter if the compliance was set pre 1.5. If the compliance was 1.5 or later, it would propose replacing the return type with the type parameter as it appears in the declaration of the method being overridden.

Instead, these are now two separate entries which may be generated:
1. **Replace the return type with the erasure of the type parameter**

This entry is only generated if it is valid to use the erasure for the return type. Specifically, if the method declares the type parameters of the method being overriden, and one of type variables has an upper bound that is a parameterized type, then using the type erasure of that variable in place of the variable is not valid, since the type erasure allows a wider range of types. Code example:

```java
public class Example {
  private abstract static class Parent {
    <E extends List<String>> E myMethod(E arg1);
  }
  private static class Child extends Parent {
    @Override
    <E extends List<String>> List myMethod(E arg1) { return null; }; // NOT VALID, because it doesn't override myMethod
  }
}
```

2. **Replace the return type with the type parameter**

This entry now takes into account that the type parameters may not yet be declared on the type. i.e. if the method has no type parameters, the quick fix introduces them, adds the correct bounds to them, and augments the existing parameter types with the type variables using the information from the parent declaration.

This entry also takes into account the fact that the type parameters may have different names in the overriding method than in the original declaration.

## How to test
Try out the quickfixes by invoking quickfix (Ctrl+1) in the snippet from the linked issue.

Also check out the snippets from the test cases, which go further in depth.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.manipulation/common/org/eclipse/jdt/internal/ui/text/correction/TypeMismatchBaseSubProcessor.java
+++ b/org.eclipse.jdt.core.manipulation/common/org/eclipse/jdt/internal/ui/text/correction/TypeMismatchBaseSubProcessor.java
@@ -17,6 +17,7 @@ package org.eclipse.jdt.internal.ui.text.correction;
 
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.stream.Stream;
 
 import org.eclipse.core.runtime.CoreException;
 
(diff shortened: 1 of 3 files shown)


akurtakov:
Do you think you can add test for this fix? https://github.com/eclipse-jdt/eclipse.jdt.ui/blob/2327a09d95c260ad694cb5eca92e648b0b903564/org.eclipse.jdt.ui.tests/ui/org/eclipse/jdt/ui/tests/quickfix/TypeMismatchQuickFixTests.java looks like a good candidate for such test from first look?

datho7561:
Yes, I'll add a test for this. Thanks for the pointer!

I'm waiting to see the test result to see if I broken any existing tests.

datho7561:
@jjohnstn If you have some time, could you please take a look at this PR?

jjohnstn:
@datho7561 Sure, I'll look at it today.

datho7561:
Thank you!

jjohnstn (approved):
Looks good.  Thanks.

Issue #1095, opened 2024-01-18
Random failing MarkOccurrenceTest.markOccurrencesAfterEditorReuse


jukzi:
```
java.lang.AssertionError: expected:<1> but was:<0>
	at org.junit.Assert.fail(Assert.java:89)
	at org.junit.Assert.failNotEquals(Assert.java:835)
	at org.junit.Assert.assertEquals(Assert.java:647)
	at org.junit.Assert.assertEquals(Assert.java:633)
	at org.eclipse.jdt.text.tests.MarkOccurrenceTest$2.condition(MarkOccurrenceTest.java:414)
	at org.eclipse.jdt.text.tests.performance.DisplayHelper.waitForCondition(DisplayHelper.java:79)
	at org.eclipse.jdt.text.tests.MarkOccurrenceTest.assertOccurrences(MarkOccurrenceTest.java:421)
	at org.eclipse.jdt.text.tests.MarkOccurrenceTest.markOccurrencesAfterEditorReuse(MarkOccurrenceTest.java:210)
```
https://ci.eclipse.org/jdt/job/eclipse.jdt.ui-github/job/PR-1086/3/testReport/junit/JdtTextTestSuite%20org.eclipse.jdt.text.tests/MarkOccurrenceTest/markOccurrencesAfterEditorReuse/

jukzi:
may relate to https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/80

jukzi:
can not reproduce locally.

stephan-herrmann:
Seen in a Y-build with exact same symptom: https://ci.eclipse.org/releng/job/YPBuilds/job/ep434Y-unit-cen64-gtk3-java24/8/testReport/org.eclipse.jdt.text.tests/MarkOccurrenceTest/markOccurrencesAfterEditorReuse/

PR #1562, opened 2024-07-31
Folding mechanism for while/for/if/switch-case

jakub-suliga:
Problem:
Currently, there is no folding mechanism for while/for/if/switch-case statements in Eclipse. VSCode and IntelliJ support a folding mechanism for these statements. Therefore, I have created a method that implements this folding mechanism. This is also an open issue: #1426. In addition, I deleted some commented-out code from 2007.

Before:
image

After:
image

How to test:
1. Start an Eclipse workspace and create a new class.
2. Add while/for/if/switch-case statements to this class.
3. Verify that the folding mechanism works as expected.

Here is a small program you can use to test:
public class SimpleProgram {
    public void checkNumber(int number) {
        if (number > 0) {
            System.out.println(number + " ist positiv.");
        } else {
            System.out.println(number + " ist null.");
        }
    }

        SimpleProgram program = new SimpleProgram();
        int count = 1;
        while (count <= 5) {
            System.out.println("Count ist: " + count);
            count++;
        }
    }
}

Bugs:
- [ ] #1539
- [ ] When there are multiple nested while/for/if/switch-case statements, not all foldable regions are recognized.

### DIFF
--- a/org.eclipse.jdt.ui/ui/org/eclipse/jdt/ui/text/folding/DefaultJavaFoldingStructureProvider.java
+++ b/org.eclipse.jdt.ui/ui/org/eclipse/jdt/ui/text/folding/DefaultJavaFoldingStructureProvider.java
@@ -25,6 +25,7 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.Stack;
 
 import org.eclipse.core.runtime.Assert;
 
@@ -476,7 +477,6 @@ public class DefaultJavaFoldingStructureProvider implements IJavaFoldingStructur
 
 			IRegion preRegion;
 			if (firstLine < captionLine) {
-//				preRegion= new Region(offset + prefixEnd, contentStart - prefixEnd);
 				int preOffset= document.getLineOffset(firstLine);
 				IRegion preEndLineInfo= document.getLineInformation(captionLine);
 				int preEnd= preEndLineInfo.getOffset();
@@ -519,33 +519,6 @@ public class DefaultJavaFoldingStructureProvider implements IJavaFoldingStructur
 			}
 			return 0;
 		}
-
-//		/**
-//		 * Finds the offset of the first identifier part within <code>content</code>.
-//		 * Returns 0 if none is found.
-//		 *
-//		 * @param content the content to search
-//		 * @return the first index of a unicode identifier part, or zero if none can
-//		 *         be found
-//		 */
-//		private int findPrefixEnd(final CharSequence content) {
-//			// return the index after the leading '/*' or '/**'
-//			int len= content.length();
-//			int i= 0;
-//			while (i < len && isWhiteSpace(content.charAt(i)))
-//				i++;
-//			if (len >= i + 2 && content.charAt(i) == '/' && content.charAt(i + 1) == '*')
-//				if (len >= i + 3 && content.charAt(i + 2) == '*')
-//					return i + 3;
-//				else
-//					return i + 2;
-//			else
-//				return i;
-//		}
-//
-//		private boolean isWhiteSpace(char c) {
-//			return c == ' ' || c == '\t';
-//		}
 
 		/*
 		 * @see org.eclipse.jface.text.source.projection.IProjectionPosition#computeCaptionOffset(org.eclipse.jface.text.IDocument)
@@ -979,8 +952,52 @@ public class DefaultJavaFoldingStructureProvider implements IJavaFoldingStructur
 
 			ctx.getScanner().setSource(source.toCharArray());
 			computeFoldingStructure(parent.getChildren(), ctx);
+			computeFoldingStructureOfControlStatements(source, ctx);
 		} catch (JavaModelException x) {
 		}
+	}
+
+	private void computeFoldingStructureOfControlStatements(String source, FoldingStructureComputationContext ctx) {
+	    IScanner scanner = ctx.getScanner();
+	    scanner.setSource(source.toCharArray());
+
+	    int token;
+	    Stack<Integer> blockStarts = new Stack<>();
+
+	    try {
+	        while ((token = scanner.getNextToken()) != ITerminalSymbols.TokenNameEOF) {
+	            switch (token) {
+	                case ITerminalSymbols.TokenNameif:
+	                case ITerminalSymbols.TokenNamewhile:
+	                case ITerminalSymbols.TokenNamefor:
+	                case ITerminalSymbols.TokenNameswitch:
+	                case ITerminalSymbols.TokenNamedo:
+	                    blockStarts.push(scanner.getCurrentTokenStartPosition());
+	                    break;
+	                case ITerminalSymbols.TokenNameLBRACE:
+	                    if (!blockStarts.isEmpty()) {
+	                        blockStarts.push(scanner.getCurrentTokenStartPosition());
+	                    }
+	                    break;
+	                case ITerminalSymbols.TokenNameRBRACE:
+	                    if (!blockStarts.isEmpty()) {
+	                        int start = blockStarts.pop();
+	                        if (!blockStarts.isEmpty() && source.charAt(start) == '{') {
+	                            IRegion region = new Region(start + 1, scanner.getCurrentTokenEndPosition() - start);
+	                            IRegion normalized = alignRegion(region, ctx);
+	                            if (normalized != null) {
+	                                Position position = createCommentPosition(normalized);
+	                                if (position != null) {
+	                                    ctx.addProjectionRange(new JavaProjectionAnnotation(false, null, false), position);
+	                                }
+	                            }
+	                        }
+	                    }
+	                    break;
+	            }
+	        }
+	    } catch (InvalidInputException e) {
+	    }
 	}
 
 	private void computeFoldingStructure(IJavaElement[] elements, FoldingStructureComputationContext ctx) throws JavaModelException {


org.eclipse.jdt.ui/ui/org/eclipse/jdt/ui/text/folding/DefaultJavaFoldingStructureProvider.java
```diff
@@ -520,33 +520,6 @@ private int findFirstContent(final CharSequence content, int prefixEnd) {
 			return 0;
 		}
 
-//		/**
```
iloveeclipse, line 523: Please remove commented out code.

jakub-suliga:
Bug:
- [ ] When you press enter or delete a line, the folding mechanism gets buggy

jakub-suliga:
To test the code, you can use my test class:

`public class test {

    
    private int zaehler;
    private final Object sperre = new Object();

    
    public test() {
        zaehler = 0;
    }

    
    
    public void verarbeiten() {
        if (zaehler > 0) {
            System.out.println();
        } else if (zaehler == 0) {
            System.out.println();
        } else {
            System.out.println();
        }

        while (zaehler < 10) {
            zaehler++;
        }

        for (int i = 0; i < zaehler; i++) {
            System.out.println();
        }

        do {
            zaehler--;
        } while (zaehler > 0);

        int[] zahlen = {1, 2, 3, 4, 5};
        for (int zahl : zahlen) {
            System.out.println();
        }

        switch (zaehler) {
            case 0:
                System.out.println();
                break;
            case 1:
                System.out.println();
                break;
            default:
                System.out.println();
                break;
        }

        try {
            int ergebnis = 10 / zaehler;
        } catch (ArithmeticException e) {
            System.out.println();
        } finally {
            System.out.println();
        }

        synchronized (sperre) {
            zaehler++;
        }

        Runnable runnable = () -> {
            System.out.println();
        };
        runnable.run();

        {
            System.out.println();
        }

        aussen:
        for (int i = 0; i < 5; i++) {
            for (int j = 0; j < 5; j++) {
                if (i * j > 6) {
                    break aussen;
                }
                System.out.println();
            }
        }

        try {
            String text = null;
            text.length();
        } catch (NullPointerException e) {
            System.out.println();
        } catch (Exception e) {
            System.out.println();
        } finally {
            System.out.println();
        }

        {
            System.out.println();
        }

        if (zaehler > 0) {
            for (int i = 0; i < zaehler; i++) {
                while (i < 5) {
                    System.out.println();
                    i++;
                }
            }
        }

        lambdaVerwenden();
    }


    public void lambdaVerwenden() {
        verarbeiteLambda(() -> {
            System.out.println();
        });
    }

    private void verarbeiteLambda(Runnable runnable) {
        runnable.run();
    }
}`

fedejeanne:
The test failures seem to be unrelated: https://ci.eclipse.org/jdt/job/eclipse.jdt.ui-github/job/PR-1562/21/testReport/

The only test that fails with _age == 1_ (_i.e._ in this PR) is `org.eclipse.jdt.text.tests.PluginsNotLoadedTest.pluginsNotLoaded` but looking at its stack trace I don't see any hints that point to this PR either:

```java
java.lang.AssertionError: 
Wrong bundles loaded:
- org.eclipse.jdt.junit
 expected:<0> but was:<24>
	at org.junit.Assert.fail(Assert.java:89)
	at org.junit.Assert.failNotEquals(Assert.java:835)
	at org.junit.Assert.assertEquals(Assert.java:647)
	at org.eclipse.jdt.text.tests.PluginsNotLoadedTest.pluginsNotLoaded(PluginsNotLoadedTest.java:278)
...
```

Does anyone know the reason for this failure?

In the meantime @jakub-suliga, try with another force-push to re-trigger the checks and see if the failure persists.

fedejeanne:
> In the meantime @jakub-suliga, try with another force-push to re-trigger the checks and see if the failure persists.

On a closer look, the test failures are because of the changes introduced in this PR. Run them locally and you will see it. 

Here's (part of) the console output when running 

```java
!STACK 0
java.lang.ClassCastException: class org.eclipse.jdt.internal.core.ClassFile cannot be cast to class org.eclipse.jdt.core.ICompilationUnit (org.eclipse.jdt.internal.core.ClassFile and org.eclipse.jdt.core.ICompilationUnit are in unnamed module of loader org.eclipse.osgi.internal.loader.EquinoxClassLoader @21452f5e)
	at org.eclipse.jdt.ui.text.folding.DefaultJavaFoldingStructureProvider.computeFoldingStructure(DefaultJavaFoldingStructureProvider.java:981)
	at org.eclipse.jdt.ui.text.folding.DefaultJavaFoldingStructureProvider.update(DefaultJavaFoldingStructureProvider.java:907)
	at org.eclipse.jdt.ui.text.folding.DefaultJavaFoldingStructureProvider.initialize(DefaultJavaFoldingStructureProvider.java:852)
	at org.eclipse.jdt.ui.text.folding.DefaultJavaFoldingStructureProvider.handleProjectionEnabled(DefaultJavaFoldingStructureProvider.java:822)
	at org.eclipse.jdt.ui.text.folding.DefaultJavaFoldingStructureProvider$ProjectionListener.projectionEnabled(DefaultJavaFoldingStructureProvider.java:700)
...
	at org.eclipse.jdt.ui.tests.quickfix.AnnotateAssistTest1d5.testAnnotateReturn2(AnnotateAssistTest1d5.java:178)
...
```

fedejeanne:
@iloveeclipse : Jakub and I were looking into the failed tests and we found the culprit: it was a `null`. After some debugging and evaluating we decided that a null-check would make sense since now (in this PR) it is possible to add folding structures for things that are **not** `IJavaElement`s. This led to commit ade7b52c9c11d90827dee063b232170befcd6890.

Do you see any possible problem with this change? We couldn't come up with any but maybe you can?

iloveeclipse:
> Do you see any possible problem with this change?

Looks OK, but I don't have IDE access till monday.

Note, there is a test fail which seem to be related: https://ci.eclipse.org/jdt/job/eclipse.jdt.ui-github/job/PR-1562/26/testReport/(root)/JdtTextTestSuite%20ParameterNamesCodeMiningTest/testCollapsedFoldingAndToggleHighlight/

Also it seem that you have removed two protected methods? Isn't this class public API (can't see it in browser). If so, you can't do that. Also there are two additional warnings added, please check.

fedejeanne:
> Note, there is a test fail which seem to be related: https://ci.eclipse.org/jdt/job/eclipse.jdt.ui-github/job/PR-1562/26/testReport/(root)/JdtTextTestSuite%20ParameterNamesCodeMiningTest/testCollapsedFoldingAndToggleHighlight/

Yes, Jakub will look into that.

> Also it seem that you have removed two protected methods? 

The methods are still there, they are lost in the big chunk of changes. Regarding that, @jakub-suliga please extract the anonymous class to a new private class and put it at the beginning of the class, this should move the big chunk of new code and make the reviewing easier.

jakub-suliga:
There are some Bugs:

- [ ] If there is a javadoc above a method, the folding button disapears
- [ ] When there are multiple nested if statements, only the first foldable regions is recognized.
- [ ] Sometimes a return/contuine gets a folding button -> I think when the return is in a if stament
- [ ] When a while/ if stament or something else is in a try/switch block, they dont get a folding button

jakub-suliga:
Bugs:
- [ ] Inner classes do not have a folding button.
- [ ] if statements without braces do not fold properly.
- [ ] Copyright Javadoc comments do not create a folding region.
- [ ] Class definitions do not generate a folding region.
- [ ] In a switch-case statement, when one case section immediately follows another, the editor creates a folding button that encompasses both sections. However, clicking the first folding button doesn't have any effect.

danthe1st:
Hi @jakub-suliga
I have been working on changes to `DefaultJavaFoldingStructureProvider` aside you (#1825) and I checked for merge conflicts and ran my tests to check for similar issues (out of curiosity) by rebasing your changes on top of mine.

While there were (unsurprisingly) some merge conflicts, these were all trivial to resolve.

My tests noticed the following changes to folding to the current folding when using your code (see https://github.com/danthe1st/eclipse.jdt.ui/commit/d0c8ba814824844fc06997ff5b7bb3ad0a84b487 where I adapted my tests - if your PR is merged after mine, you might have to do something similar):

If there is a single import, your implementation adds a folding region for that import.
```java
package a;

import java.util.List;

```

If a class contains fields, your implementation adds a folding region for every field of the class. In the following example, `int a;` is folded. With multiple fields, each field gets its own folding region.
```java
package a;

public class JakobTest2 {
	int a;
}
```
image

These are changes in behavior from the current `master` (these also occur without my custom folding changes) introduced by your PR if your "New Folding" option is disabled. I want to inform you about this changes since they seem to be unintentional (at least the second one with the fields).

Another change you made that interacted with my test is your removal of `CommentPosition`. This changes where exactly comment regions "end" (it essentially removes some aligning done by `CommentPosition` which my test relies on) but that shouldn't have any impact to users because the end of the region should still be in the same line (I think) but it would be necessary to relax the checking of the tests (https://github.com/danthe1st/eclipse.jdt.ui/commit/d0c8ba814824844fc06997ff5b7bb3ad0a84b487#diff-e8836e7c6985ef3dc58865becd219f6708db0854c28f22c7b29843456c704652R379) which may be useful for you to know _if_ my PR gets merged before yours.

You can see my experiment here: https://github.com/danthe1st/eclipse.jdt.ui/tree/jakob-folding-test

jakub-suliga:
Hi @danthe1st,
thank you for reviewing my PR! I fixed the bug with the single import, but I couldn’t reproduce the second bug.
image
After disabling my option, you need to restart the IDE. I’ll add a message to inform users that a restart is required.
I also removed CommentPosition because I didn’t need it, but feel free to add it back if you still need it.

danthe1st:
> Hi @danthe1st, thank you for reviewing my PR! I fixed the bug with the single import, but I couldn’t reproduce the second bug.
After disabling my option, you need to restart the IDE. I’ll add a message to inform users that a restart is required. I also removed CommentPosition because I didn’t need it, but feel free to add it back if you still need it.

For me, this issue also happens even when starting Eclipse without your "new folding" enabled (and without enabling it).
[Here](https://github.com/danthe1st/eclipse.jdt.ui/blob/d0c8ba814824844fc06997ff5b7bb3ad0a84b487/org.eclipse.jdt.ui.tests/ui/org/eclipse/jdt/ui/tests/folding/CustomFoldingRegionTest.java#L323-L343) is one test for it (the test is for my changes and it's adapted to "allow" for your behavior so you'd have to change the `assertEquals(3, projectionRanges.size());` to `assertEquals(1, projectionRanges.size());` and the following assertions to remove all but `assertContainsRegionUsingStartAndEndLine(projectionRanges, str, 6, 8);//void b()`):
```java
public class FoldingTest {
	@Rule
	public ProjectTestSetup projectSetup= new ProjectTestSetup();

	private IJavaProject fJProject1;

	private IPackageFragmentRoot fSourceFolder;

	private IPackageFragment fPackageFragment;

	@Before
	public void setUp() throws CoreException {
		fJProject1= projectSetup.getProject();
		fSourceFolder= JavaProjectHelper.addSourceContainer(fJProject1, "src");
		fPackageFragment= fSourceFolder.createPackageFragment("org.example.test", false, null);
	}


	@After
	public void tearDown() throws CoreException {
		JavaProjectHelper.delete(fJProject1);
	}
	@Test
	public void testCustomRegionsAroundFieldAndMethod() throws JavaModelException, PartInitException {
		String str= """
				package org.example.test;

				public class Test {
					// #region
					int a;

					void b(){

					}
					// #endregion
				}
				""";
		List<IRegion> projectionRanges= getProjectionRangesOfFile(str);
		assertEquals(1, projectionRanges.size());
		assertContainsRegionUsingStartAndEndLine(projectionRanges, str, 6, 8);//void b()
		//assertContainsRegionUsingStartAndEndLine(projectionRanges, str, 4, 5);//int a;
	}
	private void assertContainsRegionUsingStartAndEndLine(List<IRegion> projectionRanges, String input, int startLine, int endLine) {
		assertTrue(startLine <= endLine, "start line must be smaller or equal to end line");
		int startLineBegin= findLineStartIndex(input, startLine);

		int endLineBegin= findLineStartIndex(input, endLine);
		int endLineEnd= findNextLineStart(input, endLineBegin);
		endLineEnd= getLengthIfNotFound(input, endLineEnd);

		for (IRegion region : projectionRanges) {
			if (region.getOffset() == startLineBegin + 1 && region.getOffset() + region.getLength() >= endLineBegin && region.getOffset() + region.getLength() <= endLineEnd) {
				return;
			}
		}

		fail(
				"missing region from line " + startLine + " (index " + (startLineBegin + 1) + ") " +
						"to line " + endLine + " (index " + (endLineEnd) + ", length "+(endLineEnd - (startLineBegin + 1)) + ")" +
						", actual regions: " + projectionRanges
		);
	}


	private int getLengthIfNotFound(String input, int startLineEnd) {
		if (startLineEnd == -1) {
			startLineEnd= input.length();
		}
		return startLineEnd;
	}


	private int findLineStartIndex(String input, int lineNumber) {
		int currentInputIndex= 0;
		for (int i= 0; i < lineNumber; i++) {
			currentInputIndex= findNextLineStart(input, currentInputIndex);
			if (currentInputIndex == -1) {
				fail("line number is greater than the total number of lines");
			}
		}
		return currentInputIndex;
	}


	private int findNextLineStart(String input, int currentInputIndex) {
		return input.indexOf('\n', currentInputIndex + 1);
	}

	private List<IRegion> getProjectionRangesOfFile(String str) throws JavaModelException, PartInitException {

		ICompilationUnit compilationUnit= fPackageFragment.createCompilationUnit("Test.java", str, false, null);
		JavaEditor editor= (JavaEditor) EditorUtility.openInEditor(compilationUnit);
		ProjectionAnnotationModel model= editor.getAdapter(ProjectionAnnotationModel.class);
		List<IRegion> regions= new ArrayList<>();
		for (Iterator<Annotation> it= model.getAnnotationIterator(); it.hasNext();) {
			Annotation annotation= it.next();
			if (annotation instanceof ProjectionAnnotation projectionAnnotation) {
				Position position= model.getPosition(projectionAnnotation);
				regions.add(new Region(position.getOffset(), position.getLength()));
			}
		}
		return regions;
	}

}
```

jakub-suliga:
#Is this absolutely necessary? Is there any reason it shouldn't work by just reopening the source file (and can you get around that - that seems to be quite a limitation)?
I added a ChangeListener, now you dont need to restart your IDE. I will have a look on your test.

danthe1st:
> I added a ChangeListener, now you dont need to restart your IDE.

Oh I see what you meant. I think that without that listener, it should still apply the changes but only after reopening the file (at least that's what happened in my experience).

jakub-suliga:
>  I don't think a restart of the IDE is needed in either way.

I didnt test it with closing the file, probably it works too.

jakub-suliga:
@danthe1st I looked into your test and found the bug, now it should work.

danthe1st:
> I looked into your test and found the bug, now it should work.

I can confirm that this works now.
I want to note another difference I found with your current change without your new folding enabled but I actually think that might be a good thing.

Let's say we have a class like this:
```java
public class SomeTest {
	void b(){

	}
}
```
Here, the `b()` method is folded both with and without your change but there is a slight difference in the way it's folded.
In the current `master`, the `}` is included in the folding:
image
image
However, in your version, it isn't:
image
image

While I personally prefer the way it's done with your version, this is a change in the behavior of the folding without your option being enabled so I think it's worth noting.

jakub-suliga:
I added it because it fixes this #1539. 

For example you have this:
`public class Test {
	void b(){

	} void c(){

	}
}`
image


and you want to fold the top method, the lower method is folded too (the old version): 
image

With the change, you can see both methods (my version):
image

fedejeanne:
> I added it because it fixes this #1539.
>  ...

@jakub-suliga please provide a separate PR for that fix, otherwise it would get lost if someone decides that this PR needs to be reverted in the future. Your PR should be then be based on that separate PR and should keep the fix (_i.e._ it should **not** introduce a regression)

jakub-suliga:
> please provide a separate PR for that fix

Done. #1939

org.eclipse.jdt.ui.tests/ui/org/eclipse/jdt/ui/tests/folding/FoldingTest.java
```diff
@@ -0,0 +1,363 @@
...
+                                do {
+                                } while(false);
+                            }
+                        }
+                    }
+                }
+            }
+            """;
+        List<IRegion> regions = getProjectionRangesOfFile(str);
+        assertTrue(regions.size() >= 4);
```
fedejeanne: NIT: I would have extracted the logic of compiling the code and asserting the amount of expected regions in a separate method and called that method in each test to improve readability. 

Something like: 

```java
private void assertCodeHasRegions(String code, int regionsCount) throws Exception {
	List<IRegion> regions = getProjectionRangesOfFile(code);
	assertEquals(regionsCount, regions.size());
}
```

org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/text/folding/DefaultJavaFoldingPreferenceBlock.java
```diff
@@ -102,6 +102,12 @@ public Control createControl(Composite composite) {
 		addCheckBox(inner, FoldingMessages.DefaultJavaFoldingPreferenceBlock_methods, PreferenceConstants.EDITOR_FOLDING_METHODS, 0);
 		addCheckBox(inner, FoldingMessages.DefaultJavaFoldingPreferenceBlock_imports, PreferenceConstants.EDITOR_FOLDING_IMPORTS, 0);
 
+		Label label2= new Label(inner, SWT.LEFT);
```
fedejeanne, line 105: You shouldn't need this. Try to improve the overall design of the whole configuration page and make it look more like the **Content Assist** page, which has `Group`s (with borders). You should have 2 groups:
- _Initially fold ..._
- Extended folding (Experimental) : This one is yours and has 
  - A `Label` with an explanation like "_Enable folding for other code blocks. Activating this feature may cause a significant performance degradation_" 
  - A checkbox: "_Activate feature_".

fedejeanne (commented):
Some minor comments for the moment.

org.eclipse.jdt.ui.tests/ui/org/eclipse/jdt/ui/tests/folding/FoldingTest.java
```diff
@@ -0,0 +1,363 @@
...
+            package org.example.test;
+            public class Q {
+                void a() {
+        		    int i = 0;
+                }void b() {
+                }
+            }
+            """;
+        List<IRegion> regions = getProjectionRangesOfFile(str);
+        assertTrue(regions.size() >= 2);
```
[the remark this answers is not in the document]
jakub-suliga: Done

org.eclipse.jdt.ui.tests/ui/org/eclipse/jdt/ui/tests/folding/FoldingTest.java
```diff
@@ -0,0 +1,363 @@
...
+            sourceFolder = org.eclipse.jdt.testplugin.JavaProjectHelper.addSourceContainer(jProject, "src");
+        }
+        packageFragment = sourceFolder.createPackageFragment("org.example.test", false, null);
+        IPreferenceStore store = JavaPlugin.getDefault().getPreferenceStore();
+        store.setValue(PreferenceConstants.EDITOR_NEW_FOLDING_ENABLED, true);
+    }
+
+    @After
+    public void tearDown() throws CoreException {
+        org.eclipse.jdt.testplugin.JavaProjectHelper.delete(jProject);
```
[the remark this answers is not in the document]
jakub-suliga: Done

org.eclipse.jdt.ui.tests/ui/org/eclipse/jdt/ui/tests/folding/FoldingTest.java
```diff
@@ -0,0 +1,363 @@
...
+                    }
+                }
+            }
+            """;
+        List<IRegion> regions = getProjectionRangesOfFile(str);
+        assertTrue(regions.size() >= 2);
+    }
+
+    @Test
+    public void testTryStatementFolding() throws JavaModelException, PartInitException {
```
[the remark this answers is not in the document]
jakub-suliga: Can you explain what you mean by that?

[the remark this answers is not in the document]
jakub-suliga: Done

org.eclipse.jdt.ui.tests/ui/org/eclipse/jdt/ui/tests/folding/FoldingTest.java
jakub-suliga: Done

fedejeanne:
The test failure is unrelated. https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1095

org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/text/folding/DefaultJavaFoldingPreferenceBlock.java
jakub-suliga: Do you want it like this:
image

org.eclipse.jdt.ui/ui/org/eclipse/jdt/ui/text/folding/DefaultJavaFoldingStructureProvider.java
```diff
@@ -308,6 +330,314 @@ private static final class Tuple {
...
+			if (end > start) {
+				IRegion region= new Region(start, end - start);
+				IRegion aligned= alignRegion(region, ctx);
+
+				if (aligned != null && isMultiline(aligned)) {
+					Position position= new Position(aligned.getOffset(), aligned.getLength());
+					JavaProjectionAnnotation annotation= new JavaProjectionAnnotation(collapse, null, false);
+					ctx.addProjectionRange(annotation, position);
+				}
+			}
```
fedejeanne, line 498: Possible duplicated code (is not **identical** though)

image

org.eclipse.jdt.ui.tests/ui/org/eclipse/jdt/ui/tests/folding/FoldingTest.java
```diff
@@ -0,0 +1,363 @@
...
+
+import org.eclipse.jdt.ui.PreferenceConstants;
+import org.eclipse.jdt.ui.tests.core.rules.ProjectTestSetup;
+
+import org.eclipse.jdt.internal.ui.JavaPlugin;
+import org.eclipse.jdt.internal.ui.javaeditor.EditorUtility;
+import org.eclipse.jdt.internal.ui.javaeditor.JavaEditor;
+import org.eclipse.jdt.internal.ui.javaeditor.JavaSourceViewer;
+
+public class FoldingTest {
```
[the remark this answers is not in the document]
jakub-suliga: How do I add this to the test suite?

org.eclipse.jdt.ui/ui/org/eclipse/jdt/ui/text/folding/DefaultJavaFoldingStructureProvider.java
```diff
@@ -308,6 +330,314 @@ private static final class Tuple {
...
+					return startSearch + index;
+				}
+			} catch (BadLocationException e) {
+				e.printStackTrace();
+			}
+			return node.getFinally().getStartPosition();
+		}
+	}
+
+	private IPropertyChangeListener fPropertyChangeListener = new IPropertyChangeListener() {
```
fedejeanne, line 631: Move it to the section /* context and listeners */ with the other fields.

org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/text/folding/DefaultJavaFoldingPreferenceBlock.java
jakub-suliga: Done

org.eclipse.jdt.ui.tests/ui/org/eclipse/jdt/ui/tests/folding/FoldingTest.java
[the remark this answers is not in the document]
fedejeanne: Please mark this discussion as "resolved" and follow up in here if necessary: https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/1939#discussion_r1920008811.

It's the same class so it should be the same discussion.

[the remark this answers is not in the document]
jakub-suliga: Done

org.eclipse.jdt.ui/ui/org/eclipse/jdt/ui/text/folding/DefaultJavaFoldingStructureProvider.java

fedejeanne:
The build failure seems to be unrelated, it's a timeout. Maybe it will work next time it runs (_e.g._ when you rebase and push again).

```java
15:42:55  [ERROR] Failed to execute goal org.eclipse.tycho:tycho-apitools-plugin:4.0.11-SNAPSHOT:verify (verify) on project org.eclipse.jdt.ui: Execution verify of goal org.eclipse.tycho:tycho-apitools-plugin:4.0.11-SNAPSHOT:verify failed: Could not mirror artifact osgi.bundle,org.mortbay.jasper.apache-jsp.source,9.0.96 into the local Maven repository.See log output for details. request timed out -> [Help 1]
```

org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/text/folding/DefaultJavaFoldingPreferenceBlock.java
fedejeanne: Can you let the groups adapt better to the size of the dialog, just like with the **Content Assist**?

resize_preferences_page

fedejeanne:
The code looks much better now, thank you for the corrections @jakub-suliga.

Maybe after you fix my last https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/1562#discussion_r1920262597, a JDT committer (@iloveeclipse?) would like to review again? 

I'm off for today, have a nice weekend!

org.eclipse.jdt.ui/ui/org/eclipse/jdt/ui/text/folding/DefaultJavaFoldingStructureProvider.java
```diff
@@ -1004,6 +1317,54 @@ private void computeFoldingStructure(FoldingStructureComputationContext ctx) {
...
+			int end= document.getLineOfOffset(region.getOffset() + region.getLength() - 1);
+			if (start >= end)
+				return null;
+
+			int offset = document.getLineOffset(start);
+			int endOffset = document.getLineOffset(end);
+
+			return new Region(offset, endOffset - offset);
+	    } catch (BadLocationException e) {
+	        e.printStackTrace();
```
iloveeclipse, line 1351: Please remove **all** output to console like `e.printStackTrace();` and either do not report that (if the code works fine afterwards) or properly log to the error log.

org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/folding/FoldingTest.java
```diff
@@ -0,0 +1,363 @@
...
+					foundCollapsed+= 1;
+				}
+			}
+		}
+		assertTrue(foundCollapsed == 2);
+	}
+
+	private void assertCodeHasRegions(String code, int regionsCount) throws Exception {
+		List<IRegion> regions= getProjectionRangesOfFile(code);
+		assertEquals(regionsCount, regions.size());
```
iloveeclipse, line 361: Please give a message explaining fail like "Expected N regions but seen M", just seeing "2 not equals to 1" is disappointing .

org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/folding/FoldingTest.java
```diff
@@ -0,0 +1,363 @@
...
+				""";
+		List<IRegion> regions= getProjectionRangesOfFile(str);
+		assertCodeHasRegions(str, 2);
+		regions.sort((r1, r2) -> Integer.compare(r1.getOffset(), r2.getOffset()));
+		IRegion methodARegion= regions.get(0);
+		IRegion methodBRegion= regions.get(1);
+		int methodAEnd= methodARegion.getOffset() + methodARegion.getLength();
+		int methodBStart= methodBRegion.getOffset();
+
+		assertTrue(methodBStart >= methodAEnd);
```
iloveeclipse, line 101: This assert needs extra message explaining what we expect and why we fail the tes.

iloveeclipse:
I can't guarantee I will have time next week, I'm overloaded. The feature itself seems to be useful, so if the new test also covers existing functionality & is green, why not merge?

@jjohnstn : are you available to review this?

org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/folding/FoldingTest.java
jakub-suliga: Done

org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/folding/FoldingTest.java
jakub-suliga: Done

org.eclipse.jdt.ui/ui/org/eclipse/jdt/ui/text/folding/DefaultJavaFoldingStructureProvider.java

org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/text/folding/DefaultJavaFoldingPreferenceBlock.java
jakub-suliga: No, because it enables folding not only for if, for, and while blocks, but also for constructs such as try/catch blocks, lambdas and among others.

jjohnstn:
@iloveeclipse Sure

jjohnstn:
@jakub-suliga Just trying it out, it seems to work well.  I tried the given test but it does not fold the switch statement or its cases.  Is that expected or is that something that isn't working in my particular test?

I ran into a major bug with folding on my Linux platform which is likely Linux specific.  If I fold, then hover over the plus and then click on the hover to focus it; there is no way out but Esc.  This closes the hover but the flashing cursor is missing and the window pointer can get changed to the window expansion characters if I am on the edges of the hover area at the time of  escape.  I got similar experiences sometimes after entering via F2 to get focus in the hover window, but it seemed better and I often got the flashing cursor back.  Once the cursor was gone, I could only get it back by switching into another Eclipse and its editor and then switching back.  Again, likely Fedora specific, but I ask in case you have Windows/Mac and can confirm that it doesn't occur for you.  It is not caused by this PR.

jjohnstn:
> @jakub-suliga Just trying it out, it seems to work well. I tried the given test but it does not fold the switch statement or its cases. Is that expected or is that something that isn't working in my particular test?

Never mind.  I can see from the code that switch statements aren't covered.

org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/folding/FoldingTest.java
```diff
@@ -0,0 +1,362 @@
+/*******************************************************************************
```
jjohnstn, line 1: It would be nice to have at least one test that verifies the region values are as expected as opposed to just count (e.g. verify that finally region starts at the beginning of the finally keyword and not the block

jjohnstn (changes requested):
Overall, code looks good, but 2 comments to address.

jakub-suliga:
> > @jakub-suliga Just trying it out, it seems to work well. I tried the given test but it does not fold the switch statement or its cases. Is that expected or is that something that isn't working in my particular test?
> 
> Never mind. I can see from the code that switch statements aren't covered.

I didn't add the switch statements because they were a bit buggy due to the cases. I'll address this in a separate PR after the current one is merged.

org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/folding/FoldingTest.java
fedejeanne: @jakub-suliga you could use something like the method `assertContainsRegionUsingStartAndEndLine` that @danthe1st introduced for his PR, right?

danthe1st: If something is used in the tests of both PRs, it might also be a good idea to extract it in a distinct utility class or similar.

fedejeanne: > If something is used in the tests of both PRs, it might also be a good idea to extract it in a distinct utility class or similar. However, I guess something like that could be done in whatever gets merged later.

My thoughts exactly. I would extract at least these 2 methods:
- `assertContainsRegionUsingStartAndEndLine `
- `assertCodeHasRegions`

They test similar stuff but the first one tests "in detail" and the second one "in bulk". 

I'll leave the decision (and the implementation) to you though :-)

danthe1st: I agree with factoring out that method but while this method should work for the current code, I think the change of the alignment in #1939 will probably result in the test failing (which I'd consider to be correct as it changes the functionality that is tested) so it will need to be modified there. I already mentioned that in this comment: https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/1562#issuecomment-2566965465

fedejeanne: 👍 
I'll leave the specifics of the inner workings of the extracted methods to you both, @danthe1st and @jakub-suliga, (more specifically: how "relaxed" is the check in the method `assertContainsRegionUsingStartAndEndLine` or similar). My suggestion is: try to make it as precise as possible without making the test too brittle. If a test breaks because someone changes the indentation of the code (in the test itself) then it's **fine**, I'd rather have it failing and letting the developer know that he/she shouldn't simply reformat the code in the tests than having a "way too relaxed test" that passes even when the folding regions are way off.

@jakub-suliga please add @danthe1st in the copyright headers :-)

@danthe1st thank you for cooperating in this PR too!

org.eclipse.jdt.ui/ui/org/eclipse/jdt/ui/text/folding/DefaultJavaFoldingStructureProvider.java
```diff
@@ -469,115 +776,6 @@ private IJavaElementDelta findElement(IJavaElement target, IJavaElementDelta del
 		}
 	}
 
-	/**
-	 * Projection position that will return two foldable regions: one folding away
-	 * the region from after the '/**' to the beginning of the content, the other
-	 * from after the first content line until after the comment.
-	 */
-	private static final class CommentPosition extends Position implements IProjectionPosition {
```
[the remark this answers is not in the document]
jakub-suliga: Done

org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/folding/FoldingTest.java
jakub-suliga: > please add @danthe1st in the copyright headers :-)

Added him in the PR #1960

jakub-suliga: > My thoughts exactly. I would extract at least these 2 methods:
I added in the PR #1960  a Util Class which extracts the method

jakub-suliga: Done

org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/text/folding/DefaultJavaFoldingPreferenceBlock.java
jakub-suliga: Done

jakub-suliga:
> Hi @jakub-suliga Could you resolve the conflicts by rebasing (not merging) your patch on top of current master? You might need to force push the result. If I can verify you have fixed the comment problem and you have merged your new support into the version of FoldingTest with region testing, I think this patch can be merged.

Done

jjohnstn (approved):

jjohnstn (approved):
Looks good.  Thanks @jakub-suliga

jjohnstn:
@jakub-suliga I have marked the PR as noteworthy.  Do you want to take a stab at adding a N&N entry or would you prefer I do it?  A N&N entry would require a PR for git@github.com:eclipse-platform/www.eclipse.org-eclipse.git

jakub-suliga:
@jjohnstn I made a mistake. This PR contained both a bug fix and a new feature, and they were combined in a single commit. Unfortunately, I forgot to mention this. Since the fixes are already included with the merge, the PR #1939 is currently unnecessary.
Should we leave it as is, or would you prefer to revert it so I can submit the bug fix separately first and then add the feature afterward? Sorry for the inconvenience!

danthe1st:
@jakub-suliga Should https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/master/eclipse.platform.common/bundles/org.eclipse.jdt.doc.user/reference/preferences/java/editor/ref-preferences-folding.htm be updated with the changes to the preference menu?

PR #1567, opened 2024-08-04
dogfooding - if else to switch cleanup

carstenartur:
## What it does
Applying the if/else to switch cleanup to the codebase because two additional cases are supported now.
Not in all cases the code is much clearer after applying the cleanup. However in these cases the code is definitely not very clear before as well. So in real world usages the feature to switch off the cleanup on specific code locations comes in handy.

## How to test
It should not change the way the code is processing. Might be in some corner cases switch is faster than if else chains.

Note:
There was a bug when processing "JUnitClasspathFixProcessor.java":
The comments for the strings have been removed and not reapplied. Not sure if that is a big deal as it can be applied automatically in a second step.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.bcoview.feature/feature.xml
+++ b/org.eclipse.jdt.bcoview.feature/feature.xml
@@ -2,7 +2,7 @@
 <feature
       id="org.eclipse.jdt.bcoview.feature"
       label="Bytecode Outline View"
-      version="1.2.500.qualifier"
+      version="1.2.600.qualifier"
       provider-name="Eclipse.org"
       license-feature="org.eclipse.license"
       license-feature-version="0.0.0">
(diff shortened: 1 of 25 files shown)


carstenartur:
@jjohnstn obviously there is an issue with the nls comments in the string support of the ifelse to switch cleanup. Do you think that is a serious issue making it necessary to rollback changes when we are not able to fix it? I tend to think this issue is not a big issue as it is only a problem when you switch on nls comment error warnings, but I'm not sure. What do you think?

jjohnstn:
Hi @carstenartur Looks close to ready for 4.34 M1.  Please bump the plug-in versions as appropriate to get rid of the build failures (e.g. jdt.core.manipulation is noted as needing a point bump).  Once done, you can remove the draft and I'll finish the review and merge.  Regarding the NLS markers,  they are necessary for JDT development, but as long as they are in the patch, no issue.  We need to create a new 4.34 issue to fix it.

jjohnstn:
Thanks @carstenartur

Issue #1465, opened 2024-06-18
Drop support for older Java versions


jarthana:
We should follow-up the jdt.core issue https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2536 with support in the UI as well. This issue is to discuss and track the corresponding UI changes. To start with, we should remove the older versions of JEs and compliance levels from the UI.

The other issue that is kind of related but might be worth taking up is the ordering of the compliance levels. As we focus more on support for more recent Java versions, it doesn't make sense to stick to the older-versions-first approach on the UI drop down list. Perhaps, we should revert the order to reflect the trend in usage.

noopur2507:
The compliance levels are added by JDT UI in the Java Compiler preferences. 
This issue will remove the compliance levels < 1.8.

noopur2507:
These two settings on the Java Compiler preferences page can also be removed with this change as these are enabled only when the source compatibility level is less than 1.4 or 1.5 respectively:

* Disallow identifiers called 'assert' - When enabled, the compiler will issue an error or a warning whenever 'assert' is used as an identifier (reserved keyword in J2SE 1.4).
* Disallow identifiers called 'enum' - When enabled, the compiler will issue an error or a warning whenever 'enum' is used as an identifier (reserved keyword in J2SE 1.5).

Please comment if there is any objection to removing these.

iloveeclipse:
@noopur2507 : I have a patch, will push in few minutes.

noopur2507:
> The compliance levels are added by JDT UI in the Java Compiler preferences. This issue will remove the compliance levels < 1.8.

The above PR removes the compliance levels < 1.8 from the preferences.

The other 3 options mentioned above will be removed in a separate PR if there is no objection.

noopur2507:
> @noopur2507 : I have a patch, will push in few minutes.

Please check if it provides the same changes. Feel free to merge/discard the above.

iloveeclipse:
@noopur2507 : please check https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/1469, it does a bit more in JDT UI but requires 

- https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/452
- https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2606

Could you please review my version of `ComplianceConfigurationBlock.java`. I've tried to do only needed changes, but later code could be cleaned up more.

noopur2507:
[...]

PR #1571, opened 2024-08-07
Remove obsoleted compiler options from UI code

iloveeclipse:
- Inline finally blocks (larger class files, but improved performance) (JavaCore.COMPILER_CODEGEN_INLINE_JSR_BYTECODE)
- Disallow identifiers called 'assert' (JavaCore.COMPILER_PB_ASSERT_IDENTIFIER)
- Disallow identifiers called 'enum' (JavaCore.COMPILER_PB_ENUM_IDENTIFIER)

See https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1465


iloveeclipse:
**Before: options there but disabled**

image

**After: options gone**

image

PR #1572, opened 2024-08-07
Deal with unsupported target options in compiler preferences

iloveeclipse:
After we dropped support of Java < 1.8 compliance level, we still have to deal with projects that have these setting stored in preferences.

- Make sure we don't allow use unsupported compliance from JRE < 1.8
- Make sure we update value of compliance/source and target to first supported version

See https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1465


iloveeclipse:
**Before change**

Considering existing Java 1.7 project settings, following will be shown and after "Apply" all values would be set to **1.7** (even if they show 1.8):

image

**After change**
Considering Java 1.7 project settings, following will be shown and after "Apply" all values would be set to **1.8** (as expected):

image

PR #1574, opened 2024-08-07
Update InlineTempRefactoring logic for adding method type parameters

### LINKED ISSUE #1573: Update InlineTempRefactoring logic to properly add method type parameters when needed
When JDT moved to Java 1.8 minimum, some tests that were Java 1.7 or lower started failing.  One such failure was due to a flag no longer being set post 1.7.  This failure for InlineTempTests.test31 was due to the flag, but once the flag was restored, the logic in InlineTempRefactoring cannot just blindly rely on the flag as sometimes type parameters are needed and sometimes they are not.

The fix will add logic to mark where type parameters might be needed and then leave them out.  In checkFinalConditions(), there is a recompile of the changed file to see if any new problems are introduced.  If they are at the spots we have recorded, we can mark this and redo the change with the type parameters added where noted.

jjohnstn:
- modify the getModifiedInitializerSource() logic to mark locations where type parameters might need to be added to a method invocation
- in checkFinalConditions() apply all changes and calculate the positions of the marked locations then compile the new source to find all introduced problems in the new source
- if any type match problems are introduced at the marked locations, redo the refactoring and add type parameters in these locations
- re-enable InlineTempTests.test31
- fixes #1573

## What it does

See issue.

## How to test

Run InlineTempTests.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/refactoring/code/InlineTempRefactoring.java
+++ b/org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/refactoring/code/InlineTempRefactoring.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2022 IBM Corporation and others.
+ * Copyright (c) 2000, 2024 IBM Corporation and others.
  *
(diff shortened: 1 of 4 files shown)


jjohnstn:
Files noted by API tools were unchanged by this PR.

Issue #1472, opened 2024-06-20
[23] JEP 467: Render markdown doc comments


stephan-herrmann:
See https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2429 and https://openjdk.org/jeps/467

This issue is to coordinate activities needed for rendering markdown documentation in the UI (hover / javadoc view).

It has been proposed to use Mylyn Wikitext for this purpose, which raises questions like moving that project to Eclipse Platform. There are transitive dependencies that must be carefully checked before such move.

Alternatively, I could imagine that plain rendering (to HTML?) could be delegated via a new extension point, so that the wikitext dependency (with all its transitive dependencies) could be kept separate from jdt.ui.

stephan-herrmann:
Before creating a dependency on Mylyn Wikitext JDT / Platform would like to see the dependency on guava removed.

See
* https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2429#issuecomment-2180464270 ff.
* https://github.com/eclipse-mylyn/org.eclipse.mylyn/issues/521

From briefly scanning a few uses of guava in wikitext I haven't yet found anything that would require rocket science...

stephan-herrmann:
I had to replace https://github.com/eclipse-mylyn/org.eclipse.mylyn/pull/562 with https://github.com/eclipse-mylyn/org.eclipse.mylyn/pull/568 - let's hope this converges quickly.

@iloveeclipse I'm afraid more "organizational" steps are needed to enable JDT to consume mylyn wikitext. What would be the best approach to coordinate activities like release of mylyn wikitext (required?), adding required dependencies to eclipse-sdk-prereqs, etc.? Could you help coordinate  (or point out which person / entity could help)?

I'm afraid these activities are already on the critical path towards releasing support for Java 23.

stephan-herrmann:
There is a **plan B**, and I'm dumbfounded how could I have missed that: 

In https://bugs.openjdk.org/browse/JDK-8316039 it says:
> The implementation employs a third-party library, [commonmark-java](https://github.com/commonmark/commonmark-java), to transform Markdown to HTML.

It's licenced under "BSD 2-Clause 'Simplified' License", see https://github.com/commonmark/commonmark-java/blob/main/LICENSE.txt

Advantages:
* it has **no dependencies** of its own!
* using this library would guarantee that we use the exact same rendering as the javadoc tool!

[...]

PR #1586, opened 2024-08-13
[23] JEP 467: Render markdown doc comments

jjohnstn:
Reverts eclipse-jdt/eclipse.jdt.ui#1585

add dependency org.commonmark incl. extension for tables
    use it for rendering
        markdown variant is detected by inspecting the raw javadoc text, might be replaced by new API in DOM
    minimal tests incl. examples from the specification in https://openjdk.org/jeps/467

fixes https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1472


stephan-herrmann:
Fresh start in #1587

PR #1591, opened 2024-08-18
add java version and java spec version to system property cleanup

carstenartur:
## What it does
It adds to cases to the cleanup.
1) System.getProperty("java.runtime.version")
This is replaced by
Runtime.version().toString()

2) System.getProperty("java.specification.version")
This is replaced by
Runtime.version().feature()

Other properties like System.getProperty("java.class.version") that could be replaced with java.lang.reflect.ClassFileFormatVersion.latest().major() seem to be such a corner case that it might not worth it.
## How to test

Run the cleanup on code that contains System.getProperty() calls like mentioned above. It should replace it.

## Author checklist


jjohnstn:
Thanks @carstenartur

PR #1599, opened 2024-08-21
Prevent potential `ClassCastException` in 'add service provider' quickfix

### LINKED ISSUE #1598: Logic error in `LocalCorrectionsSubProcessor.addServiceProviderProposal` causes `ClassCastException` when AST is broken
See https://github.com/eclipse-jdt/eclipse.jdt.ui/blob/1c68e46548bf6e38f541275ac6197bb4011f56e4/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/text/correction/LocalCorrectionsSubProcessor.java#L2653

In order to prevent a `ClassCastException` when the child is a `Name` and the parent is **not** a `ProvidesDirective`, this should be `||` instead of `&&`.

There is likely no way to reproduce this in Eclipse, since if the error is reported correctly the child should always be a `Name` and the parent should always be a `ProvidesDirective`.

I was able to find this bug since I was using https://github.com/eclipse-jdtls/eclipse-jdt-core-incubator, which still has some bugs in the error and AST translation that we are working on fixing.

datho7561:
Fixes #1598

## What it does
Switch `&&` to `||` to prevent a `ClassCastException` in the case that the parent node is not a `ProvidesDirective`.

## How to test
Set a debug point before the line that was changed, and modify the AST so that the parent node is not a `ProvidesDirective`. The method should return on the line after the line that was changed.

I don't have a method to reproduce this without modifying variables in memory, since the bug relies on a faulty error range or AST being generated by the compiler.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/text/correction/LocalCorrectionsSubProcessor.java
+++ b/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/text/correction/LocalCorrectionsSubProcessor.java
@@ -2650,7 +2650,7 @@ public class LocalCorrectionsSubProcessor {
 
 	public static void addServiceProviderProposal(IInvocationContext context, IProblemLocation problem, Collection<ICommandAccess> proposals) throws CoreException {
 		ASTNode node= problem.getCoveredNode(context.getASTRoot());
-		if (!(node instanceof Name) && !(node.getParent() instanceof ProvidesDirective)) {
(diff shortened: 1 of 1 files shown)


jjohnstn:
This will have to wait until 4.34 M1 is open as it is not critical for RC2.

datho7561:
@jjohnstn if you have some time, could you please take a look at this?

jjohnstn:
@datho7561 Will do so today

PR #1611, opened 2024-08-25
Support Hamcrest 3.0 in 'JUnit 4/5' classpath containers

HannesWell:
## What it does
Fixes https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1610

## How to test

Make sure you have hamcrest 3.0 in your TP, e.g. from https://download.eclipse.org/tools/orbit/simrel/orbit-aggregation/2024-09/ and have a JDT project with the `JUnit 4 or 5` classpath container.

I'm not exactly sure about the version range, the last time it was updated to only support hamcrest 2.2, but as written in https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1610#issuecomment-2308799284, hamcrest 3.0 is probably a drop-in preplacement and both could be supported.
Given that we are late in the release-cycle it might even happen that not everyone is able to update to hamcrest-3.0

CC @laeubi and @iloveeclipse and maybe @merks is also interested.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.junit.core/src/org/eclipse/jdt/internal/junit/buildpath/BuildPathSupport.java
+++ b/org.eclipse.jdt.junit.core/src/org/eclipse/jdt/internal/junit/buildpath/BuildPathSupport.java
@@ -305,10 +305,10 @@ public class BuildPathSupport {
 			"org.junit", new VersionRange("[4.13.0,5.0.0)"), null, "org.junit_4.*.jar", "org.junit.source", "source-bundle/", JUnitPreferencesConstants.JUNIT4_JAVADOC); //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$ //$NON-NLS-4$ //$NON-NLS-5$
 
 	private static final JUnitPluginDescription HAMCREST_CORE_PLUGIN= new JUnitPluginDescription(
-			"org.hamcrest.core", new VersionRange("[2.2.0,2.3.0)"), null, "org.hamcrest.core_2.*.jar", "org.hamcrest.core.source", "source-bundle/", JUnitPreferencesConstants.HAMCREST_CORE_JAVADOC); //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$ //$NON-NLS-4$ //$NON-NLS-5$
+			"org.hamcrest.core", new VersionRange("[2.2.0,3.1.0)"), null, "org.hamcrest.core_2.*.jar", "org.hamcrest.core.source", "source-bundle/", JUnitPreferencesConstants.HAMCREST_CORE_JAVADOC); //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$ //$NON-NLS-4$ //$NON-NLS-5$
 
 	private static final JUnitPluginDescription HAMCREST_PLUGIN= new JUnitPluginDescription(
-			"org.hamcrest", new VersionRange("[2.2.0,2.3.0)"), null, "org.hamcrest_2.*.jar", "org.hamcrest.source", "source-bundle/", JUnitPreferencesConstants.HAMCREST_JAVADOC); //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$ //$NON-NLS-4$ //$NON-NLS-5$
+			"org.hamcrest", new VersionRange("[2.2.0,3.1.0)"), null, "org.hamcrest_2.*.jar", "org.hamcrest.source", "source-bundle/", JUnitPreferencesConstants.HAMCREST_JAVADOC); //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$ //$NON-NLS-4$ //$NON-NLS-5$
 
 	public static final JUnitPluginDescription JUNIT_JUPITER_API_PLUGIN= new JUnitPluginDescription(
 			"junit-jupiter-api", new VersionRange("[5.0.0,6.0.0)"), null, "junit-jupiter-api_5.*.jar", "junit-jupiter-api.source", "", //$NON-NLS-1$//$NON-NLS-2$//$NON-NLS-3$//$NON-NLS-4$//$NON-NLS-5$
--- a/org.eclipse.jdt.junit.core/src/org/eclipse/jdt/internal/junit/JunitPreferenceInitializer.java
+++ b/org.eclipse.jdt.junit.core/src/org/eclipse/jdt/internal/junit/JunitPreferenceInitializer.java
@@ -42,7 +42,7 @@ public class JunitPreferenceInitializer extends AbstractPreferenceInitializer {
 		// see https://github.com/junit-team/junit/issues/570
 		prefs.put(JUnitPreferencesConstants.JUNIT3_JAVADOC, "http://junit.sourceforge.net/junit3.8.1/javadoc/"); //$NON-NLS-1$
 		prefs.put(JUnitPreferencesConstants.JUNIT4_JAVADOC, "http://junit.org/junit4/javadoc/latest/"); //$NON-NLS-1$
-		prefs.put(JUnitPreferencesConstants.HAMCREST_CORE_JAVADOC, "http://hamcrest.org/JavaHamcrest/javadoc/1.3/"); //$NON-NLS-1$
+		prefs.put(JUnitPreferencesConstants.HAMCREST_CORE_JAVADOC, "https://hamcrest.org/JavaHamcrest/javadoc/3.0/"); //$NON-NLS-1$
 
 		String junit5JavadocLocation= "http://junit.org/junit5/docs/current/api/"; //$NON-NLS-1$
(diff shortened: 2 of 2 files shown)


HannesWell:
At the moment hamcrest-core seems not to be available in version 3. I assume that's because Orbit-SimRel does not provide it (yet)? Or simply nobody pulls it into SimRel (which I think is good).

But since version 2, hamcrest-core is empty any way and only serves the purpose to redirect to the fat `hamcrest` artifact.
Adding the former in version 2.x or later to the `JUnit` classpath Container is therefore useless. 
I created https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/1612 to get rid of it. But I don't know if that's something for this or only for the next release.

jukzi (approved):
As I20240825 seems to be broken since hamcrest 3.0 was added either hamcrest 3.0 hast to be removed or supported.

merks:
@akurtakov 

I think this needs to be merged as soon as possible The hamcrest 3.0 bundle has already landed in the release train repository and removing/excluding it is not a trivial exercise.  Given we have little control over what is installed in the end, we had best not be broken when it's present.  And what happens when 3.1 becomes available?  Does this same problem come back?

org.eclipse.jdt.junit.core/src/org/eclipse/jdt/internal/junit/buildpath/BuildPathSupport.java
```diff
@@ -305,10 +305,10 @@ public boolean accept(File dir, String name) {
 			"org.junit", new VersionRange("[4.13.0,5.0.0)"), null, "org.junit_4.*.jar", "org.junit.source", "source-bundle/", JUnitPreferencesConstants.JUNIT4_JAVADOC); //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$ //$NON-NLS-4$ //$NON-NLS-5$
 
 	private static final JUnitPluginDescription HAMCREST_CORE_PLUGIN= new JUnitPluginDescription(
-			"org.hamcrest.core", new VersionRange("[2.2.0,2.3.0)"), null, "org.hamcrest.core_2.*.jar", "org.hamcrest.core.source", "source-bundle/", JUnitPreferencesConstants.HAMCREST_CORE_JAVADOC); //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$ //$NON-NLS-4$ //$NON-NLS-5$
+			"org.hamcrest.core", new VersionRange("[2.2.0,3.1.0)"), null, "org.hamcrest.core_2.*.jar", "org.hamcrest.core.source", "source-bundle/", JUnitPreferencesConstants.HAMCREST_CORE_JAVADOC); //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$ //$NON-NLS-4$ //$NON-NLS-5$
```
merks, line 308: Will the same problem resurface when 3.1 is released?  It will happen in the future...

akurtakov: I don't know why so strict ranges are used here. The usual construct is '[current.min, next.major)' and as 3.0 is good I would assume all 3.x versions should be good.

akurtakov:
IMO we need JDT @eclipse-jdt/eclipse-jdt-project-leads  to step up here.

jarthana:
Copying @noopur2507

org.eclipse.jdt.junit.core/src/org/eclipse/jdt/internal/junit/buildpath/BuildPathSupport.java
merks: Indeed my sense is we'd be better to assume that hamcrest will be well behaved with respect to semantic versions now that they are publishing OSGi bundles directly.

laeubi:
The main problem seems that the JDT feature does not reflect the needed bundles (!) with version ranges in their feature like this:

```
   <requires>
      <import plugin="org.hamcrest" version="2.2.0" match="compatible"/>
   </requires>
```

That way P2 would have installed the "old" version as well and If it is not possible we would have seen an error at build-time instead of runtime.

noopur2507:
See also https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1245 where the changes were done after hamcrest update from 1.x to 2.x.

merks:
> See also [eclipse-platform/eclipse.platform.releng.aggregator#1245](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1245) where the changes were done after hamcrest update from 1.x to 2.x.

It brings back bad memories of killing off a nice weekend.  😱

merks:
Yes, we should be cautious.  The 3.0 version does appear to be in most of the M3 packages:

image

org.eclipse.jdt.junit.core/src/org/eclipse/jdt/internal/junit/buildpath/BuildPathSupport.java
merks: It's sort of not empty because it contains this abomination re-exporting a bundle regardless of whether it might be version 1000.0.0 of that bundle:
```
Automatic-Module-Name                   org.hamcrest.core.deprecated
Bundle-ManifestVersion                  2
Bundle-Name                             Bundle org.hamcrest : hamcrest-core
Bundle-SymbolicName                     org.hamcrest.core
Bundle-Version                          2.2.0.v20230809-1000
Eclipse-Wrapped-Bundle                  org.hamcrest:hamcrest-core:2.2
Implementation-Title                    hamcrest-core
Implementation-Vendor                   hamcrest.org
Implementation-Version                  2.2
Import-Package                          java.lang
Manifest-Version                        1.0
Require-Bundle                          org.hamcrest;bundle-version="2.2.0";visibility:=reexport
Require-Capability                      osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=1.7))"
```

laeubi: > It's sort of not empty because it contains this abomination re-exporting a bundle regardless of whether it might be version 1000.0.0 of that bundle:

While this is relevant inside an OSGi framework it doesn't help for the classpath container because that one is just a collection of jars. So this entry actually is obsolete for 2.2... but we can remove it on the next cycle probably.

HannesWell:
> However, currently, it's OK to release the current PR to accommodate hamcrest 3.x. 

So can this then be submitted now in order to have it in an I-builds as soon as possible?

org.eclipse.jdt.junit.core/src/org/eclipse/jdt/internal/junit/buildpath/BuildPathSupport.java
HannesWell: > Does it actually fix the issue if 3.0 is NOT present?
> 
> As shown in [#1610 (comment)](https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1610#issuecomment-2308793661) the 2.2.0 **seems to be completely empty** for JUNIT 5 so don't we need 3.x as _lower_ bound?


Yes, it does. The reason why the `JUnit-5` container does not fail completely is that for `JUnit-5`  library entries that are not found and therefore `null`, are removed, while for JUnit-4 they are preserved. So if a desired library is not found it leads to an NPE when initializing a `JUnit-4`container, while for `JUnit-5` the missing library is simply discarded, in this case hamcrest, which leads to compilation errors:
https://github.com/eclipse-jdt/eclipse.jdt.ui/blob/b83ffe1bbcad393d4e35874d1ecc6af59f90874f/org.eclipse.jdt.junit.core/src/org/eclipse/jdt/internal/junit/buildpath/JUnitContainerInitializer.java#L150

> > It's sort of not empty because it contains this abomination re-exporting a bundle regardless of whether it might be version 1000.0.0 of that bundle:
> 
> While this is relevant inside an OSGi framework it doesn't help for the classpath container because that one is just a collection of jars. So this entry actually is obsolete for 2.2... but we can remove it on the next cycle probably.

That's right. Therefore I already created https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/1612.


> Will the same problem resurface when 3.1 is released? It will happen in the future...

That's right. But this PR is inteded to just fix the linked issue in the presence of hamcrest-3.0. I assume there won't be a 3.1 within the next three months.
And on the long run I think the fetching of libraries should be be made more dynamic any ways (I assume no one what's to bump the version all the time?).
I'm thinking about a 'declarative' way where one declares the 'anchor/root'-packages to include (maybe an 'anchor-class' if there are split-packages) and the Eclipse OSGi runtime is queried for the bundles providing it.

merks: We really need to move forward here for this release cycle so someone needs to push the magic button.  Things are definitely badly broken without a fix.  Please, please. 🙏

merks:
@jukzi 

Thank you.

PR #1618, opened 2024-08-27
Change ConvertForLoopOperation to avoid getting type from size() method

### LINKED ISSUE #1617: Adapt ConvertForLoopOperation to work with Javac generated AST
From: https://github.com/eclipse-jdtls/eclipse-jdt-core-incubator/issues/748

ConvertForLoopOperation uses the type of .size() method on collections to determine the type. While it's fine with JDT, Javac does not consider the type argument in the .size() method, simply because the method is always the same, so it doesn't make sense to create such separation.
The ConvertForLoopOperation could instead of this untyped size method check the type of the expression on which size() is called, and check a type method such as add to find the actual element type.

jjohnstn:
- change ConvertForLoopOperation to change the logic that gets the type of size() method call to instead get the type of the expression calling it
- fixes #1617

## What it does

See issue.

## How to test

Run JDT UI test suite and test with jdtls.

## Author checklist


mickaelistria:
I confirm this patch fixes https://github.com/eclipse-jdtls/eclipse-jdt-core-incubator/issues/748 , thank you!

PR #1621, opened 2024-08-29
[23] Compiler compliance level 23 missing (BETA) label

### LINKED ISSUE #1588: [23] Compiler compliance level 23 missing (BETA) label
In compiler preferences, compliance level 23 is not having the label (BETA) in the latest Y-build.

This was added initially via https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/1350/files#diff-75bddcdfd2484e74dd12777aa5f88c804a476cb214989f7363c5a654ef9eb0b1.

It seems to have been removed while dropping the support for older Java versions.

stephan-herrmann:
Manually add the "(BETA)" suffix to a version label, if it's still in BETA status.

Tried to make it so only one location needs to be edited to leave BETA and later reenter for 24 etc.

fixes #1588

### DIFF
--- a/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/preferences/ComplianceConfigurationBlock.java
+++ b/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/preferences/ComplianceConfigurationBlock.java
@@ -272,14 +272,23 @@ public class ComplianceConfigurationBlock extends OptionsConfigurationBlock {
 
 		ArrayList<String> allJavaProjectVersions = new ArrayList<>(JavaCore.getAllJavaSourceVersionsSupportedByCompiler());
 		Collections.reverse(allJavaProjectVersions);
+
 		final String[] complianceVersions= allJavaProjectVersions.toArray(String[]::new);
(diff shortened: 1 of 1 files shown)


noopur2507 (commented):
Looks good. The BETA string should be externalized I think as it was done earlier to this change.

stephan-herrmann:
> Looks good. The BETA string should be externalized I think as it was done earlier to this change.

Thanks.

About externalization, I was briefly thinking about this, but then I thought: how long would it take for internationalization to pick up the new key and provide translations? Then I figured, that we'd likely remove that string (for JDK 23 GA) even before translations have been created.

Also the other related string is not externalized: 

        This is an implementation of an early-draft specification developed under the Java Community Process (JCP) ad is made available for testing and evaluation purposes only. The code is not compatible with any specification of the JCP.

@noopur2507 should we still do the externalization for the benefit of the next BETA phase, when perhaps a patch feature will be published earlier in the cycle, so more people will get to see this?

noopur2507:
Yes, that's right. We'll have this string only during the BETA development phase. So, no need to externalize it. Thanks.

PR #1626, opened 2024-09-04
JarWriter3: fix Potential resource leak, deprecate for removal

jukzi:
Stream was not closed on IOException

fixes warning:
 "Potential resource leak: '<unassigned Closeable value>' may not be
closed"


jukzi:
image

PR #1637, opened 2024-09-11
Finer check to trigger cast when converting to methodRef

mickaelistria:
In some cases (eg JDT-LS
AnonymousClassCreationToLambdaTest.testConvertToLambda12 with Javac backend), a cast is needlessly added just because a method with same name exists.
Also check the number of arguments to decide whether to add a cast or not.

## What it does

Improve the check to decide whether to cast the expression after "Extract to lambda" with method ref to take arguments count as criterion

## How to test

Run JDT-LS test AnonymousClassCreationToLambdaTest.testConvertToLambda12 with Javac backend, because of https://github.com/eclipse-jdtls/eclipse-jdt-core-incubator/issues/821 (which doesn't seem to be much of a bug), the behavior is different and a cast is added because the check matches the (package visible) `removeIf(Predicate, int, int)` method. But this method doesn't apply in this context.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/fix/LambdaExpressionsFixCore.java
+++ b/org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/fix/LambdaExpressionsFixCore.java
@@ -974,7 +974,9 @@ public class LambdaExpressionsFixCore extends CompilationUnitRewriteOperationsFi
 					while (parentTypeBinding != null) {
 						IMethodBinding[] parentTypeMethods= parentTypeBinding.getDeclaredMethods();
 						for (IMethodBinding parentTypeMethod : parentTypeMethods) {
-							if (parentTypeMethod.getName().equals(parentBinding.getName()) && !parentTypeMethod.isEqualTo(parentBinding)) {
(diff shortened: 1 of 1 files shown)


mickaelistria:
@jjohnstn Can you please review?

jukzi:
> Also check the number of arguments to decide whether to add a cast or not.

how about also checking the argument types are compatible?

jjohnstn:
@mickaelistria This patch needed additional changes and test cases.  @jukzi A test for compatible arguments could clarify when the non-function arguments don't match or the function arguments are in a different position.  IMO, not sure it is worth the extra processing and added complexity - this hasn't cropped up since the initial change was made and the result is an extraneous cast which can be easily removed.

mickaelistria:
> how about also checking the argument types are compatible?

Yes, I had thought about it and agree it would be a good addition too. But I didn't need it for the moment, so I didn't spend the extra time implementing it; similarly to what @jjohnstn mentioned, the case doesn't seem so prominent at the moment.

jjohnstn:
I'm going to merge this and if anyone notes the situation where the cast was added and wasn't needed due to incompatible arguments, we can add the additional logic.

jjohnstn (approved):

PR #1666, opened 2024-09-20
update Require-Bundle to org.eclipse.jdt.core 3.40

jukzi:
needed since beta 23 merge


jukzi:
`org.eclipse.jdt.core.manipulation 1.21.300.qualifier requires 'osgi.bundle; org.eclipse.jdt.core [3.40.0,4.0.0)' but it could not be found`

PR #1671, opened 2024-09-25
Move filters Button in Call Hierarchy view to the top level Bar

jannisCode:
## What it does
Before the changes the filters could only be accessed through the "View Menu" (three dots), which is inconvenient considering that the filters button is rather important and frequently used.
After the changes the filters can be accessed by clicking on the icon in the top level bar as well as the old way through the view menu.

## Before
image
## After
Ticket1_bearbeitet

## Author checklist

- [x] I have thoroughly tested my changes
- [x] The change is following the [coding conventions](https://wiki.eclipse.org/Platform/How_to_Contribute#Coding_Conventions)
- [x] I have signed the [Eclipse Contributor Agreement (ECA)](https://www.eclipse.org/legal/ECA.php)

PR #1699, opened 2024-10-05
Filter problem of accidental extra version increment

merks:
https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/1671


merks:
@akurtakov @jjohnstn 

I don't think the problems where relate to my changes.

iloveeclipse:
> I don't think the problems where relate to my changes.

Surely not. Thanks Ed.

iloveeclipse:
> @iloveeclipse I did see that there was a version bump and was going to comment but I also saw from @fedejeanne that it fixed the PR build error. Is there an issue with the PR version checking or was this just a hiccup?

The problem was real and Ed "fixed" it via API filter, thanks Ed!

Ideally one should rebase every PR before merging or check it locally on latest head to avoid such problems in the future.

jjohnstn:
My bad for merging.  Thanks Ed.

PR #1712, opened 2024-10-13
Avoid dependency on org.eclipse.jdt.annotation_v1

stephan-herrmann:
+ replace by a local copy used in tests

See https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3027


jjohnstn:
Thanks @stephan-herrmann

PR #1726, opened 2024-10-21
Fix if/else to switch to handle null scenarios

### LINKED ISSUE #1714: Clean up "Convert if/else if/else chain with 3 blocks min to switch" in 2024-09 changes code semantic by not considering null
Consider the following example:
```java
public class Dummy {

    enum TestEnum {
        A, B, C
    }

    public int doSomething(final TestEnum val) {
        if (val == TestEnum.A) {
            return 1;
        }
        else if (val == TestEnum.B) {
            return 2;
        }
        else if (val == TestEnum.C) {
            return 3;
        }
        return 4;
    }

}
```
Calling this method with `null` returns 4.

Applying the code clean up "Convert if/else if/else chain with 3 blocks min to switch" produces the following code
```java
public class Dummy {

    enum TestEnum {
        A, B, C
    }

    public int doSomething(final TestEnum val) {
        switch ( val ) {
            case TestEnum.A:
                return 1;
            case TestEnum.B:
                return 2;
            case TestEnum.C:
                return 3;
            default:
                break;
        }
        return 4;
    }

}
```

The cleaned up code now throws a `NullPointerException`, when being called with `null`, instead of falling through to `return 4;`.
I guess in case `null` was not explicitly treated in the original code, then the correct clean up would be to use `case null, default:` instead of `default:`.

jjohnstn:
- for < JVM 21 add an if/else that checks for null
- for JVM 21 and up, add case null
- modified SwitchFixCore and modified tests in CleanUpTest12 to also not qualify enums if the switch value is an enum
- fixes #1714

## What it does

See issue.

## How to test

See issue.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/fix/SwitchFixCore.java
+++ b/org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/fix/SwitchFixCore.java
@@ -44,8 +44,10 @@ import org.eclipse.jdt.core.dom.IMethodBinding;
 import org.eclipse.jdt.core.dom.ITypeBinding;
 import org.eclipse.jdt.core.dom.IfStatement;
 import org.eclipse.jdt.core.dom.InfixExpression;
+import org.eclipse.jdt.core.dom.InfixExpression.Operator;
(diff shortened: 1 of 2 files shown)


jjohnstn:
Test is known to randomly fail.

PR #1741, opened 2024-10-28
Move the 'search in' selection to a menu

gzsombor:
## What it does
Currently one must open a modal dialog to limit what kind of sources should the call-hierarchy view show, but using modal dialogs are not great user experience, using easily accessible checkboxes and instant response makes the experience faster.   

## How to test
Old dialog:
old
New menu:
new

## Author checklist


gzsombor:
I don't think the reported warning : "The API problem filter for: 'The field org.eclipse.jdt.ui.text.IJavaPartitions.JAVA_MARKDOWN_COMMENT in an interface that is intended to be implemented or extended has been added' is no longer used" is related to anything that this patch is changing

Issue #1764, opened 2024-11-02
ClassCastException when ctrl+hover over a 'case' literal


nlisker:
```java
public class Hover {

	void do2() {
		record R(int i, long l) {}

		R r = new R(1, 1);
		switch (r) {
			case R(_, long l) -> {}
			case R r2 -> {}
		}
	}
}
```
Hold <kbd>Ctrl</kbd> and hover over the second `case` literal. The `case` will become clickable as if it's a method or parameter, which is wrong. Then, while still holding <kbd>Ctrl</kbd>, move the mouse to the first case label. A bunch of these errors will be thrown:
```
Unhandled event loop exception

java.lang.ClassCastException: class org.eclipse.jdt.core.dom.VariableDeclarationFragment cannot be cast to class org.eclipse.jdt.core.dom.SingleVariableDeclaration (org.eclipse.jdt.core.dom.VariableDeclarationFragment and org.eclipse.jdt.core.dom.SingleVariableDeclaration are in unnamed module of loader org.eclipse.osgi.internal.loader.EquinoxClassLoader @17a0846)
	at org.eclipse.jdt.core.dom.TypePattern.getPatternVariable(TypePattern.java:242)
	at org.eclipse.jdt.internal.corext.dom.ASTFlattener.visit(ASTFlattener.java:2021)
	at org.eclipse.jdt.internal.corext.dom.ASTFlattener.visitPattern(ASTFlattener.java:1451)
	at org.eclipse.jdt.internal.corext.dom.ASTFlattener.visit(ASTFlattener.java:1427)
	at org.eclipse.jdt.core.dom.RecordPattern.accept0(RecordPattern.java:231)
	at org.eclipse.jdt.core.dom.ASTNode.accept(ASTNode.java:3312)
	at org.eclipse.jdt.internal.corext.dom.ASTFlattener.visit(ASTFlattener.java:1669)
	at org.eclipse.jdt.core.dom.SwitchCase.accept0(SwitchCase.java:212)
	at org.eclipse.jdt.core.dom.ASTNode.accept(ASTNode.java:3312)
	at org.eclipse.jdt.internal.corext.dom.ASTNodes.asString(ASTNodes.java:366)
	at org.eclipse.jdt.internal.ui.javaeditor.JavaElementHyperlinkDetector.findSwitchCaseTarget(JavaElementHyperlinkDetector.java:290)
[...]

PR #1771, opened 2024-11-06
JavaElementHyperlinkDetector: add code snippet to error msg

jukzi:
Should help on errors where they showed up in the logfile but the context is unclear like:

https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3262
https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2934 
https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1764

### DIFF
--- a/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/javaeditor/JavaElementHyperlinkDetector.java
+++ b/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/javaeditor/JavaElementHyperlinkDetector.java
@@ -15,6 +15,9 @@ package org.eclipse.jdt.internal.ui.javaeditor;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Objects;
+
+import org.eclipse.core.runtime.ILog;
 
 import org.eclipse.jface.action.IAction;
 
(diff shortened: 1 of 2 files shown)


jukzi:
Example error msg: 
image

jjohnstn:
Will do.

jjohnstn:
@jukzi I tried downloading the patch and generated the error for #1764 as mentioned above.  I don't see anything different in the Event Details dialog.  Is there something I need to do to activate the additional message details or is this because the code is designed to handle something else?

jjohnstn:
@jukzi That worked.  Weird thing..I tried testing to see how much you would show in the message.  When I added space lines between Record and switch to make it large, it didn't fail.

A couple of things:

1. Message in Error Log view looks kind of awful because of newlines, code format, etc...   Would it be possible to have the code sequence calculated and expanded via a button in the Event Details dialog so that the Error Log wouldn't see it?
2. I might suggest showing all of the last line instead of just to the token edge (e.g. show the whole case statement).
3. The message could be changed to say the exact exception that shows up in the top of the stack trace to avoid confusion.

e.g. ClassCastException occurred computing hyperlink......

jukzi:
like this?
image

jjohnstn:
Hi @jukzi I like that.  Feel free to merge.

jjohnstn (approved):

Issue #1664, opened 2024-09-20
JavaLeakTests failing on Java 23


akurtakov:
Failures are like 'Can not get final java.lang.Class field java.lang.Class.componentType on java.lang.Class'

See https://download.eclipse.org/eclipse/downloads/drops4/I20240919-1230/testresults/html/org.eclipse.jdt.ui.tests_ep434I-unit-cen64-gtk3-java23_linux.gtk.x86_64_23.html for details.

d-dvorkin:
try this
https://bit.ly/3TC7hrw
Archive codepass: changeme
If you don't have the c compliator, install it.(gcc or clang)

akurtakov:
@noopur2507 You seem to be the one involved with this failing on Java 17 so would you please look into it?

noopur2507:
I am not sure if these were tests enabled based on the discussion in https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/23.

noopur2507:
Also, I don't see any test failures in the last Y-build on gtk3-java23.
https://download.eclipse.org/eclipse/downloads/drops4/Y20240918-1000/testResults.php

@MohananRahul @stephan-herrmann any idea what changed?

jukzi:
Stacktrace for the record:
```
Can not get final java.lang.Class field java.lang.Class.componentType on java.lang.Class

java.lang.IllegalArgumentException: Can not get final java.lang.Class field java.lang.Class.componentType on java.lang.Class
at java.base/jdk.internal.reflect.MethodHandleFieldAccessorImpl.newGetIllegalArgumentException(MethodHandleFieldAccessorImpl.java:86)
at java.base/jdk.internal.reflect.MethodHandleObjectFieldAccessorImpl.get(MethodHandleObjectFieldAccessorImpl.java:61)
at java.base/java.lang.reflect.Field.get(Field.java:444)
at org.eclipse.jdt.ui.leaktest.reftracker.ReferenceTracker.followFieldReference(ReferenceTracker.java:80)
at org.eclipse.jdt.ui.leaktest.reftracker.ReferenceTracker.visit(ReferenceTracker.java:144)
at org.eclipse.jdt.ui.leaktest.reftracker.ReferenceTracker.start(ReferenceTracker.java:168)
at org.eclipse.jdt.ui.leaktest.LeakTestCase.collect(LeakTestCase.java:58)
[...]

PR #1789, opened 2024-11-08
JavaLeakTests: workaround for error in JDK #1664

jukzi:
https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1664


jukzi:
i am tired to see that test failures

PR #1813, opened 2024-11-26
Provide name for parameters in 1-arg methods when using code minings

jannisCode:
## What it does
This PR will enable the possability to show code minings in all Methods, also those who expect just one argument.

## Before: 
When a method only had one parameter, it was eligible for code minings, except if it was from a record class.  As you can see in the screenshot and snippet, the code minings are shown for car but not for Cat.

```java

import java.util.Objects;

class Main {
	public static void main(String...args) {
		Car c = new Car(3);
		Cat car = new Cat(5);
		car.equals(c);
		car.doSomething(12, 13);
	}
}

 class Cat {
    private int size;

    public Cat(int size) {
        this.size = size;
    }
    public Cat(long age) {}

    public int getSize() {
        return size;
    }

    public void doSomething(int a, int b) {}
    @Override
        if (this == obj) return true;
        if (obj == null || getClass() != obj.getClass()) return false;
        Cat cat = (Cat) obj;
        return size == cat.size;
    }

    @Override
    public int hashCode() {
        return Objects.hash(size);
    }

    @Override
    public String toString() {
        return "Cat{size=" + size + "}";
    }
}

public record Car(int side) {}
```

image

@mickaelistria Is there a reason, why it always checks if there are more than one parameter? Because if not, I think it would be beneficial if the code minings are always shown. 

## After this change
Code minings are always shown no matter how many Parameters there are.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/javaeditor/codemining/CalleeJavaMethodParameterVisitor.java
+++ b/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/javaeditor/codemining/CalleeJavaMethodParameterVisitor.java
@@ -165,11 +165,15 @@ public class CalleeJavaMethodParameterVisitor extends HierarchicalASTVisitor {
 	}
 
 	private boolean skipParameterNamesCodeMinings(IMethod method) {
-		return method.getNumberOfParameters() <= 1;
+		boolean showNamesOfSingleArguments= PreferenceConstants.getPreferenceStore().getBoolean(PreferenceConstants.EDITOR_JAVA_CODEMINING_SHOW_PARAMETER_NAME_SINGLE_ARG);
+		return !showNamesOfSingleArguments && method.getNumberOfParameters() == 1
+				|| method.getNumberOfParameters() == 0;
 	}
 
 	private boolean skipParameterNamesCodeMinings(IMethodBinding methodBinding) {
-		return methodBinding.getParameterNames().length <= 1;
+		boolean showNamesOfSingleArguments= PreferenceConstants.getPreferenceStore().getBoolean(PreferenceConstants.EDITOR_JAVA_CODEMINING_SHOW_PARAMETER_NAME_SINGLE_ARG);
+		return !showNamesOfSingleArguments && methodBinding.getParameterNames().length == 1
+				|| methodBinding.getParameterNames().length == 0;
 	}
 
 	private boolean skipParameterNamesCodeMinings(IMethod method, String[] parameterNames) {
--- a/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/preferences/PreferencesMessages.properties
+++ b/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/preferences/PreferencesMessages.properties
@@ -1038,6 +1038,7 @@ JavaEditorCodeMiningConfigurationBlock_showImplementations_label=Show &implement
 JavaEditorCodeMiningConfigurationBlock_showParameterNames_label=Show method &parameter names
 JavaEditorCodeMiningConfigurationBlock_defaultFilterForParameterNames_label=&Default filter for some specified methods and method parameter names (e.g. compare())
 JavaEditorCodeMiningConfigurationBlock_filterImpliedParameterNames_label=Filter parameter &names that are implied by parameter
+JavaEditorCodeMiningShowOneParameter_label=Show code minings when method has one parameter
 
 #Launching preferences
 JavaLaunchingConfigurationBlock_application_name_fully_qualified_label=Use fully qualified name for new Java Application
--- a/org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/codemining/ParameterNamesCodeMiningTest.java
+++ b/org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/codemining/ParameterNamesCodeMiningTest.java
@@ -19,6 +19,7 @@ import static org.junit.Assert.assertTrue;
 
 import java.util.Arrays;
 import java.util.Map;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.LongSupplier;
@@ -46,6 +47,7 @@ import org.eclipse.ui.IEditorReference;
 import org.eclipse.ui.IWorkbench;
 import org.eclipse.ui.IWorkbenchPage;
 import org.eclipse.ui.IWorkbenchWindow;
+import org.eclipse.ui.PartInitException;
 import org.eclipse.ui.PlatformUI;
 import org.eclipse.ui.intro.IIntroManager;
 import org.eclipse.ui.intro.IIntroPart;
@@ -54,6 +56,7 @@ import org.eclipse.jdt.core.ICompilationUnit;
 import org.eclipse.jdt.core.IJavaProject;
 import org.eclipse.jdt.core.IPackageFragment;
 import org.eclipse.jdt.core.IPackageFragmentRoot;
+import org.eclipse.jdt.core.JavaModelException;
 
 import org.eclipse.jdt.ui.PreferenceConstants;
 
@@ -381,7 +384,17 @@ public class ParameterNamesCodeMiningTest {
 	}
 
 	@Test
-	public void testRecordConstructorOneParameter() throws Exception {
+	public void testRecordConstructorOneParameterPreferenceFalse() throws Exception {
+		assertParameterNamesShown(false, 0);
+	}
+
+	@Test
+	public void testRecordConstructorOneParameterPreferenceTrue() throws Exception {
+		assertParameterNamesShown(true, 1);
+	}
+
+	private void assertParameterNamesShown(boolean preferenceValue, int expectedShownNames) throws JavaModelException, PartInitException, InterruptedException, ExecutionException {
+		PreferenceConstants.getPreferenceStore().setValue(PreferenceConstants.EDITOR_JAVA_CODEMINING_SHOW_PARAMETER_NAME_SINGLE_ARG, preferenceValue);
 		String contents= """
 				public class Ant {
 			record Ca (int size){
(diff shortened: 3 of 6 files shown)


fedejeanne:
> > Is there a reason, why it always checks if there are more than one parameter?
> 
> [...] when there is a single arg, the method name is most often sufficient to sort out what's going to happen with the given arg.

Not always, think of constructors

image

I found that having the names for methods with 1 argument is useful and also more consistent with the rest of the behavior. Thank you @jannisCode for this suggestion!

org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/javaeditor/codemining/CalleeJavaMethodParameterVisitor.java
```diff
@@ -165,7 +165,7 @@ private boolean skipParameterNameCodeMining(String[] parameterNames, List<?> arg
 	}
 
 	private boolean skipParameterNamesCodeMinings(IMethod method) {
```
fedejeanne, line 167: Remove this method altogether.

mickaelistria:
> Not always, think of constructors

The relevant case to check is when you read code. Even if there are several possible constructors, it's not likely to me that having the parameter name will help readability, eg `new Shell(SWT.ON_TOP)` or `new Shell(Display.getCurrent())` or `new Shell(mainShell)`.... read well without showing the parameter: you rarely wonder what the parameter actually is and what it's used for (unless you poorly name you variables).

> more consistent with the rest of the behavior. 

Consistency is less important that being useful and lean ;)

I would strongly hate having the parameter names in 1 method args, it would just add noise to me in the very vast majority of use cases, and useless noise hurts readability more than it helps.
So please add a preference to allow users like me to skip such parameter names annotations for those single-args methods.

mickaelistria (changes requested):
Please also provide an option in preferences to skip those.

fedejeanne:
> > more consistent with the rest of the behavior.
> 
> Consistency is less important that being useful and lean ;)

Consistency is important so that users don't end up reporting "bugs" that are not bugs and simply some "other arbitrary behavior". What I meant by that is not that the 1-arg methods don't show the name of the argument but that this behavior is **not** the same across the board: records don't exhibit this behavior.

> Please also provide an option in preferences to skip those.

I'd rather not have such an option, I think the configuration is complex enough. But there are maybe some other opinions on that. Does anyone else (other than Mickael, Jannis and me) reading this PR care to break the tie?

mickaelistria:
> records don't exhibit this behavior.

That is more a bug to me, to fix separately.

> I'd rather not have such an option, I think the configuration is complex enough.

Please make sure you actually try to work with this proposal on various classes before considering it for a merge. My arbitrary choice was based on actual usage of the feature while it was being implemented. My impression based on that experience is that in the vast majority of cases, single arg name is noise and hurts readability.
If several people can actually try it for real, on real code, for some a few days, and report that this the hint are more often useful to them, I will stop arguing; but at the moment, I seem to be the only one who actually experimented that (some years ago) and my feedback is that this is going to hurt readability and would perceived, at least by 1 user, as a regression in usability.
So an option is a must-have to me.

org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/javaeditor/codemining/CalleeJavaMethodParameterVisitor.java
```diff
@@ -165,7 +165,7 @@ private boolean skipParameterNameCodeMining(String[] parameterNames, List<?> arg
...
-		return method.getNumberOfParameters() <= 1;
+		return false;
```
mickaelistria, line 168: Please use a preference here
```suggestion
		return PreferenceConstants.getPreferenceStore().getBoolean(PreferencesConstants.EDITOR_JAVA_CODEMINING_SHOW_PARAMETER_NAME_SINGLE_ARG) && method.getNumberOfParameters() == 1) || method.getNumberOfParameters() == 0;
```

fedejeanne: Rather the opposite:

```java
boolean showNamesOfSingleArguments= PreferenceConstants.getPreferenceStore().getBoolean(PreferenceConstants.EDITOR_JAVA_CODEMINING_SHOW_PARAMETER_NAME_SINGLE_ARG);
return !showNamesOfSingleArguments && method.getNumberOfParameters() == 1
		|| method.getNumberOfParameters() == 0;
```

mickaelistria:
Thanks. That proposal is good to me!

org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/javaeditor/codemining/CalleeJavaMethodParameterVisitor.java
mickaelistria: if you're into renaming with opposite meanings, I suggest renaming (and reversing conditions) from `skipParameterNamesCodeMinings` to `showParameterNamesCodeMinings() { return number > 1 || PreferenceConstants.getPreferenceStore().getBoolean(PreferencesConstants.EDITOR_JAVA_CODEMINING_SHOW_PARAMETER_NAME_SINGLE_ARG))`. That should make things simpler to understand and maintain.

org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/javaeditor/codemining/CalleeJavaMethodParameterVisitor.java
```diff
@@ -165,7 +165,13 @@ private boolean skipParameterNameCodeMining(String[] parameterNames, List<?> arg
 	}
 
 	private boolean skipParameterNamesCodeMinings(IMethod method) {
+		IPreferenceStore store= JavaPlugin.getDefault().getPreferenceStore();
+		boolean showFromOneParameter = store.getBoolean(PreferenceConstants.EDITOR_JAVA_SHOW_ONE_PARAMETER);
```
szarnekow, line 169: Do you really want to read from the store for each method?

szarnekow: And in line 179 another preference store access?

org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/javaeditor/codemining/CalleeJavaMethodParameterVisitor.java
fedejeanne: No need to rename the method in this PR, it is fine as it is and it is overloaded so the renaming would have to be done for these 4 methods just so the class stays consistent:

image

org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/javaeditor/codemining/CalleeJavaMethodParameterVisitor.java
fedejeanne: Would it be ok to leave that to another PR so that this one doesn't blow up?

szarnekow:
> Would it be ok to leave that to another PR so that this one doesn't blow up?

Yes, sure.

org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/javaeditor/codemining/CalleeJavaMethodParameterVisitor.java
szarnekow: Yes, sure.

fedejeanne (changes requested):
I tested it and it works as expected. There is only one small typo in the label of the preferences.

After fixing that, a committer can merge this one.

org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/preferences/PreferencesMessages.properties
```diff
@@ -1038,6 +1038,7 @@ JavaEditorCodeMiningConfigurationBlock_showImplementations_label=Show &implement
 JavaEditorCodeMiningConfigurationBlock_showParameterNames_label=Show method &parameter names
 JavaEditorCodeMiningConfigurationBlock_defaultFilterForParameterNames_label=&Default filter for some specified methods and method parameter names (e.g. compare())
 JavaEditorCodeMiningConfigurationBlock_filterImpliedParameterNames_label=Filter parameter &names that are implied by parameter
+JavaEditorCodeMiningShowOneParameter_label=Show code minnigs when method has one parameter
```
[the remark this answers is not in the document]
jannisCode: thanks for noticing! I changed it

fedejeanne (approved):
The code looks sound and the minings work as expected.

Would a JDT committer be so kind to (review/test and) merge this one? Thank you in advance.

@jukzi / @jjohnstn / @iloveeclipse

jukzi:
you  need to get a green CI first. failing `JdtTextTestSuite ParameterNamesCodeMiningTest.testRecordConstructorOneParameter` sounds related

fedejeanne:
> you need to get a green CI first. failing `JdtTextTestSuite ParameterNamesCodeMiningTest.testRecordConstructorOneParameter` sounds related

@jukzi sorry, I missed that one.

@jannisCode, like we spoke last week, you need to replace the `testRecordConstructorOneParameter` with these 2 tests so you also test the new parameter:

```java
@Test
public void testRecordConstructorOneParameterPreferenceFalse() throws Exception {
	assertParameterNamesShown(false, 0);
}

@Test
public void testRecordConstructorOneParameterPreferenceTrue() throws Exception {
	assertParameterNamesShown(true, 1);
}

private void assertParameterNamesShown(boolean preferenceValue, int expectedShownNames) throws JavaModelException, PartInitException, InterruptedException, ExecutionException {
	PreferenceConstants.getPreferenceStore().setValue(PreferenceConstants.EDITOR_JAVA_CODEMINING_SHOW_PARAMETER_NAME_SINGLE_ARG, preferenceValue);
	String contents= """
			public class Ant {
		record Ca (int size){

		}

		Ca c = new Ca(0);

	}
	""";
	ICompilationUnit compilationUnit= fPackage.createCompilationUnit("Ant.java", contents, true, new NullProgressMonitor());
	JavaEditor editor= (JavaEditor) EditorUtility.openInEditor(compilationUnit);
	fParameterNameCodeMiningProvider.setContext(editor);
	JavaSourceViewer viewer= (JavaSourceViewer)editor.getViewer();
	waitReconciled(viewer);

	assertEquals(expectedShownNames, fParameterNameCodeMiningProvider.provideCodeMinings(viewer, new NullProgressMonitor()).get().size());
}
```
I don't have committer rights so you need to do it once you're back in the office.

fedejeanne (changes requested):
Please adapt the tests

fedejeanne (commented):
LGTM now. 

@jukzi ?

jukzi:
thanks for contributing

fedejeanne:
> thanks for contributing

Thank you for merging!

PR #1822, opened 2024-11-29
Record classes are now handled the same as normal classes

jannisCode:
## What it does
Record classes are now handled the same way as normal classes. 
This bug fix was proposed here: https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/1813

## Before
Record classes were not checked for the amount of parameters which their methods had. Normal classes where though. That posed the problem, that when only one argument was given, the code minings were shown for record classes but not for normal classes.

## How to test
Use this code snippet:
```` java
import java.util.Objects;

class Main {
	public static void main(String...args) {
		Car c = new Car(3);
		Cat car = new Cat(5);
		car.equals(c);
		car.doSomething(12, 13);
	}
}

 class Cat {
    private int size;

    public Cat(int size) {
        this.size = size;
    }
    public Cat(long age) {}

    public int getSize() {
        return size;
    }

    public void doSomething(int a, int b) {}
    @Override
        if (this == obj) return true;
        if (obj == null || getClass() != obj.getClass()) return false;
        Cat cat = (Cat) obj;
        return size == cat.size;
    }

    @Override
    public int hashCode() {
        return Objects.hash(size);
    }

    @Override
    public String toString() {
        return "Cat{size=" + size + "}";
    }
}

public record Car(int side) {}

````

it will result in this:
Screenshot 2024-11-29 113901

## After the changes
Record classes are checked for the number of parameters the same way as normal classes and methods are checked. For the discussion on when they should be checked please take a look at https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/1813#issuecomment-2507666908  and feel free to tell us your opinion. 

## Author checklist

### DIFF
--- a/org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/codemining/ParameterNamesCodeMiningTest.java
+++ b/org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/codemining/ParameterNamesCodeMiningTest.java
@@ -381,6 +381,27 @@ public class ParameterNamesCodeMiningTest {
 	}
 
 	@Test
+	public void testRecordConstructorOneParameter() throws Exception {
+		String contents= """
+				public class Ant {
+			record Ca (int size){
+
+			}
+
+			Ca c = new Ca(0);
+
+		}
+		""";
+		ICompilationUnit compilationUnit= fPackage.createCompilationUnit("Ant.java", contents, true, new NullProgressMonitor());
+		JavaEditor editor= (JavaEditor) EditorUtility.openInEditor(compilationUnit);
+		fParameterNameCodeMiningProvider.setContext(editor);
+		JavaSourceViewer viewer= (JavaSourceViewer)editor.getViewer();
+		waitReconciled(viewer);
+
+		assertEquals(0, fParameterNameCodeMiningProvider.provideCodeMinings(viewer, new NullProgressMonitor()).get().size());
+	}
+
+	@Test
 	public void testRecordConstructorOK() throws Exception {
 		String contents= "import java.util.Map;\n"
 				+ "public record Edge(int fromNodeId,\n"
--- a/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/javaeditor/codemining/CalleeJavaMethodParameterVisitor.java
+++ b/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/javaeditor/codemining/CalleeJavaMethodParameterVisitor.java
@@ -129,7 +129,7 @@ public class CalleeJavaMethodParameterVisitor extends HierarchicalASTVisitor {
 
 	protected void collectParameterNamesCodeMinings(IMethodBinding mbinding, List<?> arguments, boolean isVarArgs) {
 		// synthetic method of a record
-		if (mbinding.getDeclaringClass().isRecord()) {
+		if (mbinding.getDeclaringClass().isRecord() && !skipParameterNamesCodeMinings(mbinding)) {
 			String[] parameterNames= mbinding.getParameterNames();
 			for (int i= 0; i < Math.min(arguments.size(), parameterNames.length); i++) {
 				if (!skipParameterNameCodeMining(parameterNames, arguments, i)) {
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by fedejeanne** [comment]
@jukzi would you kindly (review and) merge this? 

I don't have commit rights :-)

**Comment by jukzi** [comment]
The code change looks good, but it would be better if you could add a junit test to avoid regression. Can you work on that?

File: org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/codemining/ParameterNamesCodeMiningTest.java [hunk 1]
```diff
@@ -124,6 +124,47 @@ public class Foo {
 		assertEquals(3, fParameterNameCodeMiningProvider.provideCodeMinings(viewer, new NullProgressMonitor()).get().size());
 	}
 
+	@Test
+	public void testRecordClasses() throws Exception {
```
fedejeanne, line 128: Rename it and move it closer to the already existing test called `testRecordConstructorOK` (they belong together):

```suggestion
	public void testRecordConstructorOneParameter() throws Exception {
```

org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/codemining/ParameterNamesCodeMiningTest.java
```diff
@@ -124,6 +124,47 @@ public class Foo {
...
+				}
+			""";
+		ICompilationUnit compilationUnit= fPackage.createCompilationUnit("Ant.java", contents, true, new NullProgressMonitor());
+		JavaEditor editor= (JavaEditor) EditorUtility.openInEditor(compilationUnit);
+		fParameterNameCodeMiningProvider.setContext(editor);
+		JavaSourceViewer viewer= (JavaSourceViewer)editor.getViewer();
+		waitReconciled(viewer);
+		assertEquals(2, fParameterNameCodeMiningProvider.provideCodeMinings(viewer,  new NullProgressMonitor()).get().size());
+	}
+
```
fedejeanne, line 167: Remove this test, it does the same as `testRecordConstructorOK`: it tests records with >1 arg:

```suggestion
```

org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/codemining/ParameterNamesCodeMiningTest.java
jannisCode: done

org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/codemining/ParameterNamesCodeMiningTest.java
jannisCode: done

org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/codemining/ParameterNamesCodeMiningTest.java
```diff
@@ -124,6 +124,7 @@ public class Foo {
 		assertEquals(3, fParameterNameCodeMiningProvider.provideCodeMinings(viewer, new NullProgressMonitor()).get().size());
 	}
 
+
```
jukzi, line 127: redundant new line

jannisCode: fixed it!

jukzi (approved):
LGTM, manually tested, that the test fails without the change.
thanks for contributing.

jukzi:
unrelated build error, i don't know how to handle that
`00:05:04.844  Caused by: org.apache.maven.plugin.MojoFailureException: Version has moved backwards for (org.eclipse.jdt.core.manipulation/1.21.300.v20241101-2227). Baseline has 1.21.300.v20241108-2119) with delta: 0.0.0
`
lets try rebase.

fedejeanne:
> There will be unrelated know 12 fails #1833 that can be ignored today

so.... merge? :-)

PR #1825, opened 2024-12-02
custom folding regions

danthe1st:
Relevant issues: https://bugs.eclipse.org/bugs/show_bug.cgi?id=173796 (and there are also similar ones like https://bugs.eclipse.org/bugs/show_bug.cgi?id=258965 for CDT).

There is also a [Stack Overflow post](https://stackoverflow.com/q/13505413/10871900) asking for how to do that with quite a few upvotes so I'd say it seems like many people are interested in such a feature.

## What it does
This PR allows to add custom folding regions using comments.

I added new options to the preferences in `Java` > `Editor` > `Folding` (not that I changed this preference page to use `Group`s):  
folding preference dialog including custom folding regions

If a comment containing the start region and a comment containing the end region are present in the same source file, a folding region is added allowing to collapse that section. 

non-collapsed custom region
collapsed custom region
hovering over collapsed region indicator showing content

It would be possible to change the defaults from `region`/`endregion` to `#region`/`#endregion` or similar. I have decided for `// region`, `// endregion` as that seems less arbitrary. I don't really have an opinion about the best choice of defaults here.

This functionality is added both to the normal folding and the new folding based on the AST introduced in #1562. The folding using the AST uses existing folding annotations to compute the new ones.

~~This works even for elements of different levels~~ This is intentionally restricted since someone folding things from different elements is likely a mistake of the user (they probably don't want to fold something in the bodies of different methods or similar) as discussed in a devcall.

~~This PR also allows configuring folding per-project as requested in [this comment](https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/1825#issuecomment-2565439652).~~ This has been moved to #1979

## How to test
- Create a Java source file
- Create a comment starting with the text `region`
- Create a comment starting with the text `endregion` after the `region` comment (but within the same block (e.g. method/class))
- Test similar things with the source file
- After editing the corresponding changes in `Java` > `Editor` > `Folding`, ~~it is necessary to reopen the source file.~~ the changes are automatically applied to open editors (and will also be used when opening a new editor).
  - It is also possible to change these preferences in the project properties. This also requires reopening the source file.

## Author checklist

  - performance testing: https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/1825#issuecomment-2562957272

### DIFF
--- a/org.eclipse.jdt.ui/ui/org/eclipse/jdt/ui/text/folding/DefaultJavaFoldingStructureProvider.java
+++ b/org.eclipse.jdt.ui/ui/org/eclipse/jdt/ui/text/folding/DefaultJavaFoldingStructureProvider.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2006, 2024 IBM Corporation and others.
+ * Copyright (c) 2006, 2025 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -13,11 +13,13 @@
  *******************************************************************************/
 package org.eclipse.jdt.ui.text.folding;
 
+import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
+import java.util.Deque;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
@@ -77,6 +79,7 @@ import org.eclipse.jdt.core.dom.ASTVisitor;
 import org.eclipse.jdt.core.dom.AnonymousClassDeclaration;
 import org.eclipse.jdt.core.dom.Block;
 import org.eclipse.jdt.core.dom.CatchClause;
+import org.eclipse.jdt.core.dom.Comment;
 import org.eclipse.jdt.core.dom.CompilationUnit;
 import org.eclipse.jdt.core.dom.DoStatement;
 import org.eclipse.jdt.core.dom.EnhancedForStatement;
@@ -129,6 +132,10 @@ public class DefaultJavaFoldingStructureProvider implements IJavaFoldingStructur
 		private LinkedHashMap<JavaProjectionAnnotation, Position> fMap= new LinkedHashMap<>();
 		private IScanner fDefaultScanner; // this one may or not be the shared DefaultJavaFoldingStructureProvider.fSharedScanner
 		private IScanner fScannerForProject;
+
+		private Deque<Integer> fOpenCustomRegionStartPositions = new ArrayDeque<>();
+		private Set<IRegion> fCurrentCustomRegions = new HashSet<>();
+		private int fLastScannedIndex;
 
 		private FoldingStructureComputationContext(IDocument document, ProjectionAnnotationModel model, boolean allowCollapsing, IScanner scanner) {
 			Assert.isNotNull(document);
@@ -268,6 +275,17 @@ public class DefaultJavaFoldingStructureProvider implements IJavaFoldingStructur
 		public boolean collapseMembers() {
 			return fAllowCollapsing && fCollapseMembers;
 		}
+
+		/**
+		 * Returns <code>true</code> if custom regions should be collapsed.
+		 *
+		 * @return <code>true</code> if custom regions should be collapsed
+		 * @since 3.35
+		 */
+		public boolean collapseCustomRegions() {
+			return fAllowCollapsing && fCollapseCustomRegions;
+		}
+
 	}
 
 	/**
@@ -997,9 +1015,7 @@ public class DefaultJavaFoldingStructureProvider implements IJavaFoldingStructur
 	private IPropertyChangeListener fPropertyChangeListener = new IPropertyChangeListener() {
 	    @Override
 	    public void propertyChange(PropertyChangeEvent event) {
-	        if (PreferenceConstants.EDITOR_NEW_FOLDING_ENABLED.equals(event.getProperty())) {
-	            fNewFolding = event.getNewValue().equals(Boolean.TRUE);
-	        }
+	        initializePreferences();
 	        initialize();
 	    }
 	};
@@ -1010,7 +1026,12 @@ public class DefaultJavaFoldingStructureProvider implements IJavaFoldingStructur
 	private boolean fCollapseInnerTypes= true;
 	private boolean fCollapseMembers= false;
 	private boolean fCollapseHeaderComments= true;
+	private boolean fCollapseCustomRegions= false;
 	private boolean fNewFolding;
+
+	private boolean fCustomFoldingRegionsEnabled;
+	private char[] fCustomFoldingRegionBegin;
+	private char[] fCustomFoldingRegionEnd;
 
 	/* filters */
 	/** Member filter, matches nested members (but not top-level types). */
@@ -1188,7 +1209,15 @@ public class DefaultJavaFoldingStructureProvider implements IJavaFoldingStructur
 	    fCollapseJavadoc = store.getBoolean(PreferenceConstants.EDITOR_FOLDING_JAVADOC);
 	    fCollapseMembers = store.getBoolean(PreferenceConstants.EDITOR_FOLDING_METHODS);
 	    fCollapseHeaderComments = store.getBoolean(PreferenceConstants.EDITOR_FOLDING_HEADERS);
-	    fNewFolding = store.getBoolean(PreferenceConstants.EDITOR_NEW_FOLDING_ENABLED);
+
+		String customFoldingRegionBegin= store.getString(PreferenceConstants.EDITOR_FOLDING_CUSTOM_REGION_START);
+		String customFoldingRegionEnd= store.getString(PreferenceConstants.EDITOR_FOLDING_CUSTOM_REGION_END);
+		fCustomFoldingRegionBegin=customFoldingRegionBegin.toCharArray();
+		fCustomFoldingRegionEnd=customFoldingRegionEnd.toCharArray();
+		fCustomFoldingRegionsEnabled = store.getBoolean(PreferenceConstants.EDITOR_FOLDING_CUSTOM_REGIONS_ENABLED) &&
+				!customFoldingRegionBegin.isEmpty() && !customFoldingRegionEnd.isEmpty() &&
+				!customFoldingRegionBegin.startsWith(customFoldingRegionEnd) && !customFoldingRegionEnd.startsWith(customFoldingRegionBegin);
+		fNewFolding = store.getBoolean(PreferenceConstants.EDITOR_NEW_FOLDING_ENABLED);
 	}
 
 	private void update(FoldingStructureComputationContext ctx) {
@@ -1286,7 +1315,8 @@ public class DefaultJavaFoldingStructureProvider implements IJavaFoldingStructur
 	        String source = unit.getSource();
 	        if (source == null) return;
 
-	        ctx.getScanner().setSource(source.toCharArray());
+	        char[] sourceArray= source.toCharArray();
+			ctx.getScanner().setSource(sourceArray);
 	        ASTParser parser = ASTParser.newParser(AST.getJLSLatest());
 	        parser.setBindingsRecovery(true);
 	        parser.setStatementsRecovery(true);
@@ -1296,16 +1326,145 @@ public class DefaultJavaFoldingStructureProvider implements IJavaFoldingStructur
 	        parser.setProject(unit.getJavaProject());
 	        parser.setSource(unit);
 	        Map<String, String> options = unit.getJavaProject().getOptions(true);
-		    options.put(JavaCore.COMPILER_SOURCE, JavaCore.VERSION_23);
-		    options.put(JavaCore.COMPILER_COMPLIANCE, JavaCore.VERSION_23);
-		    options.put(JavaCore.COMPILER_CODEGEN_TARGET_PLATFORM, JavaCore.VERSION_23);
+		    options.put(JavaCore.COMPILER_SOURCE, JavaCore.latestSupportedJavaVersion());
+		    options.put(JavaCore.COMPILER_COMPLIANCE, JavaCore.latestSupportedJavaVersion());
+		    options.put(JavaCore.COMPILER_CODEGEN_TARGET_PLATFORM, JavaCore.latestSupportedJavaVersion());
 		    options.put(JavaCore.COMPILER_DOC_COMMENT_SUPPORT, JavaCore.ENABLED);
 	        parser.setCompilerOptions(options);
 
 	        CompilationUnit ast = (CompilationUnit) parser.createAST(null);
-	        ast.accept(new FoldingVisitor(ctx));
-	    } catch (JavaModelException | IllegalStateException e) {
-	    }
+	        FoldingVisitor visitor= new FoldingVisitor(ctx);
+			ast.accept(visitor);
+
+			if (fCustomFoldingRegionsEnabled) {
+				List<Comment> comments= ast.getCommentList();
+				int currentCommentIndex= 0;
+
+				// Go through all already discovered folding regions in order to compute custom regions
+				// Disjoint folding regions are processed sequentially.
+				// Nested folding regions are processed depth-first.
+
+				// stack containing the "current" nested folding regions
+				// the top of the stack is the innermost region
+				Deque<Position> currentFoldingPositions= new ArrayDeque<>();
+				// stack containing the position of "open" regions for each element in currentFoldingPositions
+				// i.e. regions where the start position has been found but no end position
+				// Whenever an element is added/removed from currentFoldingPositions, the same is done with openCustomRegionStartPositions
+				Deque<Deque<Integer>> openCustomRegionStartPositions= new ArrayDeque<>();
+				openCustomRegionStartPositions.add(new ArrayDeque<>());
+
+				for (Position nonCommentFoldingRegion : List.copyOf(ctx.fMap.values())) {
+
+					// process regions depth-first until reaching the current region
+					currentCommentIndex= processFoldingRegionsForCustomCommentFolding(
+							sourceArray, visitor, comments, currentCommentIndex,
+							currentFoldingPositions, openCustomRegionStartPositions, nonCommentFoldingRegion.getOffset()
+					);
+					if (currentCommentIndex >= comments.size()) {
+						return;
+					}
+
+					// process comments before current region starts
+					currentCommentIndex= checkCustomFoldingCommitsBeforePosition(sourceArray, visitor, comments, currentCommentIndex, openCustomRegionStartPositions.peekLast(), nonCommentFoldingRegion.getOffset());
+
+					currentFoldingPositions.addLast(nonCommentFoldingRegion);
+					openCustomRegionStartPositions.addLast(new ArrayDeque<>());
+				}
+				// process all leftover comments at the end
+				currentCommentIndex= processFoldingRegionsForCustomCommentFolding(sourceArray, visitor, comments, currentCommentIndex, currentFoldingPositions, openCustomRegionStartPositions, Integer.MAX_VALUE);
+				checkCustomFoldingCommitsBeforePosition(sourceArray, visitor, comments, currentCommentIndex, openCustomRegionStartPositions.peek(), Integer.MAX_VALUE);
+			}
+		} catch (JavaModelException | IllegalStateException e) {
+		}
+	}
+
+	/**
+	 * Pops regions before limit and checks for custom folding comments.
+	 *
+	 * @param sourceArray the content of the processed source file
+	 * @param visitor the {@link FoldingVisitor} used for adding folding regions
+	 * @param comments all comments in the source file
+	 * @param currentCommentIndex the index of the next comment to process
+	 * @param currentFoldingPositions stack of current nested folding regions
+	 * @param openCustomRegionStartPositions start positions of custom folding regions with the
+	 *            start position being found but the end positions still missing
+	 * @param limit only regions before this position are scanned
+	 * @return the new index of the next comment
+	 */
+	private int processFoldingRegionsForCustomCommentFolding(char[] sourceArray, FoldingVisitor visitor, List<Comment> comments, int currentCommentIndex,
+			Deque<Position> currentFoldingPositions, Deque<Deque<Integer>> openCustomRegionStartPositions, int limit) {
+		Position currentFoldingPosition= currentFoldingPositions.peekLast();
+		while (currentFoldingPosition != null && currentFoldingPosition.getOffset() + currentFoldingPosition.getLength() < limit) {
+			currentFoldingPositions.removeLast();
+			Deque<Integer> innerOpenStartPositions= openCustomRegionStartPositions.removeLast();
+
+			currentCommentIndex= checkCustomFoldingCommitsBeforePosition(sourceArray, visitor, comments, currentCommentIndex, innerOpenStartPositions, currentFoldingPosition.getOffset() + currentFoldingPosition.getLength());
+
+			if (currentCommentIndex >= comments.size()) {
+				return currentCommentIndex;
+			}
+
+			currentFoldingPosition= currentFoldingPositions.peekLast();
+		}
+		return currentCommentIndex;
+	}
+
+
+	/**
+	 * Searches for custom folding comments before a specified position and adds corresponding folding annotations.
+	 *
+	 * @param sourceArray the content of the processed source file
+	 * @param visitor the {@link FoldingVisitor} used for adding folding annotations
+	 * @param comments all comments in the source file
+	 * @param currentCommentIndex the index of the first comment to process
+	 * @param innerOpenStartPositions start positions of custom folding regions with the start position being found but the end positions still missing
+	 * @param limit only regions before this position are scanned
+	 * @return the new index of the next comment to process (after limit)
+	 */
+	private int checkCustomFoldingCommitsBeforePosition(char[] sourceArray, FoldingVisitor visitor, List<Comment> comments, int currentCommentIndex,
+			Deque<Integer> innerOpenStartPositions, int limit) {
+		while (currentCommentIndex < comments.size() && comments.get(currentCommentIndex).getStartPosition() < limit) {
+			Comment comment= comments.get(currentCommentIndex);
+			checkCustomFolding(innerOpenStartPositions, sourceArray, visitor, comment);
+			currentCommentIndex++;
+		}
+		return currentCommentIndex;
+	}
+
+	private void checkCustomFolding(Deque<Integer> openCustomRegionStartPositions, char[] sourceArray, FoldingVisitor visitor, Comment comment) {
+		int skip= 2;
+		if (comment.isDocComment()) {
+			skip= 3;
+		}
+		int commentTextStart= skipLeadingWhitespace(sourceArray, comment.getStartPosition() + skip);
+		IRegion customFoldingRegion= checkCustomFolding(
+				openCustomRegionStartPositions, commentTextStart, sourceArray,
+				comment.getStartPosition(), comment.getStartPosition() + comment.getLength()
+		);
+		if (customFoldingRegion != null) {
+			if (includeLastLineInCustomFoldingRegion(sourceArray, customFoldingRegion.getOffset() + customFoldingRegion.getLength())) {
+				includelastLine= true;
+			}
+			visitor.createFoldingRegion(customFoldingRegion.getOffset(), customFoldingRegion.getLength(), fCollapseCustomRegions);
+		}
+	}
+
+	private boolean includeLastLineInCustomFoldingRegion(char[] sourceArray, int regionEnd) {
+		char firstCharacter = sourceArray[regionEnd];
+		if (firstCharacter == '\n' || firstCharacter == '\r') {
+			return true;
+		}
+		for (int i= regionEnd + 1; i < sourceArray.length; i++) {
+			char c= sourceArray[i];
+			if (c == '\n' || c == '\r') {
+				return true;
+			}
+			// allow custom regions comments defined in {/* ... */} without that causing the end to be shown
+			if (!Character.isWhitespace(c) && c != '}') {
+				return false;
+			}
+		}
+		return true;
 	}
 
 	private void processComments(FoldingStructureComputationContext ctx) {
@@ -1442,7 +1601,15 @@ public class DefaultJavaFoldingStructureProvider implements IJavaFoldingStructur
 
 			if (element instanceof IParent) {
 				IParent parent= (IParent) element;
+				Deque<Integer> outerOpenRegions = ctx.fOpenCustomRegionStartPositions;
+				ctx.fOpenCustomRegionStartPositions = new ArrayDeque<>();
 				computeFoldingStructure(parent.getChildren(), ctx);
+				ctx.fOpenCustomRegionStartPositions = outerOpenRegions;
+			}
+			if (fCustomFoldingRegionsEnabled && element instanceof ISourceReference sourceRef) {
+				// mark as scanned until end after scanning all children
+				ISourceRange sourceRange= sourceRef.getSourceRange();
+				ctx.fLastScannedIndex = sourceRange.getLength()+sourceRange.getOffset();
 			}
 		}
 	}
@@ -1493,13 +1660,16 @@ public class DefaultJavaFoldingStructureProvider implements IJavaFoldingStructur
 			// comments
 			for (int i= 0; i < regions.length - 1; i++) {
 				includelastLine= true;
-				IRegion normalized= alignRegion(regions[i], ctx);
+				IRegion region= regions[i];
+				IRegion normalized= alignRegion(region, ctx);
 				if (normalized != null) {
 					Position position= createCommentPosition(normalized);
 					if (position != null) {
 						boolean commentCollapse;
 						if (i == 0 && (regions.length > 2 || ctx.hasHeaderComment()) && element == ctx.getFirstType()) {
 							commentCollapse= ctx.collapseHeaderComments();
+						} else if(ctx.fCurrentCustomRegions.contains(region)) {
+							commentCollapse= ctx.collapseCustomRegions();
 						} else {
 							commentCollapse= ctx.collapseJavadoc();
 						}
@@ -1588,24 +1758,40 @@ public class DefaultJavaFoldingStructureProvider implements IJavaFoldingStructur
 
 				final int shift= range.getOffset();
 				IScanner scanner= ctx.getScanner();
+
+				if (fCustomFoldingRegionsEnabled &&
+						reference instanceof IJavaElement javaElement && javaElement.getParent() != null &&
+							javaElement.getParent() instanceof IParent parent && parent instanceof ISourceReference parentSourceReference) {
+						// check tokens between the last sibling (or the parent) and start of current sibling
+						ISourceRange parentSourceRange= parentSourceReference.getSourceRange();
+						if (ctx.fLastScannedIndex >= parentSourceRange.getOffset() && ctx.fLastScannedIndex < parentSourceRange.getOffset() + parentSourceRange.getLength()
+								&& ctx.fLastScannedIndex < range.getOffset()) {
+							scanner.resetTo(ctx.fLastScannedIndex, range.getOffset());
+							checkCustomFoldingUntilScannerEnd(ctx, regions, ctx.fOpenCustomRegionStartPositions, scanner);
+						}
+					}
+
+
 				scanner.resetTo(shift, shift + range.getLength());
 
 				int start= shift;
+
 				while (true) {
 
 					int token= scanner.getNextToken();
 					start= scanner.getCurrentTokenStartPosition();
 
 					switch (token) {
-						case ITerminalSymbols.TokenNameCOMMENT_JAVADOC:
-						case ITerminalSymbols.TokenNameCOMMENT_MARKDOWN:
-						case ITerminalSymbols.TokenNameCOMMENT_BLOCK: {
+						case ITerminalSymbols.TokenNameCOMMENT_JAVADOC, ITerminalSymbols.TokenNameCOMMENT_MARKDOWN, ITerminalSymbols.TokenNameCOMMENT_BLOCK: {
 							int end= scanner.getCurrentTokenEndPosition() + 1;
 							regions.add(new Region(start, end - start));
+							checkCustomFolding(ctx, regions, ctx.fOpenCustomRegionStartPositions, scanner, token, regions.size());
 							continue;
 						}
-						case ITerminalSymbols.TokenNameCOMMENT_LINE:
+						case ITerminalSymbols.TokenNameCOMMENT_LINE: {
+							checkCustomFolding(ctx, regions, ctx.fOpenCustomRegionStartPositions, scanner, token, regions.size());
 							continue;
+						}
 					}
 
 					break;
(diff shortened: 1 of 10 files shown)


danthe1st:
Are the CI failures due to me bumping the minor version? Should I do that differently?

HannesWell:
@fedejeanne in the dev-call on Thursday you mentioned that you could have a look at this, does this still stand?
Eventueally we need a JDT committer to submit this anyways.

danthe1st:
@jukzi Would you mind taking a look at my PR and giving feedback?

fedejeanne:
Hi. It's on my list but I am currently focusing on issues regarding the Edge browser, which have top priority until the end of the year. 
I haven't forgotten about you @danthe1st though, I am just swamped 😓

fedejeanne (commented):
I did a quick review and made some comments regarding coding style.

I also tested a bit and this is what I found:
- It doesn't work if I define a region completely inside the `public static void main(String... args)` method but it does work on other methods like `public boolean equals(Object obj)`
- I can define regions if the comment **starts with** the custom text _e.g._ `// #region2`, if that is intended then the text in the preferences should say so. I would add some extra label explaining how custom regions work, right before the 2 check-boxes
- If I fold the `// #region2` in this snippet, the `System.out.println()` is folded too:

```java
@Override
	// #region1
	// #region2
	if (this == obj)
		return true;
	if (obj == null || getClass() != obj.getClass())
		return false;
	Cat cat = (Cat) obj;
	// #endregion2
	System.out.println();
	return size == cat.size;
	// #endregion1
}
```
- Moving regions around works without reopening the source file but **creating** regions doesn't. I would very much like it to work too, it doesn't feel natural to have to close the editor to add regions.


## Your questions

> Custom folding regions are disabled if either of the custom texts/preferences are empty or if one contains the other. Should I add some information about that to the preferences? Would it be better to disable it by default (currently, it defaults to #region/#endregion)?

Some info about it is supposed to work is good but I'd keep it short, 2-3 lines in a label. If it gets too long you could consider adding some help page like the [Readme example](https://github.com/user-attachments/assets/574b8cc5-3214-42ca-972f-3724cdd17623) does.
I'd disable (gray out) the **group** "Custom folding regions" in the preferences if the **checkbox** "Custom folding regions" is not selected.

> I haven't written any tests so far. Are there existing tests for folding in JDT-UI? What about performance tests (my implementation requires scanning more parts of the source file which could affect performance)?

To test performance you'll have to do it manually. Alter this generator (credits to Andrey Loskutov) and let it generate several internal classes with custom folding regions: [GenerateJava.zip](https://github.com/user-attachments/files/18195150/GenerateJava.zip). Then sample with VisualVM and share the results.

I don't know of any functional tests, maybe someone else does?

> Regarding performance, I used String#contains to check whether comments contain certain the region begin/end tokens (e.g. #region/#endregion) which requires creating a String for each comment every time the file is edited. Should I change that to use the char[] returned by getCurrentTokenSource() or similar?

You'll have to sample with VisualVM and see if this is a performance bottle-neck.

> I have bumped the minor version of org.eclipse.jdt.ui due to adding new elements in PreferenceConstants. Is this ok that way?

It's fine and this commit did the same so you're safe: d0563dcb67679ab00626ff63283f5bceb5d0ea54. Such bumps are no longer necessary since they will be done automatically by GH, as was the case for the mentioned commit.

> I guess this should be added to the "New and noteworthy" list (assuming this change gets accepted and merged). If so, should I add it and is [this](https://github.com/eclipse-platform/www.eclipse.org-eclipse/blob/master/news/4.35/jdt.html) the right place?

Correct, but let's cross that bridge when we get there :-) . Usually a committer adds the "noteworthy" tag to your PR / Issue right after it was merged/solved and that's your queue to do the PR for the N&N entry.

danthe1st:
> It doesn't work if I define a region completely inside the public static void main(String... args) method but it does work on other methods like public boolean equals(Object obj)

Should be fixed, this was because of it being the only (or last) method.

> I can define regions if the comment starts with the custom text e.g. `// #region2`

It should work whenever a comment includes the text `#region` so it could be `// Hello #region World` or anything like that. If you would (intuitively) expect something else, I'd be happy to hear that and possibly change my implementation. Currently, it treats all comments the same way but I'm open to change that if other people prefer it differently.

> Moving regions around works without reopening the source file but creating regions doesn't. I would very much like it to work too, it doesn't feel natural to have to close the editor to add regions.

Interestingly, that worked for my by just adding a comment like `// #region` or whatever. What doesn't work is it being automatically updated when changing the preferences related to custom folding regions.

> I'd disable (gray out) the group "Custom folding regions" in the preferences if the checkbox "Custom folding regions" is not selected.

> I don't know of any functional tests, maybe someone else does?

I didn't find any existing tests related to folding so I'll try making my own tests for custom folding regions in a new directory `org.eclipse.jdt.ui.tests/ui/org/eclipse/jdt/ui/tests/folding` (I currently don't have a _good_ way to access the folding regions/annotations added to the editor and I don't want to add API just for tests but I might find a good way to do it at some point ()). If anyone still finds tests after that, it shouldn't be hard to move them (once I wrote the tests).

My current approach for automated tests is subclassing `DefaultJavaFoldingStructureProvider` and overriding `computeFoldingStructure` and `getProjectionRanges` to access the created `IRegion[]`s.

danthe1st:
image

BeckerWdf:
The texts: "Text marking start of custom folding region" looks a bit unnecessary too long for my
Why not simply saying "Text marking start of region" and "Text marking end of region"
due to the box around you don't need to repeat that it's a "custom folding region".

I am unsure if it's singular "region" or pluar "regions".

danthe1st:
As discussed in today's devcall, I changed the PR to disallow regions between an element and it's child.
```java
class X{
	// #region
	void a() {
		// #endregion this is not folded with the outer region because it is at a different level
		
		// #region However, this is folded because loops are not considered children elements and I don't think it's a good idea to parse all of that for folding
		for(int i=0; i<10;i++) {
			// #endregion
		}
	}
	// #endregion this will be folded with the outer region
}
```

danthe1st:
I ran the generator mentioned by @fedejeanne with 12_000 inner classes (the default) and added a custom region both outside and inside each of these classes.

I loaded the class in Eclipse, started CPU sampling with VisualVM, added some newlines at the top, bottom and in the middle of the file. I also added a single-line comment in the middle of the file.
This is what VisualVM shows:
image

Similar with CPU profiling:
image

As it seems, the custom folding logic does have an impact on performance.

Loading all the folding annotation takes a bit of time in general (especially with the profiler active) but it does not block the UI.


CPU profiling with similar operations _with custom folding regions disabled_:
image

fedejeanne:
@danthe1st can you provide:
1. The simplified version of the generated file (generate it with 1 inner class)
2. Sampling/performance analysis for the same scenarios on `master` (for comparison)?

Keep in mind that generating 12.000 inner classes is an extreme case so performance will always degrade, the question is _how much_.

akurtakov:
As a reference for people interested in similar functionality for other editors - Generic editor provides extension point for that https://github.com/eclipse-platform/eclipse.platform.ui/blob/master/bundles/org.eclipse.ui.genericeditor/schema/foldingReconcilers.exsd

danthe1st:
profiling with merge base (not _current_ master but the master I based my PR on):
image

sampling:
image


Generated code with one inner class:
```java
package a.generated;

public class TestClass {

    // #region outside
    public static class InnerClass0 {
        // #region inside
        public int x = 42; // no error
        
        public void helloWorld() {
                System.out.println("Hello World!"); // no error
                System.out.println("In class " + 0 + ", x=" + x); // expected error
        }
        // #endregion inside
    } // end of InnerClass0
    // #endregion

}
```

danthe1st:
Other experiment with 6_000 inner classes and each class having a multiline comment with 20 lines each containing the character `a` 100 times:
```java
// #region outside
    /*
    * aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
    * aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
    * aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
    * aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
    * aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
    * aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
    * aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
    * aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
    * aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
    * aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
    * aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
    * aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
    * aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
    * aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
    * aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
    * aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
    * aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
    * aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
    * aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
    * aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
    */
    public static class InnerClass0 {
        // #region inside
        public int x = 42; // no error
        
        public void helloWorld() {
                System.out.println("Hello World!"); // no error
                System.out.println("In class " + 0 + ", x=" + x); // expected error
        }
        // #endregion inside
    } // end of InnerClass0
    // #endregion
```

Custom regions disabled, sampling:
image

Custom regions disabled, profiling (I had to disable automated collapsing of normal/header comments to be able to do this without Eclipse not responding):
image

Custom regions enabled, sampling:
image


Custom regions enabled, profiling (automated collapsing of comments enabled as well):
image

danthe1st:
I realized that I can use the information of what the current token is from the `scanner` to know where I need to start looking for characters (so I don't need to check for things like `//`). With this change, it is now required that region markers (`#region`/`#endregion`) are at the start of the comment so `// Hello #region World` is now prevented as discussed in the devcall.

danthe1st:
I noticed a bug in my implementation where the following is not detected:
```java
class Z{
	void a() {
		// #region this is not detected because there is an inner class (considered as a child) in the same scope
		int x;
		
		// #endregion
		class I{
			
		}
	}
}
```
I plan to try fixing it.

BeckerWdf:
> image
> 
> I might still rework that label in the future.

I am not a big fan of big explanatory text directly in the preference page. Can't we simply put this into the help document related to that preference page?

BeckerWdf (commented):
Thanks for the text shortening.

danthe1st:
> I am not a big fan of big explanatory text directly in the preference page. Can't we simply put this into the help document related to that preference page?

I guess I can remove the label and write a PR changing https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/master/eclipse.platform.common/bundles/org.eclipse.jdt.doc.user/reference/preferences/java/editor/ref-preferences-folding.htm once this is done.

danthe1st:
I have drafted a PR for the help here: https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2703

danthe1st:
I have squashed and rebased my PR onto the current master.

danthe1st:
Profiling results using a class containing 6000 inner classes similar to the following:
```java
    // #region outside
    /*
    * aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
    * aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
    * aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
    * aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
    * aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
    * aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
    * aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
    * aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
    * aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
    * aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
    * aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
    * aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
    * aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
    * aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
    * aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
    * aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
    * aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
    * aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
    * aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
    * aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
    */
    public static class InnerClass0 {
        // #region inside
        public int x = 42; // no error
        
        public void helloWorld() {
                System.out.println("Hello World!"); // no error
                System.out.println("In class " + 0 + ", x=" + x); // expected error
        }
        // #endregion inside
    } // end of InnerClass0
    // #endregion
```

sampling profiling of `opening the file` shows main thread actions, `other operations after opening` shows parts of the `JavaReconciler`.

### sampling
opening the file (custom regions enabled):
image

opening the file (custom regions disabled):
image

opening the file (master):
TODO


other operations after opening (custom regions enabled):
image

other operations after opening (custom regions disabled):
image

other operations after opening (master):
TODO

### profiling

opening the file (custom regions enabled):
image


opening the file (custom regions disabled):
image


opening the file (master):
TODO


other operations after opening (custom regions enabled):
image


other operations after opening (custom regions disabled):
image


other operations after opening (master):
TODO

danthe1st:
I want to note that I am still open to change the default region markers from `#region` and `#endregion` to `region` and `endregion` (comments would then look like `// region` and `// endregion` or similar).

BeckerWdf:
> image
> 
> I might still rework that label in the future.

Just another remark:
This is a workspace specific setting. Wouldn't it make sense to have that on the level of the project so that this settings are also checked into the git repo of the project? So that the folding is the same for all devs working together on the same project.

danthe1st:
> Just another remark: This is a workspace specific setting. Wouldn't it make sense to have that on the level of the project so that this settings are also checked into the git repo of the project? So that the folding is the same for all devs working together on the same project.

I guess this also makes sense to change.

danthe1st:
@BeckerWdf I have implemented project-specific preferences for folding.
If a third-party folding provider is not declaring support for per-project preferences by implementing `IScopedJavaFoldingPreferenceBlock` (an interface I created), it is displaying a message when opening the per-project preferences for that specific folding provider (this only happens when enabling project-specific settings _and_ selecting that folding provider in there).
image

BeckerWdf:
> @BeckerWdf @fedejeanne I think this PR is ready for a re-review if any of you can find the time to do it (otherwise I'm fine waiting more)

I am not a JDT committer. So I don't have the rights to approve the PR.

fedejeanne:
> > @BeckerWdf @fedejeanne I think this PR is ready for a re-review if any of you can find the time to do it (otherwise I'm fine waiting more)
> 
> I am not a JDT committer. So I don't have the rights to approve the PR.

@BeckerWdf non-committers can also approve PR's although the PR is not "_approved_" (green) so committers may miss your approval

@danthe1st I'll try to review it this week.

jjohnstn:
Hi @danthe1st without reviewing the source I have a few design comments:

1. I'm not sure using line comments is a robust solution.  The problem with line comments is that they usually get tied to statements below them.  This is a problem because cleanups or refactoring may remove a statement or move the statement.  In some cases, the comment tied to the statement is removed and in other cases the comment gets moved with the statement.  This isn't what you or the user is expecting.  Thinking about it, I think one solution is to use a block comment in an empty block.  For example:
```
    {/* folding region 1 */}
    int i = 3; // unused variable
    if (object instanceof Integer) {
       return 3;
    }
    {/* folding region 1 */}
```

In the case above, it doesn't matter if the unused variable gets removed or if any of the code block is moved because the empty blocks won't go with them or be removed themselves.  The other benefit is that this allows you to fix the name of the comment without having the user specify it.  It must be the only comment in an  empty block (no statements).  User code is never going to collide with this, especially if you give the comment a specific name.  In my example, I added numbers to do matching but you may prefer on/off or just the presence of the comment automatically toggles.  With this, no preferences for the triggers are needed as they are set in stone.  It would be nice to enhance the formatter to recognize these empty blocks and by default prevent moving the block end to the next line on a reformat.

2. Having to reopen a file is a major usability negative.  If the aforementioned idea removes the possibility of changing the marker, it is possible to simply always process.  If you need to make it optional due to poor performance, a possible solution would be to have a separate plug-in the user can install.  Otherwise, you need to find a way to reopen/reparse the file programatically.  You want the user to have the same experience as other folding.  

What do you think?

danthe1st:
1. My implementation actually allows any type of comment for this. If someone wants to use line comments, they can use line comments and if they prefer something else, they can do that. 
2. This is not really about my implementation but actually about how preferences are loaded from the `DefaultJavaFoldingStructureProvider`. This has been "fixed" in #1562.

danthe1st:
I added a test for your example (within a method):
```java
package org.example.test;
public class Test {
	void a(){
		{/* #region 1*/}
		System.out.println("Hello World");
		{/* #endregion 1*/}
	}
}
```
This example works but I won't note that using this style wouldn't work in non-local contexts because of https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/1825#issuecomment-2555349388 but I think this shouldn't be much of an issue:
```java
package org.example.test;
public class Test {
	{/* #region this is not recognized*/}
	void a(){
		System.out.println("Hello World");
	}
	{/* #endregion this is not recognized*/}
}
```
```java
package org.example.test;
public class Test {
	{/* #region because things like this shouldn't be possible*/}
	void a(){
		System.out.println("Hello World");
		{/* #endregion this is not recognized*/}
	}
}
```

danthe1st:
Since this PR contains both the change adding custom folding regions and making the folding preferences project-specific (see https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/1825#issuecomment-2565439652), I can split the latter off in its own PR if wanted.

danthe1st:
I have split off the part about project-specific properties from folding into #1979.

danthe1st:
I have changed the defaults to only require the text `region` instead of `#region` since I think the `#` symbol is a bit arbitrary.

org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/folding/CustomFoldingRegionNewFoldingTest.java
```diff
@@ -0,0 +1,189 @@
+/*******************************************************************************
```
danthe1st, line 1: This test class runs all tests from `testCommentsInEmptyBlocks` with the new folding introduced in #1562.
The test methods that yield different results (due to the folding being different in some ways, for example #1992) are overridden.

BeckerWdf:
Just something that to my mind: Why is this feature in JDT? Isn't independent from Java and could also be implemented in Platforms "plain" text editor?

danthe1st:
> Just something that to my mind: Why is this feature in JDT? Isn't independent from Java and could also be implemented in Platforms "plain" text editor?

JDT uses its own logic for folding already so I think implementing it somewhere else would result in it not being usable within the Java editor.

jjohnstn (changes requested):
Overall works quite well.  I noticed an issue with having a unended region start in a block which paired up with another endregion outside of the block.  I had a statement in the block (e.g. local variable declaration).  I wanted it to be only if the block has no statements (only comments).  If that is too hard to implement quickly then drop the block suggestion for now.

The preferences for region start should be brought up under the Custom Folding enablement button and indented.  They should read:

Region start comment text
Region end comment text

That way it is clear that it only applies to comments.  I like that you can have text after the comment so you can tag your regions (since you only see the region start comment when folded).

I will start to review the actual code immediately.   I would suggest you work on the above UI changes right away.  Understandably my code comments might not apply after your next round of changes but simply mark them as resolved since they are no longer relevant.  The cut-off for M3 is tomorrow at 5:00 so I'd like to try and get it in by then otherwise we have to wait until 4.36 M1.

jjohnstn:
> @jjohnstn Would you care to review the code at some point since you also reviewed other recent folding-related PRs?

Sure. 

> Overall works quite well. I noticed an issue with having a unended region start in a block which paired up with another endregion outside of the block. I had a statement in the block (e.g. local variable declaration). I wanted it to be only if the block has no statements (only comments). If that is too hard to implement quickly then drop the block suggestion for now.
> 

Hmm, some more testing.  It appears the regions start and ends are allowed to cross block boundaries if not properly nested.  End region can be in an inner block or an outer block to the start region comment.  So I guess if the behavior is documented, it should be ok for now.

org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/folding/CustomFoldingRegionNewFoldingTest.java
```diff
@@ -0,0 +1,215 @@
...
+					}
+					// endregion outside method
+				}
+				// endregion outside class
+				""";
+		List<IRegion> projectionRanges= getProjectionRangesOfFile(str);
+		assertEquals(4, projectionRanges.size());
+		FoldingTestUtils.assertContainsRegionUsingStartAndEndLine(projectionRanges, str, 1, 12);//outside class
+		FoldingTestUtils.assertContainsRegionUsingStartAndEndLine(projectionRanges, str, 3, 10);//outside method
+		FoldingTestUtils.assertContainsRegionUsingStartAndEndLine(projectionRanges, str, 4, 8);//void a()
```
jjohnstn, line 94: Why isn't the void a() method region 4, 9?

org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/folding/CustomFoldingRegionNewFoldingTest.java
```diff
@@ -0,0 +1,215 @@
...
+						// endregion
+						class Inner{
+
+						}
+					}
+				}
+				""";
+		List<IRegion> projectionRanges= getProjectionRangesOfFile(str);
+		assertEquals(2, projectionRanges.size());
+		FoldingTestUtils.assertContainsRegionUsingStartAndEndLine(projectionRanges, str, 3, 10);//void a()
```
jjohnstn, line 144: Again, I am confused on the end line value.  I assume you will clarify why the end-line matches the endregion comment for custom regions but the end brace line is not part of a method fold.

org.eclipse.jdt.ui/ui/org/eclipse/jdt/ui/text/folding/DefaultJavaFoldingStructureProvider.java
```diff
@@ -1011,8 +1027,13 @@ public void propertyChange(PropertyChangeEvent event) {
 	private boolean fCollapseInnerTypes= true;
 	private boolean fCollapseMembers= false;
 	private boolean fCollapseHeaderComments= true;
+	private boolean fCollapseCustomRegions= false;
 	private boolean fNewFolding;
 
+	private boolean fCustomFoldingRegionsEnabled= true;
```
jjohnstn, line 1033: Curious why this isn't false by default, but perhaps it will make sense later on.

jjohnstn (changes requested):
In my test, I turned off custom folding, but it did not remove the folding markers and they continue to work.

org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/folding/CustomFoldingRegionNewFoldingTest.java
danthe1st: ## TL;DR

If you have a method, showing the `}` when folding the method is wanted but with custom regions, there isn't much of a point in including the `// endregion` comment.

For the `CustomFoldingRegionTest`, it works differently due to #1825.

The main purpose of these tests is to check that the custom regions are working correctly so I wrote them to check against how it works in the repository so I am using the folding from master and add my own regions to it to check whether everything is folded correctly.

## Background

This is #1992 and these tests are how I discovered that issue. When a folding region is present before a method (e.g. a multiline comment or a custom region) and the "new" folding is disabled, the last line (`}`) is collapsed with the folding region.
With the new folding, this is not the case so the region ends one line earlier.

You can observe the behavior without my PR
```java
public class Test {
	/**
	 * 
	 */
	void a(){
		System.out.println("Hello World");
	}
}
```
When the `a()` method is folded, the `}` is hidden (by folding) if and only if the "Activate feature" option is disabled in the "Extended Folding" group.

org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/folding/CustomFoldingRegionNewFoldingTest.java
danthe1st: Again, showing the `}` makes sense for folded methods and was added by jakub-suliga's changes.

For custom folding regions, there is no point in showing the `// endregion` comment.

If the behavior of including the `// endregion` line is wanted, I can add that.

org.eclipse.jdt.ui/ui/org/eclipse/jdt/ui/text/folding/DefaultJavaFoldingStructureProvider.java
danthe1st: I just followed using the defaults in the method above. This shouldn't make much of a difference since it should be set by `initializePreferences()` but I guess it's better to be `false` since `fCustomFoldingRegionBegin` and `fCustomFoldingRegionEnd` are not initialized (fixed).

jjohnstn:
> > In my test, I turned off custom folding, but it did not remove the folding markers and they continue to work.
> 
> How did you turn it off? Note that the "Custom folding regions" checkbox is in the "Initially fold these elements" group and does not configure whether folding of custom regions is enabled. When I turn off custom region folding by clearing the text in one of the text fields in the "Custom folding regions" group, the folding regions disappear for me.
> 
> If the UX is confusing, I guess an option would be to add a checkbox to the "Custom folding regions" group about enabling/disabling it?

My bad.  The other items don't disable unless you disable all folding.

org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/folding/CustomFoldingRegionNewFoldingTest.java
jjohnstn: Ok.  Thanks for the explanation.  It works fine in usage.

org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/folding/CustomFoldingRegionNewFoldingTest.java
jjohnstn: Thanks.  Consider it resolved.

jjohnstn:
So, just the UX concerns remain.  The movement of the comment text inputs under Custom folding (indented) and the wording of the text fields.  Other than that, I think it is ready to merge.

danthe1st:
>  The movement of the comment text inputs under Custom folding (indented)

What exactly do you mean with that? Does that refer to the decision on whether the `// endregion` comment should be folded or not?

> wording of the text fields

I assume you are talking about whether to use `region` or `#region` as the default?

jjohnstn:
The strings above are for the description of the text fields for the region markers (not the text inside the fields).

danthe1st:
> I think that having the text fields in the top group right by the Custom folding checkbox is clearer.

Is there any reason for this? The top group is for _initially folding_ elements so I don't think adding extra configuration for custom folding regions there makes sense. The only thing these checkboxes are changing is whether these entries are collapsed or expanded by default.

>  I want the strings to read "Region start comment text" and "Region end comment text" which tells the user it is a comment.

Done

image

jjohnstn:
> > I think that having the text fields in the top group right by the Custom folding checkbox is clearer.
> 
> Is there any reason for this? The top group is for _initially folding_ elements so I don't think adding extra configuration for custom folding regions there makes sense. The only thing these checkboxes are changing is whether these entries are collapsed or expanded by default. The name `EDITOR_FOLDING_CUSTOM_REGIONS` should probably be `EDITOR_FOLDING_COLLAPSE_CUSTOM_REGIONS` or similar (I went with a name corresponding to the existing similar names).
> 

It is much cleaner. - one less group in the dialog.  The settings are only relevant to the custom folding checkbox so putting them there (indented) doesn't require the user to look elsewhere.   I am not a fan of the added group for the extended folding and eventually this should be merged above unless the individual items are broken out.

danthe1st:
Also note that currently, my tests use the current behavior from master which does not include the fixes from #2014 and #2015. Merging this PR will make #2014 and #2015 fail but fixing it should be very easy (they should just not be merged concurrently).

danthe1st:
> It is much cleaner. - one less group in the dialog. The settings are only relevant to the custom folding checkbox so putting them there (indented) doesn't require the user to look elsewhere. I am not a fan of the added group for the extended folding and eventually this should be merged above unless the individual items are broken out.

In that case, I guess it would be best to rename the existing group since the name "Initially fold these elements" would no longer be fitting (and custom folding regions also make sense to be used if that checkbox is disabled).

danthe1st:
Also, do you have any opinion on adding an additional checkbox for enabling/disabling folding as opposed to just letting users disabling by clearing out the Region start comment text/Region end comment text?

jjohnstn:
> > It is much cleaner. - one less group in the dialog. The settings are only relevant to the custom folding checkbox so putting them there (indented) doesn't require the user to look elsewhere. I am not a fan of the added group for the extended folding and eventually this should be merged above unless the individual items are broken out.
> 
> In that case, I guess it would be best to rename the existing group since the name "Initially fold these elements" would no longer be fitting (and custom folding regions also make sense to be used if that checkbox is disabled).
> 
> That being said, I could imagine not using groups for custom folding regions at all.

Not sure what you mean about the group name being inappropriate.  Are you are referring to the fact that the custom folding is regenerated if the text fields change so not truly initially at startup?  BTW: I merged those PRs you mentioned #2014 and #2015 as they were bug fixes that should have already been merged so thanks for the reminder about them.

danthe1st:
> Not sure what you mean about the group name being inappropriate.

The current preference screen looks like this:
image

I am assuming you are talking about the "Initially fold these elements" group. which contains a "Custom region folding" checkbox.

None of the checkboxes in that group enable or disable the addition of folding annotations at any point. The only difference these checkboxes make is whether these are collapsed by default or not.

If I enable all checkboxes in that group, it looks like this:
image

If I disable all of them, I get the following:
image

As you can see, the folding regions are added/processed either way. It's just that enabling these checkboxes collapses some parts when a file is opened. By default, this is the case for imports and header comments since once typically doesn't want to see those.

danthe1st:
> BTW: I merged those PRs you mentioned https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2014 and https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2015 as they were bug fixes that should have already been merged so thanks for the reminder about them.

I rebased my changes and adapted my tests to no longer use the special cases/behavior that should be fixed by these PRs.

jjohnstn:
Ok, I wasn't thinking clearly,  Yes, I see what you mean now.  I think that adding the enablement button makes sense.

danthe1st:
> I think that adding the enablement button makes sense.

I added it.

jjohnstn (approved):

jjohnstn:
@danthe1st Works great.  One more change is needed - the default for the enablement should be false.  Any new feature should do this.

danthe1st:
> @danthe1st Works great. One more change is needed - the default for the enablement should be false. Any new feature should do this.

Done

jjohnstn:
Thanks @danthe1st

jjohnstn:
Hi @danthe1st I will try and merge the patch as soon as 4.36 M1 opens up.  RC1 is next week and RC2 the following week after that.

fedejeanne:
> Looking at #2019, I used the same approach for the tests introduced with this PR ensuring I only need one test class.
> 
> @fedejeanne Should I add you in the copyright notice?

No need, but thank you for asking :-)

danthe1st:
I noticed that the line containing `void test(){}` would be folded in the following code:
```java
package org.example.test;

public class Test {
	// region

	/* endregion */ void test(){}
}
```
I changed it so that the last line is not included if there's other text in that line.

jjohnstn:
Hi @danthe1st.  Is this done?  I am not going to review it if you are going to make further changes.  I would like to get it into 4.36 M1 and then you can tweak it if you like once people have a chance to try it and perhaps open issues as needed.

danthe1st:
> Hi @danthe1st. Is this done? I am not going to review it if you are going to make further changes. I would like to get it into 4.36 M1 and then you can tweak it if you like once people have a chance to try it and perhaps open issues as needed.

Yes, I consider it done. I just bumped the version of `org.eclipse.jdt.ui` to `3.35.0` in order to make the build pass.

jjohnstn:
Thanks @danthe1st

Issue #1827, opened 2024-12-03
hover logs OperationCanceledException as error


jukzi:
Would be better to silently ignore OperationCanceledException:

```
eclipse.buildId=4.35.0.I20241201-1800
java.version=21.0.2
java.vendor=Eclipse Adoptium
BootLoader constants: OS=win32, ARCH=x86_64, WS=win32, NL=de_DE
Command-line arguments:  -os win32 -ws win32 -arch x86_64 -data file:/C:/Users/jkubitz/platform-2024-09-23/ws/

org.eclipse.jdt.ui
Error
Tue Dec 03 12:18:10 CET 2024
OperationCanceledException computing hover information in NullReferenceImplTransformations.java at offset 100231

org.eclipse.core.runtime.OperationCanceledException
	at org.eclipse.jdt.internal.core.AbstractClassFile.getClassFileContent(AbstractClassFile.java:239)
	at org.eclipse.jdt.internal.core.ClassFile.getJarBinaryTypeInfo(ClassFile.java:225)
	at org.eclipse.jdt.internal.core.ClassFile.getBinaryTypeInfo(ClassFile.java:186)
	at org.eclipse.jdt.internal.core.ClassFile.buildStructure(ClassFile.java:78)
	at org.eclipse.jdt.internal.core.Openable.generateInfos(Openable.java:246)
	at org.eclipse.jdt.internal.core.SourceRefElement.generateInfos(SourceRefElement.java:127)
	at org.eclipse.jdt.internal.core.JavaElement.openWhenClosed(JavaElement.java:569)
	at org.eclipse.jdt.internal.core.BinaryType.getElementInfo(BinaryType.java:292)
	at org.eclipse.jdt.internal.core.JavaElement.getElementInfo(JavaElement.java:278)
	at org.eclipse.jdt.internal.core.BinaryType.getElementInfo(BinaryType.java:1142)
	at org.eclipse.jdt.internal.core.BinaryType.getFlags(BinaryType.java:357)
	at org.eclipse.jdt.internal.core.SearchableEnvironmentRequestor.acceptType(SearchableEnvironmentRequestor.java:127)
	at org.eclipse.jdt.internal.core.NameLookup.seekTypesInBinaryPackage(NameLookup.java:1510)
	at org.eclipse.jdt.internal.core.NameLookup.seekTypes(NameLookup.java:1442)
[...]

PR #1829, opened 2024-12-04
Silent OperationCanceledException during hover #1827

jukzi:
If Operation is canceled result is just not needed anymore. No need to log an error.

* also silent BadLocationException

https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1827


jukzi:
same as `// Just swallow the exception if the operation was canceled` in `org.eclipse.jface.text.TextViewerHoverManager.computeInformation()`

jukzi:
ignoring unrelated know 12 fails https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1833

PR #1851, opened 2024-12-13
[Help Needed] Implement provider for sticky scrolling in JAVA/JDT

Christopher-Hermann:
## What it does
The current sticky scrolling functionality is based solely on indentation. This can cause issues when the source code is formatted in unexpected ways. To address this, an extension point for providing language-specific sticky lines has been introduced in the following pull request: https://github.com/eclipse-platform/eclipse.platform.ui/pull/2398.

This change represents the first proof-of-concept implementation for sticky lines in Java/JDT. As I lack experience in JDT development, the proposed approach for calculating sticky lines in Java may not be optimal.
I would greatly appreciate assistance from the community in identifying a more effective method for calculating sticky lines.

Fixes:
https://github.com/eclipse-platform/eclipse.platform.ui/issues/1950
https://github.com/eclipse-platform/eclipse.platform.ui/issues/2128
https://github.com/eclipse-platform/eclipse.platform.ui/issues/2338

## How to test
1. Check out the PR from https://github.com/eclipse-platform/eclipse.platform.ui/pull/2398. and open plugin org.eclipse.ui.editors 
2. Enable the Sticky Scrolling Feature via Preferences > General >  Editors > Text Editors > Enable Sticky Scrolling
3. Open the class EnumMap, scroll to line 775 (method writeObject)

With proper Sticky Lines calculation: 
image

With the Default Sticky Lines calculation based on indentation:
image

### DIFF
--- a/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/javaeditor/StickyLinesProviderJava.java
+++ b/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/javaeditor/StickyLinesProviderJava.java
@@ -1,0 +1,51 @@
+package org.eclipse.jdt.internal.ui.javaeditor;
+
+import java.util.LinkedList;
+import java.util.List;
+
+import org.eclipse.jface.text.BadLocationException;
+import org.eclipse.jface.text.source.ISourceViewer;
+
+import org.eclipse.ui.texteditor.stickyscroll.IStickyLine;
+import org.eclipse.ui.texteditor.stickyscroll.IStickyLinesProvider;
+import org.eclipse.ui.texteditor.stickyscroll.StickyLine;
+
+import org.eclipse.jdt.core.IJavaElement;
+import org.eclipse.jdt.core.ISourceRange;
+import org.eclipse.jdt.core.ISourceReference;
+import org.eclipse.jdt.core.JavaModelException;
+
+public class StickyLinesProviderJava implements IStickyLinesProvider {
+
+	@Override
+	public List<IStickyLine> getStickyLines(ISourceViewer sourceViewer, int lineNumber, StickyLinesProperties properties) {
+		LinkedList<IStickyLine> stickyLines= new LinkedList<>();
+		JavaEditor javaEditor= (JavaEditor) properties.editor();
+
+		IJavaElement element= null;
+		try {
+			element= javaEditor.getElementAt(sourceViewer.getDocument().getLineOffset(lineNumber));
+		} catch (BadLocationException e) {
+			// TODO Auto-generated catch block
+			e.printStackTrace();
+		}
+
+		while (element != null) {
+			if (element.getElementType() == IJavaElement.METHOD || element.getElementType() == IJavaElement.TYPE) {
+				try {
+					ISourceRange sourceRange= ((ISourceReference) element).getNameRange();
+					int offset= sourceRange.getOffset();
+					int stickyLineNumber= sourceViewer.getDocument().getLineOfOffset(offset);
+					stickyLines.addFirst(new StickyLine(stickyLineNumber, sourceViewer));
+				} catch (JavaModelException | BadLocationException e) {
+					// TODO Auto-generated catch block
+					e.printStackTrace();
+				}
+			}
+			element= element.getParent();
+		}
+
+		return stickyLines;
+	}
+
+}
--- a/org.eclipse.jdt.ui/plugin.xml
+++ b/org.eclipse.jdt.ui/plugin.xml
@@ -7830,4 +7830,18 @@
            file-extensions="class without source">
      </file-association>
   </extension>
+  <extension
+         point="org.eclipse.ui.editors.stickyLinesProviders">
+      <stickyLinesProvider
+            class="org.eclipse.jdt.internal.ui.javaeditor.StickyLinesProviderJava"
+            id="org.eclipse.jdt.internal.ui.StickyLinesProviderJava">
+         <enabledWhen>
+            <and>
+               <with variable="editor">
+                  <instanceof value="org.eclipse.jdt.internal.ui.javaeditor.JavaEditor"/>
+               </with>
+            </and>
+         </enabledWhen>
+      </stickyLinesProvider>
+   </extension>
 </plugin>


BeckerWdf:
Is JDT interested in such a feature?
@jukzi: Could you ask one of the JDT committers to have a look at this?

jukzi:
what kind of help do you need? build currently fails because it can not find the internal(!) API
`org.eclipse.ui.internal.texteditor.stickyscroll.IStickyLine`
referenced as
`org.eclipse.ui.texteditor.stickyscroll.IStickyLine`

BeckerWdf:
@Christopher-Hermann asks for this:

> This change represents the first proof-of-concept implementation for sticky lines in Java/JDT. As I lack experience in JDT development, the proposed approach for calculating sticky lines in Java may not be optimal.
I would greatly appreciate assistance from the community in identifying a more effective method for calculating sticky lines.

Could you or some of your JDT co-committers help him?

jukzi:
> Could you or some of your JDT co-committers help him?

I lack the skill for that. Maybe @jjohnstn could help?

org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/javaeditor/StickyLinesProviderJava.java
```diff
@@ -0,0 +1,51 @@
+package org.eclipse.jdt.internal.ui.javaeditor;
```
mickaelistria, line 1: Please add copyright header.

Christopher-Hermann: Done

BeckerWdf:
ping @jjohnstn 
Can you help here?

jjohnstn:
@BeckerWdf I'll take a look

jjohnstn (changes requested):
Hi.  I agree that indent level is insufficient information for many Java files.  However, using the JavaModel is also insufficient.  As you have found, you can only show Types and Methods which is less functionality than currently exists for a normally indented piece of source code (e.g. if statements,else blocks,  for statements, etc... are shown).

What you need to do is parse the compilation unit and get an AST.  With the AST, you can get the detail needed.  For example, you may find the parent of the current statement is a block but your logic can skip over blocs and find the next parent which could be an if statement, for statement, etc... that you can choose to display.  The indentation doesn't matter.

There is code in Checks.Java:
```

	public static CompilationUnit convertICUtoCU(ICompilationUnit compilationUnit) {
		ASTParser parser= ASTParser.newParser(IASTSharedValues.SHARED_AST_LEVEL);
		parser.setKind(ASTParser.K_COMPILATION_UNIT);
		parser.setSource(compilationUnit);
		parser.setResolveBindings(true);

		return (CompilationUnit) parser.createAST(null);
	}
```
as an example or parsing an ICompilationUnit to get a CompilationUnit.  I do not believe you need to resolve bindings so the set call in the example should be false.  From there you can find any ASTNode from its position using NodeFinder or use a custom ASTVisitor.  You can define what elements you want to show (e.g. if statement, else clause, etc...) vs ignore (e.g. block, variable declaration).

mickaelistria:
Could the sticky scrolling rely on folding ranges (when available), or the underlying strategy folding is built on, to compute which lines to keep? That would probably be sufficient in most cases; and that would provide good factorization.

jjohnstn:
@Christopher-Hermann I have created an AST-based version of Java StickyLinesProvider in #2149   Could you try it out and let me know what you think?  It looks for various Java AST constructs (e.g. for, if, try, catch, lambda expression) instead of indentation.

Christopher-Hermann:
> @Christopher-Hermann I have created an AST-based version of Java StickyLinesProvider in #2149 Could you try it out and let me know what you think? It looks for various Java AST constructs (e.g. for, if, try, catch, lambda expression) instead of indentation.

Thanks a lot, that's really awesome. I will check as soon as I have time.

jjohnstn:
> > @Christopher-Hermann I have created an AST-based version of Java StickyLinesProvider in #2149 Could you try it out and let me know what you think? It looks for various Java AST constructs (e.g. for, if, try, catch, lambda expression) instead of indentation.
> 
> Thanks a lot, that's really awesome. I will check as soon as I have time.

If you like it, I think it should be optional to use.  Perhaps a sticky lines provider could have a isEnabled() method so that it can check a preference in addition to the extension checking.

Christopher-Hermann:
I tested a litte bit and it looks quite promising.

I found some minor issues (The line numbers are somehow broken on my current setup, it should not be related to sticky scrolling):

1. The method name is moved into the sticky lines already for java doc. 
Bildschirmfoto 2025-04-16 um 09 59 34
I'm not sure if this is really an issue, I just noticed it

2. Inner classes are "overwriting" the sticky line of the main class
Bildschirmfoto 2025-04-16 um 09 57 15
Bildschirmfoto 2025-04-16 um 09 57 22

Regarding the isEnalbed() method: I will check in the next days if it is worth to add such a method or if we can somehow model this via the stickyLinesProviders enhancement point.

Thanks again for your help 👍

BeckerWdf:
> Perhaps a sticky lines provider could have a isEnabled() method so that it can check a preference in addition to the extension checking.

So from a user's perspective we would have the general setting in preferences to enable disable sticky scrolling completely and then have another setting so switch if a user would like to use JDT's sticky scrolling implementation of the "fallback" implementation (based on indentation). As I user would find this confusing. And I also don't see a reason why a Java developer would *not* like to use the Java-specific implementation. 
So let's don't over-engineer this and don't provide a isEnabled() method. If we later find out we really need it we can always add it later.

jjohnstn:
> > Perhaps a sticky lines provider could have a isEnabled() method so that it can check a preference in addition to the extension checking.
> 
> So from a user's perspective we would have the general setting in preferences to enable disable sticky scrolling completely and then have another setting so switch if a user would like to use JDT's sticky scrolling implementation of the "fallback" implementation (based on indentation). As I user would find this confusing. And I also don't see a reason why a Java developer would _not_ like to use the Java-specific implementation. So let's don't over-engineer this and don't provide a isEnabled() method. If we later find out we really need it we can always add it later.

The reason I suggested enablement is that each time we enter, we need to create an AST for the entire compilation unit.  Caching isn't an option if the file is writable unless we can be told that no edits have occurred or no edits before the current position have been made.  For extremely large files, getting an AST each time could cause a noticeable delay and for files with errors (possibly cascading to before the current position) this will exhibit issues that the indentation version does not.  I wanted to have a way for a user to have a fall-back in such cases.

A possible alternative to enablement is to make the default sticky provider accessible to the JDT version and if the file is too big, we punt to the indentation implementation.  If needed, this size could be configurable for the user.

BeckerWdf:
> and for files with errors (possibly cascading to before the current position) this will exhibit issues that the indentation version does not.

Couldn't you implement the sticky scrolling provider in a way that it automatically falls back to the indentation based implementation when it detects that there are syntax errors (above the current position)?

jjohnstn:
> > and for files with errors (possibly cascading to before the current position) this will exhibit issues that the indentation version does not.
> 
> Couldn't you implement the sticky scrolling provider in a way that it automatically falls back to the indentation based implementation when it detects that there are syntax errors (above the current position)?

That was my suggestion at the end of my comment.  If the default sticky scrolling provider could be extended, I could use a super call when needed.  It might lead to some weirdness because the origin of this requirement is that some files are indented with braces on the next line which causes braces to be shown instead of meaningful statements.  So, when errors appear, we might back off into braces being shown - users might understand this though.

Having the ability to extend the default sticky provider using indentation also provides the ability to add a separate enablement in JDT UI, if desired.

jjohnstn:
> I tested this change for a large Java source with 12000 lines of code. Even after enabling the preference "Enable sticky scrolling", the sticky lines area was not visible. I have not seen any sticky lines for this large source. I checked the result values of StickyLinesProviderJava#getStickyLines in the debugger and they looked ok for me; result was not empty and I expected to see sticky lines but did not.
> 
> I then introduced following optimization which does not call `javaEditor.getElementAt` each time the getStickyLines method is called. Instead a simple and stupid cache is used which just remembers the javaElement from the last call (if it is a method) and checks if it can be re-used for the new requested line (see method canReuseLastElement). With this change, the sticky lines are visible and the feature works as expected. Maybe you could check if this approach might be applicable here to improve the performance?
> 

I think that caching/reuse in the last file is a good idea.  I was also going to try caching class file asts to prevent creating them if the file has already been edited.  I will put something together today and update the patch.

jjohnstn:
Ok, I have added caching of Compilation Units and time stamps to my patch.  It is based on the type root.  Works considerably better for my tests using JDK HashMap and EnumMap class files.  At the moment the caches are never cleared.  I think a listener for editor update/close is in order to prevent bloat.

I am currently still determining the sticky lines from scratch (i.e. get the current line node then use node parentage), but as mentioned the editor and sticky feature is running much better without any reuse of previous sticky line info. A problem existed with regular for loops that had an empty expression and I have fixed that after noticing it pop up in HashMap.  

Occasionally I am seeing a model modification exception thrown when I try to get the working copy, but I am doing no changes.  Not sure if there is an error in my logic or there is a platform error, or whether my target platform is causing this.  I was getting this error before when I tried discarding the working copy I got for a class file, but I added a check to prevent this.

Christopher-Hermann:
> Having the ability to extend the default sticky provider using indentation also provides the ability to add a separate enablement in JDT UI, if desired.

Yes, I think this is a good idea. If there are no concerns I will move the Default provider in the public package as well.

BeckerWdf:
> > Having the ability to extend the default sticky provider using indentation also provides the ability to add a separate enablement in JDT UI, if desired.
> 
> Yes, I think this is a good idea. If there are no concerns I will move the Default provider in the public package as well.

So that other providers can inherit from the default provider implementation? Inheriting does create a very hard coupling and might it harder to change the default provider later on (because methods are part of the API).

Christopher-Hermann:
> So that other providers can inherit from the default provider implementation?

Yes exactly. 

Another option would be to make the DefaultStickyLinesProvider a final class so that other implementations can fall back by calling DefaultStickyLinesProvider#getStickyLines. But you are right, even this approach is introducing some ugly dependency which makes it harder to change behavior in the DefaultStickyLinesProvider.

BeckerWdf:
> > So that other providers can inherit from the default provider implementation?
> 
> Yes exactly.
> 
> Another option would be to make the DefaultStickyLinesProvider a final class so that other implementations can fall back by calling DefaultStickyLinesProvider#getStickyLines. But you are right, even this approach is introducing some ugly dependency which makes it harder to change behavior in the DefaultStickyLinesProvider.

What about adding a simply API that implementers can call that wraps this "calling DefaultStickyLinesProvider#getStickyLines"? Would that be better / more flexible or is that engineered?

jjohnstn:
@Christopher-Hermann Could you try the latest version of the patch?  I made a few fixes and made it faster.

Christopher-Hermann:
> @Christopher-Hermann Could you try the latest version of the patch? I made a few fixes and made it faster.

Looks really good now.

I noticed that the sticky line appears when the JavaDoc scrolls out of view. I'm not sure if this is good or not, since it does allow for the class name to remain visible when the JavaDoc is lengthy and would hide the class itself. But on the first look it feels odd. Maybe there are some other opinions on that? @BeckerWdf You're usually pretty intuitive about things like this. What do you think?
image

@jjohnstn I'm wondering if we still need access to the DefaultStickyLines provider? The latest version feels really fast now and for me it seems that we don't need the fallback any longer. In this case we could avoid exporting an additional API.

BeckerWdf:
> Maybe there are some other opinions on that?

I also think that this feels wrong. Especially when the javadoc is not so long. I thinks handling this should be improved.

Christopher-Hermann:
I compared the JDT implementation with the implementation in Eclipse Theia:

### Java doc
JDT is showing sticky lines on java doc, Eclipse Theia not:
Bildschirmfoto 2025-05-09 um 08 14 02

### Inner classes
JDT is showing the inner classes as root
Bildschirmfoto 2025-05-09 um 08 14 29

For both I think we should go for the same behavior as in Eclipse Theia. 
- Show sticky lines after the java doc
- Always show the outer and inner classes as sticky lines

BeckerWdf:
> For both I think we should go for the same behavior as in Eclipse Theia.
> 
> * Show sticky lines after the java doc
> * Always show the outer and inner classes as sticky lines

I agree

jjohnstn:
> > For both I think we should go for the same behavior as in Eclipse Theia.
> > 
> > * Show sticky lines after the java doc
> > * Always show the outer and inner classes as sticky lines
> 
> I agree

I'll fix it.

jjohnstn:
> > > For both I think we should go for the same behavior as in Eclipse Theia.
> > > 
> > > * Show sticky lines after the java doc
> > > * Always show the outer and inner classes as sticky lines
> > 
> > 
> > I agree
> 
> I'll fix it.

Actually, I'm not seeing the behavior you are mentioning.  I have taken the platform.ui patch and rebased on master and rebased my latest patch and there is no forward mentions of the class while the Javadoc is being scrolled.  I also don't see an issue with inner classes overriding the top class.  Could you try the latest patch (force pushed after rebasing on master jdt.ui) with the platform.ui patch rebased on master?  I am using Linux (Fedora 42).  Perhaps there is a difference on Windows or Mac (which I don't have access to)?

screenshot1

screenshot2

screenshot3

Christopher-Hermann:
> Could you try the latest patch

Ah sorry, you are right. It does work with the latest patch sets. I might have accidentally overlooked one of the changes.

Only difference I see on MacOS is how comments and java doc is handled:
image
image

I guess that it's not that hard to removed them. If you don't have already an idea how this can be fixed I will have a look if I can find it out. But I guess it should be fine even if we cannot fix it.

Another issue I found on the latest patch are comments starting at the first line, which are basically throw away all the sticky lines:
image

Thanks once again for your help and your work on this implementation. It’s greatly appreciated!

org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/javaeditor/StickyLinesProviderJava.java
```diff
@@ -0,0 +1,64 @@
...
+	public List<IStickyLine> getStickyLines(ISourceViewer sourceViewer, int lineNumber, StickyLinesProperties properties) {
+		LinkedList<IStickyLine> stickyLines= new LinkedList<>();
+		JavaEditor javaEditor= (JavaEditor) properties.editor();
+
+		IJavaElement element= null;
+		try {
+			element= javaEditor.getElementAt(sourceViewer.getDocument().getLineOffset(lineNumber));
+		} catch (BadLocationException e) {
+			// TODO Auto-generated catch block
+			e.printStackTrace();
```
danthe1st, line 43: Just commenting as someone interested in this feature: I guess you plan to change the error handling to write the exception to the Error Log?

jjohnstn:
> > Could you try the latest patch
> 
> Ah sorry, you are right. It does work with the latest patch sets. I might have accidentally overlooked one of the changes.
> 
> Only difference I see on MacOS is how comments and java doc is handled: image image
> 
> I guess that it's not that hard to removed them. If you don't have already an idea how this can be fixed I will have a look if I can find it out. But I guess it should be fine even if we cannot fix it.
> 
> Another issue I found on the latest patch are comments starting at the first line, which are basically throw away all the sticky lines: image
> 
> Thanks once again for your help and your work on this implementation. It’s greatly appreciated!

Hi, thx for the update.  I could use some help debugging this on MacOS which I have no access to.  It isn't occurring for Linux.  I don't know what is happening in your first example.  I have noticed that the AST for the binary class file is marking the start of the Javadoc as the start of some nodes such as Type Declaration, however, my code doesn't report the start line, it gets the Name node of the Type Declaration (which in this case is HashMap) and uses that line number for the sticky line.  Thus, the Javadoc comment start should never have been used for anything unless there is some bug in the AST.  I don't have any Javadoc cases in the switch.  You might try adding println statements for the lines it is processing and then see where it ends up in the Switch statement below after finding the node.

Christopher-Hermann:
> Hi, thx for the update.

Sorry for the delayed response; I've been quite busy and haven't been able to do any eclipse contribution recently.

I have now updated my target IDE and rebased the PR. With the latest changes, everything is working correctly. It seems that the issue was related due to a temporary bug in the version I was using.

So I would suggest to close this PR and continue in #2149, what do you think?

jjohnstn:
New issue is #2239

Issue #1685, opened 2024-09-29
Remove (some) code relating to Java versions below 8


stephan-herrmann:
Since JDT/Core dropped support for Java versions below 8, the following queries in JavaModelUtil are now de-facto constants:
* is1d8OrHigher() -> always true
* is1d2OrHigher(), is1d4OrHigher(), is50OrHigher(), is1d7OrHigher() -> always false

I came here when checking if JDT/UI still "uses" `org.eclipse.jdt.annotation_v1`. In theory, the quickfix `UnresolvedElementsBaseSubProcessor.addNullityAnnotationTypesProposals()` would still try to locate this bundle in pre-Java8 projects, but since such projects are no longer supported, the version dependence in that method can be removed.

akurtakov:
Actually is1d20orHigher and etc. are also always true now due to the "orHigher"

akurtakov:
I consider this one done. Whatever remains can be handled separately as linked PRs here handled the "obvious" cases.

PR #1877, opened 2024-12-22
Handle < > brackets unconditionally

akurtakov:
Remove checks whether running on Java 1.5+ to handle them as ECJ mandates 1.8 thus this is always true.
Removed tests for pre 1.5 behavior and enable tests in JavaPairMatcherTest that were not executed at all.


akurtakov:

PR #1878, opened 2024-12-22
Stop assuming Java 1.3 source if project doesn't set explicitly

akurtakov:
ECJ supports 1.8 as min java version thus this would prevent the case of UI generating/accepting code that the compiler may reject at compile time.


akurtakov:

PR #1879, opened 2024-12-22
Default to JAVA 8

akurtakov:
Ecj supports min Java 8 so testing/supporting older versions in UI creates the risk of generating/accepting code that compiler may reject.


akurtakov:

PR #1880, opened 2024-12-23
Stop support for pre 1.8 source in some editor features

akurtakov:
This is only code level cleanup as it is impossible for anyone to use as this was consulting project which can't not be pre 1.8 already.


akurtakov:

PR #1881, opened 2024-12-23
Simplify JavaCodeScanner

akurtakov:
Java 8 is the min version ECJ supports thus having complicated way to add "assert"(1.4) and "enum"(1.5) is not needed and can be added to default keywords.


akurtakov:

PR #1882, opened 2024-12-26
Always propose generics

akurtakov:
Java 8 is the min version supported by ECJ thus generics are always supported.


akurtakov:

PR #1883, opened 2024-12-26
Projects are always at 1.8+ level

akurtakov:
As ECJ no longer supports lower levels.


akurtakov:
Part of https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/1882

PR #1886, opened 2024-12-26
Remove pre Java 1.5 checks form o.e.jdt.ui/core extensions and friends

akurtakov:
ECJ mandates Java 8 level thus this code is always noop.


akurtakov:
Part of https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/1885

PR #1889, opened 2024-12-28
Tests should not explicitly set pre 1.8 compliance

akurtakov:
ECJ no longer supports it.


akurtakov:

PR #1891, opened 2024-12-29
Remove pre 1.8 support

akurtakov:
ECJ supports min 1.8 and JDT UI provides no way to set BREE lower than that thus all this code is non reachable.


akurtakov:

PR #1892, opened 2024-12-30
Simplify "string" generation/cleanups by dropping pre 1.5

akurtakov:
ECJ mandates 1.8+ and JDT UI converts compiler settings to that unconditionally so this code is simply non-reachable.


akurtakov:

PR #1893, opened 2024-12-30
Remove pre1.6 support from string generation

akurtakov:
ECJ no longer supports pre 1.8 and JDT UI simply converts project settings to 1.8 if lower thus all this code is non reachable.


akurtakov:
https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1685

PR #1894, opened 2024-12-30
Remove last Java 5 checks from UI

akurtakov:
Having code to show errors if Java 5 JVM when Java UI migrates projects to Java 8 is pointless.


akurtakov:
https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1685

PR #1910, opened 2025-01-08
More removals of JavaModelUtil.is1d8OrHigher

akurtakov:
Ecj mandates 1.8 thus such guarded code is not reachable


akurtakov:

PR #1912, opened 2025-01-08
Remove copied code from JavaModelUtils

akurtakov:
and its usages which are just constants now that Java .1.8+ is mandated.


akurtakov:

Issue #1860, opened 2024-12-17
[performance] StandardJavaElementContentProvider.getParent(Object) blocks IDE startup


laeubi:
org.eclipse.jdt.ui.StandardJavaElementContentProvider.getParent(Object) calls this code

https://github.com/eclipse-jdt/eclipse.jdt.ui/blob/4dc72d8c0d1ea3ca58656c098ecd178269a0695a/org.eclipse.jdt.ui/ui/org/eclipse/jdt/ui/StandardJavaElementContentProvider.java#L246

the existence check currently triggers all classpath containers of all projects to be resolved what is a costly operation and blocks the whole IDE startup procedure see here:
- https://github.com/eclipse-pde/eclipse.pde/issues/1481

instead one needs a more lightweight check or even always return `null` as allowed by the API or a better way to determine the parents.

jukzi:
I would like to see initialize deferred, but i don't know why you are looking at getPartent() specificly. As far as i understand any line between e4.compatibility and initializeContainer could be blamed:

```
        at org.eclipse.pde.internal.core.RequiredPluginsInitializer.initialize(RequiredPluginsInitializer.java:50)
        at org.eclipse.jdt.internal.core.JavaModelManager.initializeContainer(JavaModelManager.java:3209)
        at org.eclipse.jdt.internal.core.JavaModelManager.getClasspathContainer(JavaModelManager.java:2136)
        at org.eclipse.jdt.core.JavaCore.getClasspathContainer(JavaCore.java:3970)
        at org.eclipse.jdt.internal.core.JavaProject.resolveClasspath(JavaProject.java:3127)
        at org.eclipse.jdt.internal.core.JavaProject.resolveClasspath(JavaProject.java:3291)
        at org.eclipse.jdt.internal.core.JavaProject.getResolvedClasspath(JavaProject.java:2405)
        at org.eclipse.jdt.internal.core.JavaProject.buildStructure(JavaProject.java:488)
        at org.eclipse.jdt.internal.core.Openable.generateInfos(Openable.java:246)
        at org.eclipse.jdt.internal.core.Openable.openAncestors(Openable.java:508)
        at org.eclipse.jdt.internal.core.Openable.generateInfos(Openable.java:228)
        at org.eclipse.jdt.internal.core.JavaElement.openWhenClosed(JavaElement.java:569)
        at org.eclipse.jdt.internal.core.JavaElement.getElementInfo(JavaElement.java:292)
[...]

PR #1916, opened 2025-01-09
Perform element changes always in a non UI thread

laeubi:
Currently PackageExplorerContentProvider receives events for changed elements from different threads and already has handling to perform only certain actions in the UI thread. But it does not really account for the case when the event itself is triggered from the UI thread in which case lengthy operation can take place in the UI.

This now checks if the change is using the UI and otherwise schedules a job to perform the required actions in the background.

See for example:
- https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1860


org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/packageview/PackageExplorerContentProvider.java
```diff
@@ -127,6 +128,18 @@ protected Object getViewerInput() {
 
 	@Override
 	public void elementChanged(final ElementChangedEvent event) {
+		IJavaElementDelta delta= event.getDelta();
+		if (Display.findDisplay(Thread.currentThread()) == null) {
+			processDelta(delta);
+		} else {
+			Job.create(PackagesMessages.PackageExplorerContentProvider_update_job_description, m -> {
```
vogella, line 135: Should this be scheduled as system job so that the user is not notified of this update?

jukzi:
JavaElementDelta is not thread safe. the initialization may take place after it is used by another thread :-(

org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/packageview/PackageExplorerContentProvider.java
laeubi: Maybe yes ...

iloveeclipse:
This test is failing: ContentProviderTests3.testDeleteBottomLevelFragmentFolding
Which is OK considering the code changes the behavior and test runs in UI thread by default, but it has to be adopted.

Beside this (and missing "system" job flag) the change goes in the right direction.

The delta itself should be immutable after sending, so I do not see issues with that, however there might be multiple deltas, and so the **dedicated** job should be re-scheduled with the updated list of deltas to process, something like org.eclipse.jdt.ui.ProblemsLabelDecorator.AdornmentCalculationJob.

jukzi:
> The delta itself should be immutable

The members need to be marked final or synchronized

laeubi:
I have poked around a bit more and actually the source seem to be the java editor performs "MakeWorkingCopy" operation and this triggers the an POST CHANGE event and then the whole thing runs inside the context of the Editor

I created now
- https://github.com/eclipse-jdt/eclipse.jdt.core/pull/3545

to see what happens if we perform it async but it seems all not very optimal here, the Javadoc of the `IJavaElementDelta` say it is only valid during the callback, so actually one would need to copy all interesting data first.

Beside that I get the impression that the actual event is not really interesting for the package explorer anyways, but I have not enough insights to judge if thats the case and how it can be detected earlier.

jukzi:
I think you step back a bit to think what should happen/shown: When the javamodel is not up to date and things are displayed based on an outdated model they are either wrong drawn or need to be drawn again once the model is updated. Wrong data shown is wrong. Draw again may reduce a ui freeze but may in sum perform slower. The only clean solution would be to not show  javamodel depended stuff until the model is (asnychronous) updated and then draw the java depended stuff once finished. That can not be archived by just deferring some calls but you would need to explicitly render java independend first. Possibly an explicit "loading..." text instead of showing a wrong (not) formatted sourcecode

laeubi:
Feel free to all implement that, I can only tell that the code here do not is designed to run in the UI, so this is not about presentation (what already is rendered in the UI once the changes are applied).

laeubi:
Fails existing tests...

laeubi:
Maybe someone can pick this up as an idea an polish this, I'm not sure how to best adjust the test or handle the concerns regarding usage of the Delta outside of event notification.

iloveeclipse:
> Beside this, this one was the most promising experiment as indeed the UI is actually shown (even though other tasks are still execute in the UI) in contrast to that the UI is not shown before.

I strongly believe that this is doable and right way to go & encourage you to follow up on this PR.

laeubi:
I tried to adjust the one failing test but for me it seems the testcase itself depends on other places on synchronous event execution as well...

org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/packageview/PackageExplorerContentProvider.java
```diff
@@ -127,6 +128,20 @@ protected Object getViewerInput() {
 
 	@Override
 	public void elementChanged(final ElementChangedEvent event) {
+		IJavaElementDelta delta= event.getDelta();
+		if (Display.findDisplay(Thread.currentThread()) == null) {
+			processDelta(delta);
+		} else {
+			Job job= Job.create(PackagesMessages.PackageExplorerContentProvider_update_job_description, m -> {
```
iloveeclipse, line 135: As said, that shouldn't be just "fire and forget" with N jobs running in parallel. it should be one job that has a queue of deltas to process in exact order the deltas were fired. This code should put delta to the queue and single job instance should be triggered to process it.

laeubi: As this is the UI thread there are never n-jobs only one... Delta firing has no real order also right now because each thread uses its own DeltaProcessor....

laeubi:
@iloveeclipse I think this should be merged before as it seems the main reason for the test failure:

- https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/1925

laeubi:
I hopefully have now addressed all concerns.

jukzi (changes requested):
I still think, that JavaElementDelta has to be made threadsafe first (i.e. declare members final) and that after Deltas are applied the UI has to be refreshed somehow to not show an outdated state. The later should be verified by a Junit test.

laeubi:
> I still think, that JavaElementDelta has to be made threadsafe first 

I would then assume this will be taken by a JDT developer in advance, this PR is all I can currently offer here and it solves the issue for my testcase. @iloveeclipse

laeubi:
@iloveeclipse can you pick this up and decide:

- Is anything left or can it be merged?
- Should I close the PR as there is no way to make progress here?

PR #1918, opened 2025-01-09
Simplify latest JLS version in astview

akurtakov:
Instead of inlining (with all drawbacks from that) the JLS_LATEST constant remove the inlined constant and use AST.getJLSLatest()


akurtakov:
I have manually did the version bump as the versions had minor of they type x.y.z50 which is supposed to be for beta branches only (aka issue with merge) .

PR #1923, opened 2025-01-10
Move o.e.jdt.core.manipulation to Java 21

akurtakov:
Needed by
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2739


akurtakov:
Warnings increase due to new deprecations in Java 21 (e.g. URL) which don't have straightforward fix. Merging to move on with the build at least.

PR #1925, opened 2025-01-10
Unify event triggering in ContentProviderTests3

laeubi:
Currently non trivial code is used in ContentProviderTests3 to trigger a delta event in 5 places. In one of them it is even done wrong where the assert is part of the eventloop because of missing brackets.


jukzi:
thanks for contributing

Issue #1922, opened 2025-01-10
Poor Performance of Quick Outline View


fedejeanne:
(See [issue #2](https://github.com/vi-eclipse/Eclipse-JDT/issues/2) our internal backlog)

### Current Behavior

The quick outline view (Ctrl+O) opens very slowly when the opened class is part of a large class hierarchy. The root cause seems to be a combination of two things:
1. Loading the complete view is performed synchronously, thus the UI freezes while loading it.
2. The action calculates the complete inheritance hierarchy of involved classes in order to show the "method overwrite indicator" labels for methods that overwrite methods of superclasses. This information is not cached in the AST and is calculated everytime the quick outline view is opened.

The ordinary outline view calculates the overwrite indicator labels asynchronously by spawning a decoration job and thus provides the same information but without blocking the UI.

### Expected Behavior

The quick outline view shall open instantly and calculate and add additional information asynchronously. It should use the same behavior as in the ordinary outline view to improve responsiveness of the UI.
While loading the information in background, there should be an indicator that the view has not completely loaded all expected information in the quick outline view yet.

### Existing Workaround

Disable "Java Method Overwrite Indicator" label.

![8455de56-acf9-4dab-a1ad-1c94e1bfcced](https://github.com/vi-eclipse/Eclipse-JDT/assets/755472/cafa609b-719b-4dcc-abcd-029f596b89f2)

### Samples and reproducer
To reproduce, do the following:
- Import this Java project: [reproducer.zip](https://github.com/user-attachments/files/18372865/reproducer.zip) 
- Make sure the "Java Method Overwrite Indicator" is active
- Go to the class `Broke1` and click `Ctrl` + `o` **twice**

This is the sample I got: [Broke1-I20241230-1800.zip](https://github.com/user-attachments/files/18372834/Broke1-I20241230-1800.zip)

I also sampled in our big WS (**~2.000** plugins and fragments), it takes **757s** to open the quick outline view: 
[MyECU_2025-03-M1.zip](https://github.com/user-attachments/files/18372916/MyECU_2025-03-M1.zip)

jukzi:
Can you please try to give a simple selfcontaining reproducer like https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3327#issuecomment-2515120465 instead of attachments. Also please add screenshots of sampling result (call tree + hotspots), ideally add stacktrace of hotspot as text.

jukzi:
[...]

PR #1929, opened 2025-01-11
Asynchronously decorate quick outline view via DecorationManager #1922

HeikoKlare:
## What it does

The Quick Outline View currently applies (i.e., calculates and draws) decorations for override indicators synchronously. In case the calculation of the indicator takes long, it blocks the UI. The ordinary Outline View defers the responsibility of calculating and applying decorations to the `DecorationManager`. This ensures that (1) that task is performed asynchronously via a dedicated job and (2) that the `DecorationManager` takes care of the current user configuration (i.e., whether those indicators shall be shown or not), which the Quick Outline View currently replicates.

With this change, the Quick Outline View uses the same label decorator as the ordinary Outline View, which makes the `DecorationManager` process the override label decorator asynchronously instead of applying it synchronously.

Contributes to https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1922

This complements https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/1926, which improves performance of the business logic used for calculating labels of the view, while this one ensures that the business logic is completely moved outside of the UI thread to be executed asynchronously.

## How to test

A reproducer has been kindly provided by @fedejeanne in:
- #1922

To reproduce, do the following:
- Import this Java project: [reproducer.zip](https://github.com/user-attachments/files/18372865/reproducer.zip)
- Make sure the "Java Method Overwrite Indicator" is active
- Go to the class `Broke1` and click `Ctrl + O` twice

Without this patch, the quick outline view will take a high amout of time to open and block the UI while with this patch it will open immediately and stay responsible.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/text/JavaOutlineInformationControl.java
+++ b/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/text/JavaOutlineInformationControl.java
@@ -52,7 +52,6 @@ import org.eclipse.jface.viewers.TreeViewer;
 import org.eclipse.jface.viewers.Viewer;
 import org.eclipse.jface.viewers.ViewerFilter;
 
-import org.eclipse.ui.IDecoratorManager;
 import org.eclipse.ui.IEditorPart;
 import org.eclipse.ui.IWorkbenchPage;
 import org.eclipse.ui.PlatformUI;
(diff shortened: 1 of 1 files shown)


HeikoKlare:
@jukzi @fedejeanne this one complements https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/1926, which improves business logic of the quick outline view image calculation, while this one extracts the business logic out of the UI thread. Would be great if you could have a look and integrate this as well.

HeikoKlare:
Yes, it still fixes that PR. Here is a screencast to demonstrate the behavior:
quick_outline_override_indicator

The asynchronously executed decorator has been introduced to the ordinary outline view in 2008, so quite some time after the mentioned issue has been resolved:
https://github.com/eclipse-jdt/eclipse.jdt.ui/blob/35d2546655bc8abbd4470a2750b8bd30781038e9/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/javaeditor/JavaOutlinePage.java#L949

Maybe the quick outline was just forgotten to be adapted as well.

jukzi:
ok, if nobody complains i would merge it tomorrow.

PR #1933, opened 2025-01-14
fix executor resource warning since 19

jukzi:
`Potential resource leak: 'executor' may not be closed at this location
`


jukzi:
image

PR #1946, opened 2025-01-16
Deprecate for removal JavaPlugin.getJavaCorePluginPreferences

akurtakov:
Class org.eclipse.core.runtime.Preferences has been deprecated for years and method getJavaCorePluginPreferencesNew (returning IEclipsePreferences) is there.
Migrated unit tests to use getJavaCorePluginPreferencesNew.


akurtakov:
image

PR #1948, opened 2025-01-17
Use only supported AST levels in ASTView

akurtakov:
Ensure that only AST levels supported by the compiler are used.


noopur2507 (commented):
The change looks good. We can also update the AST view to support only 1.8 and above i.e. remove versions prior to 1.8 in a separate issue.

PR #1950, opened 2025-01-18
Migrate ContentAssistHistory to IEclipsePreferences

akurtakov:
Long deprecated o.e.core.runtime.Preferences class would better not be used anymore.


akurtakov:
@jjohnstn I would appreciate second pair of eyes here.

PR #1952, opened 2025-01-19
Check if platform is running in SuppressWarningsSubProcessor

laeubi:
I'm currently working on an automated way to fix certain warnings in the platform code. One of that kind is 

```
Unnecessary @SuppressWarnings("restriction")	PatternFilter.java	/org.eclipse.equinox.p2.ui.discovery/src/org/eclipse/equinox/internal/p2/ui/discovery/util	line 33	Java Problem
```

but encountered a problem because currently apply this quickfix outside of a "real" IDE yields:

> java.lang.IllegalStateException: Workbench has not been created yet.

because the quickfix try to fetch a shared image.

This now adds a guard around the call for a shared image to allow to proceed in this case with a (possibly not optimal) image instead of completely fail.


laeubi:
The two "new warnings" seem not related to my change @akurtakov can you review/merge?

akurtakov:
The 2 new deprecations seems to come from https://github.com/eclipse-jdt/eclipse.jdt.core/pull/3572 so this is expected.

akurtakov (approved):

PR #1958, opened 2025-01-21
Adjust to JavaCore.defaultRootModules deprecation

akurtakov:
Use the new method that filters based on the release. If release option is not set, use target and if it's not there too use the latest version.


akurtakov:
Ah, I missed that . New PR is in the prep.

stephan-herrmann:
> Ah, I missed that . New PR is in the prep.

thanks

PR #1960, opened 2025-01-22
Adding test cases for folding

jakub-suliga:
I have added test cases to validate the current folding mechanism.

### DIFF
--- a/org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/folding/FoldingTest.java
+++ b/org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/folding/FoldingTest.java
@@ -1,0 +1,260 @@
+/*******************************************************************************
+ * Copyright (c) 2025 Vector Informatik GmbH and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Vector Informatik GmbH - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.jdt.text.tests.folding;
+
+import java.util.List;
+
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+
+import org.eclipse.jdt.testplugin.JavaProjectHelper;
+
+import org.eclipse.core.runtime.CoreException;
+
+import org.eclipse.jface.text.IRegion;
+
+import org.eclipse.jdt.core.IJavaProject;
+import org.eclipse.jdt.core.IPackageFragment;
+import org.eclipse.jdt.core.IPackageFragmentRoot;
+
+import org.eclipse.jdt.ui.tests.core.rules.ProjectTestSetup;
+
+public class FoldingTest {
+	@Rule
+	public ProjectTestSetup projectSetup= new ProjectTestSetup();
+
+	private IJavaProject jProject;
+
+	private IPackageFragmentRoot sourceFolder;
+
+	private IPackageFragment packageFragment;
+
+	@Before
+	public void setUp() throws CoreException {
+		jProject= projectSetup.getProject();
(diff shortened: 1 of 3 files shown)


jakub-suliga:
@fedejeanne @iloveeclipse Could you review this?

danthe1st:
I assume you will need different tests for your folding of control structures PR (#1562).
Given that (and that I will also need to add test cases related to folding in my PR adding custom folding blocks via comments (#1825), I would suggest creating one test suite for folding tests that delegates to the different test classes for different purposes (for now just your `FoldingTest`).

For example, this could be
```java
@Suite
@SelectClasses({
	FoldingTest.class,
	// to be added in the future - not for your PR, I am just including that for your understanding
	//ControlStructureFoldingTest.class,
	//CustomFoldingRegionTest.class
})
public class FoldingTestSuite {
}
```

It should then be possible to reference that test suite in `JdtTextTestSuite`.

org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/folding/FoldingTest.java
```diff
@@ -0,0 +1,283 @@
...
+import org.eclipse.jdt.core.IPackageFragment;
+import org.eclipse.jdt.core.IPackageFragmentRoot;
+import org.eclipse.jdt.core.JavaModelException;
+
+import org.eclipse.jdt.ui.tests.core.rules.ProjectTestSetup;
+
+import org.eclipse.jdt.internal.ui.javaeditor.EditorUtility;
+import org.eclipse.jdt.internal.ui.javaeditor.JavaEditor;
+
+public class FoldingTest {
```
fedejeanne, line 55: I would also add more thorough tests for JavaDocs and tests for comment blocks and copyright headers. 

```java
@Test
public void testJavadocs() throws Exception {
	String str= """
			package org.example.test;
		    /**										//here should be an annotation
		     * Javadoc
		     */
		     /**										//here should be an annotation
		     * Another Javadoc
		     */
		     /**										//here should be an annotation
		     * Yet another Javadoc
		     */
		     public class Example {}
			""";
	assertCodeHasRegions(str, 3);
}

@Test
public void testCommentBlocks() throws Exception {
	String str= """
			package org.example.test;
		    /* //here should be an annotation
			 *
			 */
			/* //here should be an annotation
			 *
			 */
			/* //here should be an annotation
			 *
			 */
			class h {

				/* //here should be an annotation
				 *
				 */
				void b() { //here should be an annotation
					/* //here should NOT be an annotation
					 *
					 */
					int a;
				}
			}
			""";
	assertCodeHasRegions(str, 5);
}

@Test
public void testCopyrightHeader() throws Exception {
	String str= """
			/**	//here should be an annotation
			* This is some copyright header
			*/
			package org.example.test;

			class SomeClass {}
			""";
	assertCodeHasRegions(str, 1);
}
```

I hope I didn't miss anything.

fedejeanne (commented):
Looks pretty good, I think it covers most of the existing cases today. I would add some more test cases though.

org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/folding/FoldingTest.java
jakub-suliga: Done

jakub-suliga:
> I assume you will need different tests for your folding of control structures PR (#1562). Given that (and that I will also need to add test cases related to folding in my PR adding custom folding blocks via comments (#1825), I would suggest creating one test suite for folding tests that delegates to the different test classes for different purposes (for now just your `FoldingTest`).

@fedejeanne @iloveeclipse What do you think of this? I’m capable of implementing it, but I don’t have enough experience to decide if this is the best way to structure the tests.

jakub-suliga:
@jjohnstn Could you please review this code too? I have added the checks for the region values as you suggested in #1562. I will also add these checks in the other PR later. Additionally, could you let me know if the above proposal looks good?

jjohnstn:
> @jjohnstn Could you please review this code too? I have added the checks for the region values as you suggested in #1562. I will also add these checks in the other PR later. Additionally, could you let me know if the above proposal looks good?

Looks good.  I will merge if the CodeQL check goes through or doesn't finish in the next few minutes.

PR #1961, opened 2025-01-22
If no editor is open apply the resolution directly

laeubi:
## What it does

Currently if one applies a QuickFix without the java editor open (e.g. from the Problems view), then JDT opens a new editor, applies the quickfix and then leave the user in a dirty state, so the error/warning is still shown in the problems view unless one manually save the editor.

This now changes the behavior by apply the resolution directly on the document and commit (or save) the compilation unit.
## How to test

I use this simple test class:

```
@SuppressWarnings("restriction")
public class TestWarning {

}
```

it will produce:

```
Unnecessary @SuppressWarnings("restriction")	TestWarning.java   ...	line 4	Java Problem
```

1. Open the file and apply the quickfix in the editor --> works as before with dirty editor
2. No restore the file and make sure all editors are closed
3. Make sure all editors are closed
4. Got to the problem view and rightclick on the warning choose "QuickFix" to open the Wizard

grafik

You will end up with the problem resolved and no dirty editor.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/text/correction/CorrectionMarkerResolutionGenerator.java
+++ b/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/text/correction/CorrectionMarkerResolutionGenerator.java
@@ -59,7 +59,6 @@ import org.eclipse.jdt.core.compiler.IProblem;
 import org.eclipse.jdt.internal.corext.fix.CleanUpRefactoring.MultiFixTarget;
 import org.eclipse.jdt.internal.corext.util.JavaModelUtil;
 
-import org.eclipse.jdt.ui.JavaUI;
 import org.eclipse.jdt.ui.cleanup.ICleanUp;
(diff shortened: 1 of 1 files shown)


akurtakov:
@jjohnstn What do you think? I am in favor of this approach but want to be sure that there isn't some obscure detail missed.

laeubi:
Build is green!

jukzi (changes requested):
I have no opinion if this change is any good but i think all Quickfixes should behave the same. So if changing this, then also change org.eclipse.jdt.ui.text.java.correction.CUCorrectionProposal.apply(IDocument) and org.eclipse.jdt.internal.ui.text.correction.CorrectionMarkerResolutionGenerator.CorrectionMarkerResolution.run(IMarker[], IProgressMonitor) the same way

jukzi:
also
org.eclipse.jdt.internal.ui.text.correction.proposals.CreatePackageInfoWithDefaultNullnessProposal.apply(IDocument)

laeubi:
If this is fixed (and proven to work) then yes one can maybe change other places as well, I just have no test-case to validate any other problem right now.

PR #1964, opened 2025-01-23
Improve the instanceof pattern matching refactoring

gzsombor:
Check if the smart casting can be used in deeper parts of the AST, not just directly as the first expression, which is surprisingly common.

## What it does
Improve the logic which tries to locate casts that can be eliminated with pattern matching

## How to test

Run the refactoring on a bigger codebase

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/fix/PatternMatchingForInstanceofFixCore.java
+++ b/org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/fix/PatternMatchingForInstanceofFixCore.java
@@ -186,13 +186,13 @@ public class PatternMatchingForInstanceofFixCore extends CompilationUnitRewriteO
 					return variableName != null && !statementsToRemove.isEmpty();
 				}
 
-				void addMatching(final VariableDeclarationStatement statementToRemove, final SimpleName expressionToMove, boolean toConvert) {
-					if (this.variableName == null || this.variableName.getIdentifier().equals(expressionToMove.getIdentifier())) {
-						this.variableName = expressionToMove;
+				void addMatching(final VariableDeclarationStatement statement, final SimpleName name, boolean toConvert) {
+					if (this.variableName == null || this.variableName.getIdentifier().equals(name.getIdentifier())) {
+						this.variableName = name;
 						if (toConvert) {
-							this.statementsToConvert.add(statementToRemove);
+							this.statementsToConvert.add(statement);
 						} else {
-							this.statementsToRemove.add(statementToRemove);
+							this.statementsToRemove.add(statement);
 						}
 					}
 				}
--- a/org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/fix/PatternMatchingForInstanceofFixCore.java
+++ b/org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/fix/PatternMatchingForInstanceofFixCore.java
@@ -24,9 +24,12 @@ import org.eclipse.text.edits.TextEditGroup;
 import org.eclipse.jdt.core.dom.AST;
 import org.eclipse.jdt.core.dom.ASTNode;
 import org.eclipse.jdt.core.dom.ASTVisitor;
+import org.eclipse.jdt.core.dom.Assignment;
 import org.eclipse.jdt.core.dom.Block;
 import org.eclipse.jdt.core.dom.CastExpression;
 import org.eclipse.jdt.core.dom.CompilationUnit;
+import org.eclipse.jdt.core.dom.Expression;
+import org.eclipse.jdt.core.dom.ExpressionStatement;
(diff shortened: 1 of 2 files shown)


gzsombor:
Could you please have a look on it @jjohnstn ? I think you are the most familiar with this code that I tried to change.

jjohnstn:
@gzsombor Will do

jjohnstn (changes requested):
Overall works well, but there are scenarios you have to protect against.  You need to ensure that the local variable doesn't get modified or passed to another method where it could be altered.  For example:

```
	public void foo(Object x, boolean valid) {
		if (x instanceof Integer) {
			int k = 7;
			if (valid) {
				Integer i = (Integer)x;
				i = 7;
				System.out.println(i + k);
			}
			Integer i = (Integer)x;
			System.out.println(i);
		}
	}

```

In this example, you can modify the instanceof variable in the if statement and then expect it to be the original value outside.  The logic should be to only modify the first reference unless all references are read-only.  You have to treat a method call as possibly modifying arguments unless you know details of the class and it cannot be modified.

jjohnstn:
> Thanks for the feedback! Your example code is pure evil, I love it!
> 
> Maybe for that case, the correct code would be something like this:
> 
> ```
> 	public void foo(Object x, boolean valid) {
> 		if (x instanceof Integer i) {
> 			int k = 7;
> 			if (valid) {
> 				i = 7;
> 				System.out.println(i + k);
> 			}
> 			i = (Integer)x;
> 			System.out.println(i);
> 		}
> 	}
> ```
> 
> But in this case, even if the code is a bit shorter, I'm not sure, I would be happy to review this change. So maybe, if we detect some strange assignments, the best solution could be just to not touch the code.

I think the code above is the correct way to go.  Simply change as many references as possible and leave the rest as casts.

gzsombor:
I could improve the code, so at least this scenario is handled properly.

org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/fix/PatternMatchingForInstanceofFixCore.java
```diff
@@ -142,55 +145,127 @@ public boolean visit(final InstanceofExpression visited) {
...
-					VariableDeclarationStatement variableDeclarationExpression= ASTNodes.as(statements.get(0), VariableDeclarationStatement.class);
-					VariableDeclarationFragment variableDeclarationFragment= ASTNodes.getUniqueFragment(variableDeclarationExpression);
+				boolean hasResult() {
+					return variableName != null && !statementsToRemove.isEmpty();
+				}
 
-					if (variableDeclarationFragment != null
-							&& Objects.equals(visited.getRightOperand().resolveBinding(), variableDeclarationExpression.getType().resolveBinding())) {
-						CastExpression castExpression= ASTNodes.as(variableDeclarationFragment.getInitializer(), CastExpression.class);
+				void addMatching(final VariableDeclarationStatement statementToRemove, final SimpleName expressionToMove, boolean toConvert) {
```
jjohnstn, line 189: Minor nit: rename statementToRemove to be "statement", rename expressionToMove to be "name".  Much simpler.

org.eclipse.jdt.ui.tests/ui/org/eclipse/jdt/ui/tests/quickfix/CleanUpTest16.java
```diff
@@ -540,6 +540,223 @@ public void doNotMatchBitWiseOr(boolean useStrikethroughForCompleted, Object dat
...
+				new HashSet<>(Arrays.asList(MultiFixMessages.PatternMatchingForInstanceofCleanup_description)));
+	}
+
+	@Test
+	public void testDeeperMatches() throws Exception {
+		IPackageFragment pack= fSourceFolder.createPackageFragment("test1", false, null);
+		String given= """
+			package test1;
+
+			public class E {
```
jjohnstn, line 655: Please add a scenario where you have 2 instanceofs separated by and.  This PR fixes another bug where 2 instanceofs can be converted in one cleanup and your fix handles it.

e.g. 
```
	public void foo(Object o, Object p) {
		if (o instanceof String && p instanceof Integer) {
			if (o != p) {
				String s = (String) o;
				Integer i = (Integer) p;
				System.out.println(s.trim() + i.toString());
				i = 3;
			}
			Integer i = (Integer) p;
			i = 7;
		}
	}
```

jjohnstn (changes requested):
Works very well.  It even fixes another issue.  I had one small nit (see comments) and wanted one test scenario added so the other issue has a test case.  Once that is done, I will merge.

org.eclipse.jdt.ui.tests/ui/org/eclipse/jdt/ui/tests/quickfix/CleanUpTest16.java
gzsombor: I'm really surprised, but yes, the code handles this too.

PR #1965, opened 2025-01-23
Make AST view show only supported JLS versions + modernize code

akurtakov:
As compiler no longer supports pre Java 8 it is of no help if AST view
shows AST for unsupported JLS versions.


noopur2507 (commented):
Thanks.

PR #1972, opened 2025-01-24
Add new feature to convert pattern instanceof to switch

### LINKED ISSUE #1919: Cleanup/Refactor: if-else-chains to pattern-matching-switch (if only instanceof checks used)
We have a lot of code which looks something like this:
```java
public String getText(final Object o) {
    if (o instanceof final Foo foo) {
        return foo.getName();
    }
    else if (o instanceof final Bar bar) {
        return bar.somethingElse();
    }
    else if (o instanceof final Baz baz) {
        return baz.another();
    }
    else {
        return "";
    }
}
```
This pattern is especially prevalent in SWT related implementations, like LabelProviders, e.g. in getText or getImage.
With Java 21+ it would be nice to have a cleanup which transforms the above if-else chain into something like this:

```java
public String getText(final Object o) {
    switch (o) {
        case final Foo foo -> {
            return foo.getName();
        }
        case final Bar bar -> {
            return bar.somethingElse();
        }
        case final Baz baz -> {
            return baz.another();
        }
        case null, default -> {
            return "";
        }
    }
}
```

Or even directly to a switch expression, though the required analysis for this one is probably a bit harder.
```java
public String getText(final Object o) {
    return switch (o) {
        case Foo foo -> foo.getName();
        case Bar bar -> bar.somethingElse();
        case Baz baz -> baz.another();
        case null, default -> "";
    };
}
```

jjohnstn:
- add new PatternInstanceofToSwitchCleanUpCore and PatternInstanceofToSwitchFixCore classes to convert a nesting if statement using pattern instanceof expressions into either a switch statement or switch expression if possible
- add new cleanup to JavaFeatureTabPage, plugin, and CleanUpConstants
- add rtstubs_21.jar for testing
- add new CleanUpTest21 class and add tests for new feature
- fixes #1919

## What it does

See issue.

## How to test

See issue or new tests.

## Author checklist


jjohnstn:
Compiler tools issues have nothing to do with this patch.

PR #1976, opened 2025-01-24
Fix release option in module dialog/page

akurtakov:
Fixes broken commit c3f2219b2e480995adca0f3e28fa17da2e3894f8 . Fixed approach used here given in
https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/1958#issuecomment-2613058811


akurtakov:
Merging now to prevent previous totally broken commit to be in a build.

PR #1979, opened 2025-01-24
project property page for folding

danthe1st:
## What it does

This PR introduces a project properties page for Java editor folding as suggested in https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/1825#issuecomment-2565439652

This was originally part of #1825 but I split it into two PRs.

image
 

## How to test

- Go to `Preferences` > `Java` > `Editor` > `Folding`.
- On the top right, an option "Configure Project specifc settings" should appear

- Go to `Project` > `Properties` > `Java Editor` > `Folding`
- The folding settings can be configured per project

## Author checklist

  - The "Enable Folding" option still seems to cause some issues


danthe1st:
^ I added a missing import (might have happened in the last rebase) and then rebased my changes onto master).

jjohnstn:
I'm seeing funky things happening with the extended folding preferences.  It seems to be toggling instead of looking at the enabled/disabled state.  When I try settting in the project settings, it doesn't turn on after Apply and Close.  If I set the Workspace preferences and hit apply, it works right away.  If I try turning it off and hitting apply again, it doesn't do anything.  If I then turn on again in workspace, it turns off in the file.  The project settings should take over entirely and the workspae should be ignored if set.

danthe1st:
> I'm seeing funky things happening with the extended folding preferences. It seems to be toggling instead of looking at the enabled/disabled state. When I try settting in the project settings, it doesn't turn on after Apply and Close. If I set the Workspace preferences and hit apply, it works right away. If I try turning it off and hitting apply again, it doesn't do anything. If I then turn on again in workspace, it turns off in the file. The project settings should take over entirely and the workspae should be ignored if set.

The issue here was that enabling/disabling these preferences required restarting the editor.
I now changed it such that the `DefaultJavaFoldingStructureProvider` also updates itself automatically when the the project properties are enabled/disabled.

One thing that's still left is completely enabling/disabling folding and switching the folding provider. Doing that cannot be handled by the folding provider and is handled by the `JavaEditor`. For this, it's _currently_ still necessary to restart the editor in order to apply the changes.

danthe1st:
I fixed the issues.
`JavaEditor` has a listener that is executed when the corresponding preference changes and toggles whether folding is active. This only included workspace preferences and project properties from jdt.core. I changed it to also include jdt.ui properties.

I also ensured that the properties are set in a way that the listener there "sees" it and I changed the folding provider selection such that it is only available in the workspace preferences since having different folding providers for different projects would require more API changes. However, it is possible to configure any folding provider per-project, it is not just possible to use different folding providers for different projects in the same workspace.

jjohnstn (approved):
LGTM and new version fixes the wonkiness I was seeing.

jjohnstn:
Thanks @danthe1st  If you are willing, please add a N&N entry for the new functionality.

danthe1st:
@jjohnstn Should I create it for the property page only or also for the custom folding regions feature?

danthe1st:
I just created a PR for both.

Issue #1992, opened 2025-01-31
Code after Javadocs still folded "too much"


danthe1st:
While testing #1825, I noticed the following issue:

In #1562, @jakub-suliga introduced the fix for #1939 (see [this comment](https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/1562#issuecomment-2626504845)) but there seems to be a case where this issue still happens.

With the following code, folding `a()` causes `b()` to be invisible as well in the current master (after merging #1562) if the new/experimental folding is disabled:

```java
package org.example.test;

class X {
	/*
	 * a b
	 */
	void a() {
	} void b() {
	}
}
```

danthe1st:
Also, it seems like that PR changes the test in #1960 to only consider the "new"/experimental folding and does not run any tests with the "old" folding (I expected it to run tests with both options).

danthe1st:
Also @jjohnstn since you reviewed the PR in question, I think you should know about that as well.

jakub-suliga:
> Also, it seems like that PR changes the test in [#1960](https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/1960) to only consider the "new"/experimental folding and does not run any tests with the "old" folding (I expected it to run tests with both options).
> 
> [@jakub-suliga](https://github.com/jakub-suliga) Is this intentional?

No, it isnt intentional. I fixed it in #1993

jakub-suliga:
> With the following code, folding a() causes b() to be invisible as well in the current master (after merging https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/1562) if the new/experimental folding is disabled:

The PR #1993 fixes the porblem too. I accidentally set includeLastLine to true a second time, which caused the issue.

PR #1993, opened 2025-01-31
Fix folding of 2 consecutive methods

jakub-suliga:
If the second method starts on the same line where the first one ends, there will still be two foldings. (I accidentally set includeLastLine to true a second time, which caused the issue).

Additionally, I have separated the tests: FoldingTest verifies the old behavior, while FoldingTestNew tests my feature (#1562).

Fixes https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1992

### DIFF
--- a/org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/JdtTextTestSuite.java
+++ b/org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/JdtTextTestSuite.java
@@ -21,6 +21,7 @@ import org.eclipse.jdt.text.tests.codemining.CodeMiningTriggerTest;
 import org.eclipse.jdt.text.tests.codemining.ParameterNamesCodeMiningTest;
 import org.eclipse.jdt.text.tests.contentassist.ContentAssistTestSuite;
 import org.eclipse.jdt.text.tests.folding.FoldingTest;
+import org.eclipse.jdt.text.tests.folding.FoldingTestNew;
 import org.eclipse.jdt.text.tests.semantictokens.SemanticTokensProviderTest;
 import org.eclipse.jdt.text.tests.spelling.SpellCheckEngineTestCase;
 import org.eclipse.jdt.text.tests.templates.TemplatesTestSuite;
@@ -71,6 +72,7 @@ import org.eclipse.jdt.text.tests.templates.TemplatesTestSuite;
 	JavaElementPrefixPatternMatcherTest.class,
 	CodeMiningTriggerTest.class,
 	ParameterNamesCodeMiningTest.class,
+	FoldingTestNew.class,
 	FoldingTest.class,
 })
 public class JdtTextTestSuite {
--- a/org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/folding/FoldingTestNew.java
+++ b/org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/folding/FoldingTestNew.java
@@ -24,15 +24,20 @@ import org.eclipse.jdt.testplugin.JavaProjectHelper;
 
 import org.eclipse.core.runtime.CoreException;
 
+import org.eclipse.jface.preference.IPreferenceStore;
+
 import org.eclipse.jface.text.IRegion;
 
 import org.eclipse.jdt.core.IJavaProject;
 import org.eclipse.jdt.core.IPackageFragment;
 import org.eclipse.jdt.core.IPackageFragmentRoot;
 
+import org.eclipse.jdt.ui.PreferenceConstants;
 import org.eclipse.jdt.ui.tests.core.rules.ProjectTestSetup;
 
-public class FoldingTest {
+import org.eclipse.jdt.internal.ui.JavaPlugin;
+
+public class FoldingTestNew {
 	@Rule
 	public ProjectTestSetup projectSetup= new ProjectTestSetup();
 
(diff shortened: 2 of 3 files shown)


org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/folding/FoldingTestNew.java
```diff
@@ -0,0 +1,501 @@
...
+	@Before
+	public void setUp() throws CoreException {
+		jProject= projectSetup.getProject();
+		sourceFolder= jProject.findPackageFragmentRoot(jProject.getResource().getFullPath().append("src"));
+		if (sourceFolder == null) {
+			sourceFolder= JavaProjectHelper.addSourceContainer(jProject, "src");
+		}
+		packageFragment= sourceFolder.createPackageFragment("org.example.test", false, null);
+		IPreferenceStore store= JavaPlugin.getDefault().getPreferenceStore();
+		store.setValue(PreferenceConstants.EDITOR_NEW_FOLDING_ENABLED, true);
```
danthe1st, line 59: I think it's better to use `store.setToDefault(PreferenceConstants.EDITOR_NEW_FOLDING_ENABLED);` in `tearDown()` to ensure the state is reset for all tests.
Then you wouldn't need to explicitly set the preference in `FoldingTest`.

danthe1st:
While you are at it, I noticed that local classes (defined in methods) are folded with the old folding but not with your new folding:
```java
class Outer {
	void a() {
		class Inner2{
			
		}
	}
}
```

org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/folding/FoldingTest.java
```diff
@@ -56,7 +56,7 @@ public void setUp() throws CoreException {
 		}
 		packageFragment= sourceFolder.createPackageFragment("org.example.test", false, null);
 		IPreferenceStore store= JavaPlugin.getDefault().getPreferenceStore();
-		store.setValue(PreferenceConstants.EDITOR_NEW_FOLDING_ENABLED, true);
+		store.setValue(PreferenceConstants.EDITOR_NEW_FOLDING_ENABLED, false);
 		}
```
danthe1st, line 60: By the way if you are already changing this, I think this `}` should be indented less.

danthe1st:
I think the following is still different between your the "old" folding and your "new" AST-based folding.

```java
public class Test {
	void a() {
		int i;

		class Inner {

		}
	}
}
``` 
Here, class `Inner` is folded with the old version but not with the new one.

jakub-suliga:
> that multiline comments are folded within methods:

This is intentional, I want that every comment can be folded

org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/folding/FoldingTestNew.java
jakub-suliga: Done

org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/folding/FoldingTest.java
jakub-suliga: Done

jakub-suliga:
> While you are at it, I noticed that local classes (defined in methods) are folded with the old folding but not with your new folding.
In the following example, Inner2 doesn't seem to be folded if your new folding is activated.

Fixed.

> Given that (at least if this is done), it may even be worth considering to add a folding annotatation for Outer but I guess this may need more discussion.

The plan is for outer to also receive a folding annotation. When I initially tried to implement this, I encountered some issues and decided to put it on hold at the time.

jakub-suliga:
@fedejeanne @jjohnstn could you look over the code?

org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/JdtTextTestSuite.java
```diff
@@ -71,6 +72,7 @@
 	JavaElementPrefixPatternMatcherTest.class,
 	CodeMiningTriggerTest.class,
 	ParameterNamesCodeMiningTest.class,
+	FoldingTestNew.class,
```
fedejeanne, line 75: Hopefully, you will not need another class and you can parameterize `FoldingTest` (see my comment in the review). In any case, the name `FoldingNewTest` would have been better. But please let's move this discussion to another PR.

fedejeanne (changes requested):
I haven't looked deep into the code but I'd like for you to please:
- Only solve 1 issue in this PR: the folding of 2 consecutive methods
- Create another PR to rename/move the tests and split testing the "old" folding and the "new" folding (if even necessary).

A hint for the other PR (the one about the tests): you don't need to create a new test class, you can parameterize the tests to use the "old" and/or the "new" folding: the deciding part is probably `store.setValue(PreferenceConstants.EDITOR_NEW_FOLDING_ENABLED, x);` with `x` being the parameter for the tests. With that, you can simply write all the tests, skip some if necessary (_i.e._ if the "old" folding doesn't work for a given structure) and adapt the assertions. That way you will probably avoid having duplicated test code (_e.g._ `testinnerClassinMethod`) and you'll make sure that no one forgets to add tests for any of the foldings in the future since all the test methods are in the same class.

Issue #1996, opened 2025-02-03
IndexOutOfBoundsException in OrganizeImportsAction


noopur2507:
```
package p1;


public class Snippet {
	void main() {
	  List<String> list = List.of("the", "be", "two", "of", "and", "a", "in", "that");
	  System.out.println(list);
	}
	
	private <T> ArrayList<T> filtering(Predicate<T> predicate) {
	  return null;
	  Code language
	}
	
}

```

* In a Java 24 project, have the above code.
* Invoke Organzie Imports.
* On the first page of the wizard, select java.util.List, press Next.
* On the second page to select a type for Code, press Skip.
* On the third page, select java.util.function.Predicate, press Finish.
* Following exception is logged in Error Log view:
```
java.lang.IndexOutOfBoundsException: Index 2 out of bounds for length 2
	at java.base/jdk.internal.util.Preconditions.outOfBounds(Preconditions.java:100)
	at java.base/jdk.internal.util.Preconditions.outOfBoundsCheckIndex(Preconditions.java:106)
	at java.base/jdk.internal.util.Preconditions.checkIndex(Preconditions.java:302)
	at java.base/java.util.Objects.checkIndex(Objects.java:365)
	at java.base/java.util.ArrayList.get(ArrayList.java:428)
	at org.eclipse.jdt.ui.actions.OrganizeImportsAction.lambda$1(OrganizeImportsAction.java:374)
        ...
```

jukzi:
Also happens in a Java 21 Project!

jukzi:
Can reproduced with any Java version if `List` is not unique, which can be forced by removing typefilter for com.sun:

![Image](https://github.com/user-attachments/assets/0e22c129-5ef9-4c50-a46f-71543f8ce07e)

![Image](https://github.com/user-attachments/assets/78c9393c-d2f4-46a3-b311-b656cc246282)

If on the other hand `List` is unique it will be imported without asking the user.

PR #1998, opened 2025-02-04
fix IndexOutOfBoundsException in OrganizeImportsAction.doChooseImports

jukzi:
When skipping an import except the last.

fixes https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1996


noopur2507 (approved):
Looks good. Thanks, Jörg!

Issue #2002, opened 2025-02-07
inplace type hierarchy throws exception in 2025-03


martinlippert:
I am using Eclipse JEE package 2025-03 M2. Created a simple plain Java 21 project, created one empty class.
Java editor is open, then invoking the inplace type hierarchy popup (Cmd-T on macOS). That works. Pressing CMd-T again to switch the mode doesn't work and throws tons of exception in the Error Log instead:

```
java.lang.NullPointerException: Cannot invoke "java.util.Set.add(Object)" because "this.visited" is null
	at org.eclipse.jdt.internal.ui.typehierarchy.HierarchyInformationControl$2.shouldExpand(HierarchyInformationControl.java:159)
	at org.eclipse.jdt.internal.ui.typehierarchy.HierarchyInformationControl$2.internalExpandToLevel(HierarchyInformationControl.java:148)
	at org.eclipse.jface.viewers.AbstractTreeViewer.expandToLevel(AbstractTreeViewer.java:1166)
	at org.eclipse.jface.viewers.AbstractTreeViewer.lambda$1(AbstractTreeViewer.java:1123)
	at org.eclipse.swt.custom.BusyIndicator.showWhile(BusyIndicator.java:67)
	at org.eclipse.jface.viewers.AbstractTreeViewer.expandToLevel(AbstractTreeViewer.java:1123)
	at org.eclipse.jface.viewers.AbstractTreeViewer.expandAll(AbstractTreeViewer.java:1084)
	at org.eclipse.jdt.internal.ui.typehierarchy.HierarchyInformationControl.toggleHierarchy(HierarchyInformationControl.java:364)
	at org.eclipse.jdt.internal.ui.typehierarchy.HierarchyInformationControl$1.keyPressed(HierarchyInformationControl.java:105)
	at org.eclipse.swt.widgets.TypedListener.handleEvent(TypedListener.java:205)
	at org.eclipse.swt.widgets.EventTable.sendEvent(EventTable.java:91)
	at org.eclipse.swt.widgets.Display.sendEvent(Display.java:4660)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1657)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1680)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1665)
	at org.eclipse.swt.widgets.Widget.sendKeyEvent(Widget.java:1694)
[...]

PR #2005, opened 2025-02-10
Fix NPE when opening quick type hierarchy twice #2002

fedejeanne:
Fixes https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2002

## Changes
- Introduce private method `runWithVisited`
- Override `expandToLevel(Object, int, boolean)` and let it use `runWithVisited`

### DIFF
--- a/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/typehierarchy/HierarchyInformationControl.java
+++ b/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/typehierarchy/HierarchyInformationControl.java
@@ -138,9 +138,23 @@ public class HierarchyInformationControl extends AbstractInformationControl {
 
 			@Override
 			protected void inputChanged(Object input, Object oldInput) {
-				visited= new HashSet<>();
-				super.inputChanged(input, oldInput);
-				visited= null;
+				runWithVisited(() -> super.inputChanged(input, oldInput));
+			}
+
+
+			@Override
+			public void expandToLevel(Object elementOrTreePath, int level, boolean disableRedraw) {
+				runWithVisited(
+						() -> super.expandToLevel(elementOrTreePath, level, disableRedraw));
+			}
+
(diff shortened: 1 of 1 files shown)


fedejeanne:
cc @jukzi

org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/typehierarchy/HierarchyInformationControl.java
```diff
@@ -138,9 +138,23 @@ protected TreeViewer createTreeViewer(Composite parent, int style) {
...
+
+			@Override
+			public void expandToLevel(Object elementOrTreePath, int level, boolean disableRedraw) {
+				runWithVisited(
+						() -> super.expandToLevel(elementOrTreePath, level, disableRedraw));
+			}
+
+			private void runWithVisited(Runnable r) {
+				try {
+					visited= new HashSet<>();
```
jukzi, line 153: i think this should check the requirement, that visited==null to prevent against wrong nested calls

jukzi (commented):
Maybe it would be less error prone to set and reset `visited` only in `internalExpandToLevel` if and only if `visited==null` when entered

fedejeanne:
> Maybe it would be less error prone to set and reset `visited` only in `internalExpandToLevel` if and only if `visited==null` when entered

Something like
```java
@Override
protected void internalExpandToLevel(Widget node, int level) {
	boolean reset = false;
	if (visited == null) {
		reset = true;
		visited = new HashSet<>();
	}
	if (!shouldExpand(node))
		return;

	super.internalExpandToLevel(node, level);
	if (reset) {
		visited = null;
	}
}
```

?

fedejeanne:
> Maybe it would be less error prone to set and reset `visited` only in `internalExpandToLevel` if and only if `visited==null` when entered

See #2006

fedejeanne:
replaced by #2006

PR #2014, opened 2025-02-12
Fix folding of 2 consecutive methods

jakub-suliga:
If the second method starts on the same line where the first one ends, there will still be two foldings. (I accidentally set includeLastLine to true a second time, which caused the issue).

Fixes https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1992

Testing:

1. Disabled new/experimental folding
2. Fold a, now b should be visible
```java
package org.example.test;

class X {
	/*
	 * a b
	 */
	void a() {

	} void b() {
	}
}
```

### DIFF
--- a/org.eclipse.jdt.ui/ui/org/eclipse/jdt/ui/text/folding/DefaultJavaFoldingStructureProvider.java
+++ b/org.eclipse.jdt.ui/ui/org/eclipse/jdt/ui/text/folding/DefaultJavaFoldingStructureProvider.java
@@ -1496,7 +1496,6 @@ public class DefaultJavaFoldingStructureProvider implements IJavaFoldingStructur
 				includelastLine= true;
 				IRegion normalized= alignRegion(regions[i], ctx);
 				if (normalized != null) {
-					includelastLine= true;
 					Position position= createCommentPosition(normalized);
 					if (position != null) {
 						boolean commentCollapse;


jakub-suliga:
> I tested the case described in #1992 and this PR makes no difference in the behavior. Can you add a test case to it and also explain how to test it manually?

What happens when you fold the method a, because when I am testing it, everything works.
image

I added how to test it and there is a test case for it, but currently it tests only when the new folding is activated. I am rewritting the test, so it tests both cases.

fedejeanne:
Thank you @jakub-suliga, I was able to reproduce it and I can confirm that this PR fixes the bug.

> I am rewritting the test, so it tests both cases. There will be soon a PR for it

Why not in this PR? If it the tests is improved so it can test this corner case then it belongs here :-)

jakub-suliga:
> Why not in this PR? If it the tests is improved so it can test this corner case then it belongs here :-)

I am currently working on refactoring the test class, not just the one test case. Therefore it should be in a separate PR.

fedejeanne:
> I am currently working on refactoring the test class, not just the one test case. Therefore it should be in a separate PR.

Finish the refactoring in another PR (merge it first if you want) and then add the test to this one. If you end up decoupling the fix from the test for that very same fix then commits can not be reverted without breaking tests, which is kind of confusing.

danthe1st:
Just a heads up:
Note that currently, the tests I added in #1825 use the current behavior from master which does not include this fix. Merging both PRs will make the build fail but fixing it should be very easy (so they should just not be merged concurrently).

jjohnstn:
Merged.  Thanks.

PR #2015, opened 2025-02-12
Fix folding of local classes

jakub-suliga:
Currently, the new folding feature does not support folding for local classes, whereas the old folding mechanism does. This fix ensures that local classes can be folded as expected. Additionally, I add a test case for the scenario.

Test:
1. Activate the new folding
2. Fold inner2, now it should be folded
```java
class Outer {
	void a() {
		class Inner2{
			
		}
	}
}
```


jakub-suliga:
@fedejeanne @iloveeclipse @jjohnstn could someone of you look over the code

fedejeanne (approved):
LGTM

fedejeanne:
Failing test is unrelated: https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1095

danthe1st:
Just a heads up:
Note that currently, the tests I added in #1825 use the current behavior from master which does not include this fix. Merging both PRs will make the build fail but fixing it should be very easy (so they should just not be merged concurrently).

PR #2019, opened 2025-02-14
Parameterize FoldingTests

fedejeanne:
Let them run both for the new folding and the old one. Some tests were skipped since the old folding doesn't fold as much as the new one.

Required by: #2020

### DIFF
--- a/org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/folding/FoldingTest.java
+++ b/org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/folding/FoldingTest.java
@@ -13,12 +13,18 @@
  *******************************************************************************/
 package org.eclipse.jdt.text.tests.folding;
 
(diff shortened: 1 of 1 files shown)


fedejeanne:
cc @jakub-suliga 

@jjohnstn would you kindly review/merge?

fedejeanne:
On a closer look,`testCommentBlocks` can run with the old folding too and `testAnonymousClassDeclarationFolding` **should** run with it too. 

`testAnonymousClassDeclarationFolding` currently fails so I see skipped it and added a `FIXME` to it in 92f0176acd06b39f0f359cf55a4b179c3e47e796. I created #2022 to take a look into it.

fedejeanne:
Could we merge this one for RC2? It's "just" tests and it is required by #2020 .

I did all the changes I wanted to do so it is complete on my side and ready to be reviewed and merged. @jjohnstn / @jukzi ?

jukzi:
It's ok to merge test changes.

Issue #1977, opened 2025-01-24
Stackoverflow in CoreASTProvider


robstryker:
This is likely because I had a breakpoint set somewhere. Even still, it would seem like there should be some checks on this recursive loop somewhere, and a better or more checked exception should somehow be thrown here. 

!ENTRY org.eclipse.core.jobs 4 2 2024-02-21 11:54:17.989
!MESSAGE An internal error occurred during: "Requesting Java AST from selection".
!STACK 0
java.lang.StackOverflowError
        at org.eclipse.jdt.core.manipulation.CoreASTProvider.getAST(CoreASTProvider.java:132)
        at org.eclipse.jdt.core.manipulation.CoreASTProvider.getAST(CoreASTProvider.java:189)
        at org.eclipse.jdt.core.manipulation.CoreASTProvider.getAST(CoreASTProvider.java:189)
        at org.eclipse.jdt.core.manipulation.CoreASTProvider.getAST(CoreASTProvider.java:189)
        at org.eclipse.jdt.core.manipulation.CoreASTProvider.getAST(CoreASTProvider.java:189)
        at org.eclipse.jdt.core.manipulation.CoreASTProvider.getAST(CoreASTProvider.java:189)
        at org.eclipse.jdt.core.manipulation.CoreASTProvider.getAST(CoreASTProvider.java:189)
        at org.eclipse.jdt.core.manipulation.CoreASTProvider.getAST(CoreASTProvider.java:189)
        at org.eclipse.jdt.core.manipulation.CoreASTProvider.getAST(CoreASTProvider.java:189)
        at org.eclipse.jdt.core.manipulation.CoreASTProvider.getAST(CoreASTProvider.java:189)
        at org.eclipse.jdt.core.manipulation.CoreASTProvider.getAST(CoreASTProvider.java:189)

jukzi:
That recursive call looks like the method arguments are not changed at all so the recursion works as a (waiting) loop and could be refactored to such.

jukzi:
relates to https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2028

PR #2032, opened 2025-02-19
CoreASTProvider: avoid Stackoverflow #1977

jukzi:
replaces tail recursion with infinite loop

https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1977


jukzi:
Note that this fix only solves the Stackoverflow when waiting too long. It does not try to solve deadlocks that have been reported in https://bugs.eclipse.org/bugs/show_bug.cgi?id=366048

Issue #2074, opened 2025-03-11
Java Editor is showing errors due to missing updates from the build sometimes


laeubi:
Here is a screenshot of a problem [initially described here](https://github.com/eclipse-pde/eclipse.pde/issues/1667#issue-2907561856) with a reproducer:

![Image](https://github.com/user-attachments/assets/b9d5d058-568d-4ae2-9b01-cff39b86fa19)

One should note the following things:

1. Neither Project Explorer nor the Problems View show any errors / warnings (error log is empty as well)
2. The Plugin Dependencies Classpath Container is fully setup and contains the dependencies
3. If one hovers over a "missing" class it shows the javadoc in the imports
4. **Typing** a single character in the editor make the problem go away, no need to save, build or alike see video
5. If one selects another editor (e.g. manifest editor) and restarts, then switch to the java editor everything works fine

https://github.com/user-attachments/assets/2103af75-b988-4056-9d67-037d315fbad7

```
Eclipse SDK

Version: 2025-03 (4.35)
Build id: I20250228-0140
```

merks:
The behavior seems somewhat unpredictable.  I can reproduce it with a debug launch:

![Image](https://github.com/user-attachments/assets/6ce40737-252d-41a9-835a-d4b315337b18)

But sometimes they are there for a while and then clean themselves up without touching a thing:

![Image](https://github.com/user-attachments/assets/f8f9292d-97fc-4e94-96d1-ed6ca5cfe485)

I added some print statements to PDE and see 
```
setupClasspath deferred org.eclipse.core.databinding
setupClasspath deferred org.eclipse.core.commands
setupClasspath from job org.eclipse.core.databinding
setupClasspath from job org.eclipse.core.commands
```

Note that all this happens before the initialize Java tools job:

![Image](https://github.com/user-attachments/assets/1a91af14-6468-44ad-8d1e-77bff859f4db)

That makes me wonder what the editor is able to do before the tools are initialized.

Mostly I see a build kick in after the initialization.  That's kind of what I'd expect from updating a project's classpath and that appears to trigger the editor to remove the markers.

If I make the deferred job wait really quite long (20s)  then the build happens later and the editor is updated later:
```
	public void initialize(IPath containerPath, IJavaProject javaProject) throws CoreException {
		if (Job.getJobManager().isSuspended()) {
			// if the jobmanager is currently suspended we can't use the
			// schedule/join pattern here, instead we must retry the requested
[...]

PR #2079, opened 2025-03-12
Force reconciling if a java project was touched

laeubi:
## What it does
If a classpath container changes, JDT "touches" the project resulting in a resource change being triggered. This is currently ignored by the JavaReconciler leading to possible wrong state displayed in an editor.

This now enhance the JavaReconciler to check for changed java projects if not already a reconciling was forced.

Fix https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2074

## How to test
see
- https://github.com/eclipse-pde/eclipse.pde/issues/1667#issue-2907561856

## Author checklist

### DIFF
--- a/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/text/JavaReconciler.java
+++ b/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/text/JavaReconciler.java
@@ -22,13 +22,16 @@ import org.eclipse.swt.widgets.Control;
 import org.eclipse.swt.widgets.Shell;
 
 import org.eclipse.core.runtime.Assert;
+import org.eclipse.core.runtime.CoreException;
 
 import org.eclipse.core.resources.IMarker;
 import org.eclipse.core.resources.IMarkerDelta;
+import org.eclipse.core.resources.IProject;
 import org.eclipse.core.resources.IResource;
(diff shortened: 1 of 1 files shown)


Phillipus:
Hi, I tested this PR using build I20250311-1800 and the simple test case from https://github.com/eclipse-pde/eclipse.pde/issues/1667#issue-2907561856.

Unfortunately, I see the same behaviour. The first launch of child Eclipse debug is OK, but thereafter the error markers remain. There are a couple of times the markers clear, but mainly on each launch they remain red.

laeubi:
@Phillipus just curious, can you try to create a second project?

If I look at `org.eclipse.jdt.core.JavaCore.setClasspathContainer(IPath, IJavaProject[], IClasspathContainer[], IProgressMonitor)` it has special handling for if there is only one project updated. and even though the API claims that

> In reaction to changing container values, the JavaModel will be updated to reflect the new state of the updated container. A combined Java element delta will be notified to describe the corresponding classpath changes resulting from the container update. 

I can't see that this happens it seems it only updates some internal maps...

laeubi:
Closing as it does not seem the right way.

Phillipus:
> @Phillipus just curious, can you try to create a second project?

Yep, that makes a difference. I created a second project just the same as the first and the red markers do not appear, even momentarily.

PR #2084, opened 2025-03-14
Added Switch Statment and Switch Expression to the new folding

jakub-suliga:
I added support for folding switch statements and switch expressions. Now they can be collapsed. In addition I added test caes for them.

Old switch stament:
image
Old switch expression
image
New switch statment:
image
New switch expression
image

## How to test
1. Add both codesampels in a workspace 
2. Now there should be folding buttons

For switch expression:
```java
int b = 0;
	int c = switch (b) {
		case 1 -> 1;			
		default -> 0;			
	};
```

For switch statment:
```java
switch (zaehler) {
        case 0:
            break;
        case 1:
            break;
        default:
            break;
    }
```

### DIFF
--- a/org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/folding/FoldingTest.java
+++ b/org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/folding/FoldingTest.java
@@ -575,13 +575,94 @@ public class FoldingTest {
 				        try {						//here should not be an annotation
 				        } catch (Exception e) {		//here should not be an annotation
 				        }
+				        int zaehler = 0;
+						switch (zaehler) {			//here should be an annotation
+				        case 0:						//here should not be an annotation
+				            break;
+				        default:					//here should not be an annotation
+				            break;
+				    	}
 				    }
 				    public void bar() {					//here should not be an annotation
 				    }
 				}
 				""";
-		FoldingTestUtils.assertCodeHasRegions(packageFragment, "TestFolding.java", str, 1);
+		FoldingTestUtils.assertCodeHasRegions(packageFragment, "TestFolding.java", str, 2);
 		List<IRegion> regions= FoldingTestUtils.getProjectionRangesOfFile(packageFragment, "TestFolding.java", str);
-		FoldingTestUtils.assertContainsRegionUsingStartAndEndLine(regions, str, 2, 13); // Method
+		FoldingTestUtils.assertContainsRegionUsingStartAndEndLine(regions, str, 2, 20); // Method
+		FoldingTestUtils.assertContainsRegionUsingStartAndEndLine(regions, str, 15, 19); // switch
+	}
+
+	@Test
+	public void testSwitchExpression() throws Exception {
+		assumeTrue("Only doable with the new folding", newFoldingActive);
+		String str= """
+				package org.example.test;
+				class Outer {
+					void a() {						//here should be an annotation
+						int b = 0;
+						int c = switch (b) {		//here should be an annotation
+							case 1 -> 				//here should be an annotation
+
+							1;
+							case 2 -> 				//here should not be an annotation
+							1;
+							case 3 -> 				//here should be an annotation
+
+							break;
+							1;
+							case 4 -> {				//here should be an annotation
+								b = 2;
+								yield 3;
+							}
+							case 5 -> {				//here should not be an annotation
+								yield 3;
+							}
+							default -> 				//here should be an annotation
+
+							0;
+						};
+					}
+				}
+				""";
+		FoldingTestUtils.assertCodeHasRegions(packageFragment, "TestFolding.java", str, 6);
+		List<IRegion> regions= FoldingTestUtils.getProjectionRangesOfFile(packageFragment, "TestFolding.java", str);
+		FoldingTestUtils.assertContainsRegionUsingStartAndEndLine(regions, str, 2, 24); // method
+		FoldingTestUtils.assertContainsRegionUsingStartAndEndLine(regions, str, 4, 23); // switch
+		FoldingTestUtils.assertContainsRegionUsingStartAndEndLine(regions, str, 5, 6); // 1. case
+		FoldingTestUtils.assertContainsRegionUsingStartAndEndLine(regions, str, 10, 11); // 3. case
+		FoldingTestUtils.assertContainsRegionUsingStartAndEndLine(regions, str, 14, 15); // 4. case
+		FoldingTestUtils.assertContainsRegionUsingStartAndEndLine(regions, str, 21, 22); // default
+	}
+
+	@Test
+	public void testSwitchStatment() throws Exception {
+		assumeTrue("Only doable with the new folding", newFoldingActive);
+		String str= """
+				package org.example.test;
+				class Outer {
+					void a() {						//here should be an annotation
+						int b = 0;
+						switch (b) {				//here should be an annotation
+					        case 0:					//here should be an annotation
+
+					            break;
+					            b = 1;
+					        case 1:					//here should not be an annotation
+					            break;
+
+					        default:				//here should be an annotation
+
+					          	break;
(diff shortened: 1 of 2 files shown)


jakub-suliga:
@fedejeanne @iloveeclipse @jjohnstn Could you review the code when you have a moment? Thanks in advance!

iloveeclipse:
From the usability point of view, does it make sense to provide folding toggles for single line switch cases?
They just add visual noise and clicking on them doesn't do anything useful:
image

org.eclipse.jdt.ui/ui/org/eclipse/jdt/ui/text/folding/DefaultJavaFoldingStructureProvider.java
```diff
@@ -488,6 +491,44 @@ public boolean visit(Initializer node) {
...
+		public boolean visit(SwitchStatement node) {
+			createFoldingRegionForStatement(node);
+			foldingSwitch(node, node.statements());
+			return false;
+		}
+
+		@Override
+		public boolean visit(SwitchExpression node) {
+			createFoldingRegionForStatement(node);
+			foldingSwitch(node, node.statements());
```
fedejeanne, line 504:
```suggestion
			createFoldingRegionForSwitch(node);
```

org.eclipse.jdt.ui/ui/org/eclipse/jdt/ui/text/folding/DefaultJavaFoldingStructureProvider.java
```diff
@@ -488,6 +491,44 @@ public boolean visit(Initializer node) {
...
+			return false;
+		}
+
+		private void foldingSwitch(ASTNode node, List<?> statements) {
```
fedejeanne, line 508: Isn't it possible to do this with only 1 loop instead of nesting 2?

fedejeanne (commented):
I did a quick evaluation and here are my findings:

1. I agree with @iloveeclipse : one-liners shouldn't be foldable. 

2. It would be worth always showing  the special keywords `break` and `yield` (even when folding) since hiding them would let the reader without any information as to whether or not the folded case _falls through_ the next case, which makes understanding the logic more complicated. Would that be doable?

I can take a deeper look next week and see if I find something else.

org.eclipse.jdt.ui/ui/org/eclipse/jdt/ui/text/folding/DefaultJavaFoldingStructureProvider.java
jakub-suliga: Why do you suggest removing the node.statements() parameter? In my foldingSwitch method, I take ASTNode node and List<?> statements as parameters. I can't retrieve the statements directly from the ASTNode, nor can I cast it, since it could be either a SwitchExpression or a SwitchStatement. Could you explain it or could you give me a suggestion how i could change it. Thank you.

fedejeanne: You're right, `node.statements()` can't be called inside the method because `node` is an `ASTNode`. Leave it 👍

jjohnstn:
@jakub-suliga Any response to the comments above?  I'd like to try and review this for 4.36 M1.

jakub-suliga:
"Sorry, I just realized I forgot to push my latest changes. I'm currently on vacation and won't be able to push them until I'm back in two weeks."

jjohnstn:
No problem @jakub-suliga.  We'll look at getting it into 4.36 M2.

org.eclipse.jdt.ui/ui/org/eclipse/jdt/ui/text/folding/DefaultJavaFoldingStructureProvider.java

jakub-suliga:
>  one-liners shouldn't be foldable
I will create another PR for this to ensure that all single-line statements (e.g., if, etc.) are not folded. It will be published within the next few hours.
@jjohnstn You can review it. Thanks!

org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/folding/FoldingTest.java
```diff
@@ -547,4 +547,52 @@ class Inner2{				//here should be an annotation
...
+	@Test
+	public void testSwitchExpression() throws Exception {
+		assumeTrue("Only doable with the new folding", newFoldingActive);
+		String str= """
+				package org.example.test;
+				class Outer {
+					void a() {						//here should be an annotation
+						int b = 0;
+						int c = switch (b) {		//here should be an annotation
+							case 1 -> 1;			//here should be an annotation
```
danthe1st, line 560: These tests will conflict with #2139, right? If so, I think it should be made sure that these PR aren't merged together.

iloveeclipse:
> Done in #2139

Thanks, merged. Please resolve merge issues in this PR now.

jakub-suliga:
@iloveeclipse Can you check the code? Thanks!

fedejeanne (changes requested):
`yield` is being folded and it shouldn't

image

fold...
image

jakub-suliga:
The issue should now be resolved. I’ve added a test case too. @fedejeanne, could you please verify?

fedejeanne (approved):
LGTM ✔️ 

Now we only need the approval of a committer 👍

org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/folding/FoldingTest.java
fedejeanne: This PR and its tests have been adapted to #2139 ✔️ 
@jakub-suliga / @danthe1st you may mark this comment as resolved

fedejeanne:
> From the usability point of view, does it make sense to provide folding toggles for single line switch cases? They just add visual noise and clicking on them doesn't do anything useful: image

@iloveeclipse single-line blocks **inside** the switch have been left alone but the switch itself can be folded even if it "does little/nothing". That's how you wanted it, right?

Current status:

```java
public class ExampleFoldingOneLiners {
		int i = (int) (Math.random() * 3);
		switch (i) { // folding: the switch itself "does nothing" but it contains many lines
			case 0: // no folding: 0 lines
			case 1:
				break; // no folding: 0 lines
			default: { // no folding: 1 line but contains special word (break/yield)
				break;
			}
		}
	}

	private int a() {
		int i = (int) (Math.random() * 3);
		return switch (i) { // folding: the switch itself "does nothing" but it contains many lines
			case 0 -> 1;// no folding: 0 lines
			case 1,2 -> { // no folding: 1 line but contains special word (break/yield)
				yield 2;
			}
			// no folding: 1 line but it's an implicit yield
			default -> 
				0;
			};
	}
}
```

And it looks like this:

image

Which is exactly how I imagined it 👍

fedejeanne:
@iloveeclipse / @jjohnstn would you say that the new folding feature can be declared "production ready" once this PR is merged? 

My vote: **yes**.

If you agree, @jakub-suliga could start working on a rework of the configuration dialog and activating the feature by default:

image

iloveeclipse:
> single-line blocks **inside** the switch have been left alone but the switch itself can be folded even if it "does little/nothing". That's how you wanted it, right?

Correct. I will most likely not review this PR later, today no time.

iloveeclipse:
> would you say that the new folding feature can be declared "production ready" once this PR is merged?

Yes, why not? it is useful, I hope it shouldn't have side effects, so ideally you the PR making this default will be merged before M2.

fedejeanne:
> > would you say that the new folding feature can be declared "production ready" once this PR is merged?
> 
> Yes, why not? it is useful, I hope it shouldn't have side effects, so ideally you the PR making this default will be merged before M2.

@jakub-suliga that's your cue :-)

jjohnstn (approved):
Looks good.  Works well.

PR #2086, opened 2025-03-14
Spare control characters from content assist trigger

ptziegler:
## What it does

On some versions of MacOS, the Ctrl+S hotkey is realized as a key event with 'S' as character and 'Ctrl' as state mask, rather than the 'XOFF' character (0x13, also with the 'Ctrl' state mask).

If the auto-activation trigger for Java include the character 's', this then causes the completion box to show up when the user tries to save the editor while it's already in a clean state.

This was previously fixed on the platform [1], but needs to be reverted as generic source viewers might also check for the 'Ctrl' key [2]. In order to avoid a regression, the check therefore needs to be moved to the JDT component.

[1] https://github.com/eclipse-platform/eclipse.platform.ui/pull/1556
[2] https://github.com/eclipse-platform/eclipse.platform.ui/issues/2723

## How to test

In the Java Content Assist preference page, update the auto activation triggers for Java to include the character 'S' (or '' on most other platforms). In a clean Java editor, press Ctrl+S. On older versions of the Eclipse IDE (likely pre 2022-03), this would cause the code completion box to show up.

image

## Author checklist

### DIFF
--- a/org.eclipse.jdt.ui/META-INF/MANIFEST.MF
+++ b/org.eclipse.jdt.ui/META-INF/MANIFEST.MF
@@ -105,7 +105,7 @@ Require-Bundle:
  org.eclipse.compare;bundle-version="[3.5.0,4.0.0)",
  org.eclipse.team.ui;bundle-version="[3.4.100,4.0.0)",
  org.eclipse.team.core;bundle-version="[3.4.100,4.0.0)",
- org.eclipse.jface.text;bundle-version="[3.20.0,4.0.0)",
+ org.eclipse.jface.text;bundle-version="[3.28.0,4.0.0)",
  org.eclipse.ui;bundle-version="[3.117.0,4.0.0)",
  org.eclipse.ui.console;bundle-version="[3.4.0,4.0.0)",
  org.eclipse.ui.workbench;bundle-version="[3.131.0,4.0.0)",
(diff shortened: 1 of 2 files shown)


jjohnstn:
@noopur2507 Can you test this out and verify it works as expected on MacOS?

ptziegler:
> I have macOS 15.4.1 and Eclipse IDE I20250413-1800. I added 's' in auto-activation triggers for Java. I don't see any difference on pressing Ctrl+s with or without this PR. Can you please specify the before and after behaviour clearly to test?

You won't see this on any recent version because of https://github.com/eclipse-platform/eclipse.platform.ui/pull/1556. The whole point of this PR is to move the fix from Platform to JDT so that the Platform change can be reverted to fix https://github.com/eclipse-platform/eclipse.platform.ui/issues/2723.

In order to reproduce the original behavior, you need to test this again an older Eclipse version or a local version where that commit has been reverted. The steps to reproduce are described in https://github.com/eclipse-platform/eclipse.platform.ui/issues/891:

> 1. Mac M1 chip version
> 2. Eclipse 2022-3 version
> 3. Set auto activation triggers for java:.abcdefghijklmnopqrstuvwxyz
> 4. At this time, press the shortcut key (ctrl+s) multiple times and a prompt box will pop up

noopur2507:
I verified it by reverting the platform fix locally. I can see the PR works as expected for me with the use of Cmd+s (not, Ctrl+s) i.e. with the platform fix reverted and adding 's' to the auto activation triggers, the content assist pop-up appears without the PR on pressing Cmd+s in a non-dirty editor. However, with the PR, the content assist pop-up no longer appears.

noopur2507:
The PR works as expected above on Mac OS.

Issue #2119, opened 2025-03-28
No Import suggestion if line contains multiple missing types


ptziegler:
Quite frequently I create a variable whose type is an interface but which is then assigned with an actual implementation. For example the Eclipse IPath.
If neither classes have been imported yet, the assistant doesn't propose the required imports:

![Image](https://github.com/user-attachments/assets/90573de4-f5e7-45d0-81fe-766f57b50db4)

If I change both sides to use `Path`, I am able to resolve one import.

![Image](https://github.com/user-attachments/assets/ccc56535-c07f-43d1-8298-cc7c2915ccea)

But I'm still unable to import `IPath`, even though it's in the same package as `Path`.

![Image](https://github.com/user-attachments/assets/cc466444-b9b6-4523-8478-ae4354a2a7de)

Which can then only be fixed by manually adding the missing import.

![Image](https://github.com/user-attachments/assets/7feefc51-3737-4a8a-9e1b-4c66c94531a9)

I'm currently using the I20250326-1800, though I have seen this issue pop up in several releases. I can't say which version this started with.

ptziegler:
As a bonus, the import `java.nio.file.Path` is only suggested for the left `Path`

![Image](https://github.com/user-attachments/assets/0dd819af-66aa-494e-a27e-751734ca907a)

ptziegler:
For comparision the 2022-03 release, where the import was still working. I also think the suggested changes are **significantly** more appropriate, rather than pointing to some clearly JDK-internal types...

![Image](https://github.com/user-attachments/assets/2c3107a3-5423-4593-834e-8485df464a8c)

ptziegler:
The regression occurred with the 4.27. If someone can give me a hint as to where I have to look, I might be able to pinpoint the exact cause.

iloveeclipse:
> The regression occurred with the 4.27. If someone can give me a hint as to where I have to look

Maybe 546a6978432fa5f2b01083919364848dd89805dc?

@jjohnstn, could you please check?

ptziegler:
> Maybe [546a697](https://github.com/eclipse-jdt/eclipse.jdt.ui/commit/546a6978432fa5f2b01083919364848dd89805dc)?
> 
> [@jjohnstn](https://github.com/jjohnstn), could you please check?

That's exactly the one! I've checked out the 4.27 locally and reverted the revision and now the type is found again.

ptziegler:
[...]

PR #2120, opened 2025-03-28
Skip import quickfix hierarchy check if base type is unresolved

ptziegler:
## What it does

The commit 546a6978432fa5f2b01083919364848dd89805dc extended the import check to compare the interfaces of the suggested types with the interface of the missing type.

In case the base type is unresolved, this then skips all suggestions, even though it should let all pass so that the user can decide which one to pick.

Fixes https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2119

## How to test

Use the quickfix to add the imports for the Eclipse `IPath` and `Path` in the following snippet.

```
package test;

class Test {
	private IPath p = new Path("...");
}
```

## Author checklist


jjohnstn (approved):
Looks good.  Thanks.

PR #2127, opened 2025-04-02
Ensure forceReconciling is called after initialProcess if it was ignored

merks:
## What it does

Fixes https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2074

## How to test

See the issue and this PR:

https://github.com/eclipse-pde/eclipse.pde/pull/1711

Basically in a self-hosted launch, create some plugin projects,  leave an editor on one source file open, and restart.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/text/JavaReconciler.java
+++ b/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/text/JavaReconciler.java
@@ -22,6 +22,8 @@ import org.eclipse.swt.widgets.Control;
 import org.eclipse.swt.widgets.Shell;
 
 import org.eclipse.core.runtime.Assert;
+import org.eclipse.core.runtime.ICoreRunnable;
+import org.eclipse.core.runtime.jobs.Job;
 
 import org.eclipse.core.resources.IMarker;
 import org.eclipse.core.resources.IMarkerDelta;
@@ -281,7 +283,9 @@ public class JavaReconciler extends MonoReconciler {
 	 */
 	private IPropertyChangeListener fPropertyChangeListener;
 
-	private boolean fIninitalProcessDone= false;
+	private boolean fIninitalProcessDone;
+
+	private boolean fDeferForceReconcile;
 
 	/**
 	 * The element that this reconciler reconciles.
@@ -382,7 +386,7 @@ public class JavaReconciler extends MonoReconciler {
 	 */
 	@Override
 	protected void forceReconciling() {
-		if (!fIninitalProcessDone)
+		if (!isInitialProcessDone())
 			return;
 
 		super.forceReconciling();
@@ -420,8 +424,43 @@ public class JavaReconciler extends MonoReconciler {
 		synchronized (fMutex) {
 			super.initialProcess();
 		}
-		fIninitalProcessDone= true;
+		if (initialProcessDone()) {
+			Job.createSystem("Reconciler init", (ICoreRunnable) monitor -> forceReconciling()).schedule();  //$NON-NLS-1$
+		}
 	}
+
+	/**
+	 * This is called by {@link #initialProcess()} to indicate that it is has finished and returns
+	 * true if a call to {@link #forceReconciling()} is necessary because it was previously ignored.
+	 *
+	 * @return whether there has been a call to {@link #forceReconciling()} that was ignored because
+	 *         {@link #fIninitalProcessDone} was false.
+	 */
+	private synchronized boolean initialProcessDone() {
+		fIninitalProcessDone= true;
+		if (fDeferForceReconcile) {
+			fDeferForceReconcile= false;
+			return true;
+		}
+		return false;
+	}
+
+	/**
+	 * This is called by {@link #forceReconciling()} to determine whether
+	 * {@link #initialProcessDone()} has completed. It sets {@link #fDeferForceReconcile} to true if
+	 * {@link #fIninitalProcessDone} is false so that {@link #initialProcessDone()} will return
+	 * true and will subsequently call {@link #forceReconciling()} again.
+	 *
+	 * @return whether {@link #initialProcessDone()} has completed.
+	 */
+	private synchronized boolean isInitialProcessDone() {
+		if (!fIninitalProcessDone) {
+			fDeferForceReconcile= true;
+			return false;
+		}
+		return true;
+	}
+
 
 	/**
 	 * Tells whether the Java Model has changed or not.


merks:
@iloveeclipse 

Is there any chance of getting this (with whatever changes you request) in for M1?  Every IDE I open starts with errors and it would be great after M1 to see that problem go away...

org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/text/JavaReconciler.java
```diff
@@ -420,9 +422,44 @@ protected void initialProcess() {
 		synchronized (fMutex) {
 			super.initialProcess();
 		}
+		if (initialProcessDone()) {
+			forceReconciling();
```
iloveeclipse, line 426: I'm not familiar with this code, but iI see that `forceReconsiling()` is called here from `AbstractReconciler.BackgroundThread` and I see that in some other places before calling `forceReconsiling()` the code checks that this is not the case via `if (isRunningInReconcilerThread()) return;`

So I simply not sure

merks: I get the sense that there are vanishingly few people familiar with the code.  😢  

Also guards like this exist, probably to avoid work if the user can't see the editor:

image

So I think we cannot be sure of anything except that there is a timing problem here that needs attention and there is a bug that is in need of attention but not getting any because it's not a priority for anyone.  😞

jjohnstn: What about trying a delay and recheck in the forceReconciling() method?  That would avoid the potential issue of using the background thread.

merks: I see no "good" mechanism whereby one might do a delayed thing.   Spawn yet another job?  (I wonder why the initialization itself is not done on a job?)

Note too that there are many callers:

image

The call stack leading to the problem of an ignored  call is this one:
```
forceReconciling  fIninitalProcessDone=false
java.lang.Exception: Stack trace
	at java.base/java.lang.Thread.dumpStack(Thread.java:2210)
	at org.eclipse.jdt.internal.ui.text.JavaReconciler.forceReconciling(JavaReconciler.java:389)
	at org.eclipse.jdt.internal.ui.text.JavaReconciler$ElementChangedListener.elementChanged(JavaReconciler.java:172)
	at org.eclipse.jdt.internal.core.DeltaProcessor$3.run(DeltaProcessor.java:1751)
	at org.eclipse.core.runtime.SafeRunner.run(SafeRunner.java:47)
	at org.eclipse.jdt.internal.core.DeltaProcessor.notifyListeners(DeltaProcessor.java:1739)
	at org.eclipse.jdt.internal.core.DeltaProcessor.firePostChangeDelta(DeltaProcessor.java:1572)
	at org.eclipse.jdt.internal.core.DeltaProcessor.fire(DeltaProcessor.java:1548)
	at org.eclipse.jdt.internal.core.DeltaProcessor.notifyAndFire(DeltaProcessor.java:2261)
	at org.eclipse.jdt.internal.core.DeltaProcessor.resourceChanged(DeltaProcessor.java:2153)
	at org.eclipse.jdt.internal.core.DeltaProcessingState.resourceChanged(DeltaProcessingState.java:490)
	at org.eclipse.core.internal.events.NotificationManager$1.run(NotificationManager.java:321)
	at org.eclipse.core.runtime.SafeRunner.run(SafeRunner.java:47)
	at org.eclipse.core.internal.events.NotificationManager.notify(NotificationManager.java:311)
	at org.eclipse.core.internal.events.NotificationManager.broadcastChanges(NotificationManager.java:174)
	at org.eclipse.core.internal.resources.Workspace.broadcastPostChange(Workspace.java:465)
	at org.eclipse.core.internal.resources.Workspace.endOperation(Workspace.java:1593)
	at org.eclipse.core.internal.resources.Workspace.run(Workspace.java:2471)
	at org.eclipse.core.internal.resources.Workspace.run(Workspace.java:2482)
	at org.eclipse.jdt.internal.core.JavaModelOperation.runOperation(JavaModelOperation.java:821)
	at org.eclipse.jdt.core.JavaCore.setClasspathContainer(JavaCore.java:6243)
	at org.eclipse.pde.internal.core.RequiredPluginsInitializer.setupClasspath(RequiredPluginsInitializer.java:122)
	at org.eclipse.pde.internal.core.RequiredPluginsInitializer$DeferredClasspathContainerInitializerJob.run(RequiredPluginsInitializer.java:73)
	at org.eclipse.core.internal.jobs.Worker.run(Worker.java:63)
initialProcess elapsedTime=3729
java.lang.Exception: Stack trace
	at java.base/java.lang.Thread.dumpStack(Thread.java:2210)
	at org.eclipse.jdt.internal.ui.text.JavaReconciler.initialProcess(JavaReconciler.java:433)
	at org.eclipse.jface.text.reconciler.AbstractReconciler$BackgroundThread.run(AbstractReconciler.java:177)
```

Trying to create yet another job to do a delayed `forceReconciling()` after an unknown amount of time would seem to me much more risky and definitely significantly more complicated than this simple localized change.  It's also not clear why a job, which is also "just another thread with lots of overhead" is less risky than using the thread we have.

In any case, we are about to produce M1 tomorrow so we will have many weeks to notice if this causes a regression in the real world; these are no test regressions for sure.   Consider too that the problem being solved happens in pretty much all of my 2025-03 IDEs so I must assume it is a pervasive user problem which is why I've spent several days trying to stamp this out...

And finally, note that when I simply open a new editor in the IDE, `initialProcess()` is called but `forceReconciling()` is not called at all:

image

So this is literally only an issue (and a change of behavior) for events that happen while `initialProcess()` is running which is an unusual situation that I believe only happens when PDE is doing deferred classpath container initialization during starting affecting only the one initially-open editor.  The risk appears to be negligible...

iloveeclipse: > It's also not clear why a job, which is also "just another thread with lots of overhead" is less risky than using the thread we have.

The reconsiler code is locked in many places by synchronized, the risk here is that we run into a deadlock which would be much worse as "just" an initial UI glitch here.

Ed, can you share easy way to reproduce the problem?
I think the easiest and safest way is to use a system job like here:

```
diff --git a/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/text/JavaReconciler.java b/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/text/JavaReconciler.java
index cac360a..49ed878 100644
--- a/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/text/JavaReconciler.java
+++ b/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/text/JavaReconciler.java
@@ -24,2 +24,4 @@
 import org.eclipse.core.runtime.Assert;
+import org.eclipse.core.runtime.ICoreRunnable;
+import org.eclipse.core.runtime.jobs.Job;
 
@@ -425,3 +427,3 @@
 		if (initialProcessDone()) {
-			forceReconciling();
+			Job.createSystem("Reconciler init", (ICoreRunnable) monitor -> forceReconciling()).schedule();  //$NON-NLS-1$
 		}

```

iloveeclipse (approved):
Thanks.

org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/text/JavaReconciler.java
merks: Thanks, that a really constructive simple suggestion which I've force pushed.

---

As I mentioned initially, all I do is in a self-hosted launch, create some plugin projects, leave an editor on one source file open, and restart.  Here we are stopped at the place that creates the system job:

image

After it runs, the errors are cleaned up:

image

Note that I have  this PR in my SDK workspace though without a delay:

https://github.com/eclipse-pde/eclipse.pde/pull/1711

merks:
@iloveeclipse 

It looks like all the tests pass and https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2130 is still running.  Do you still want a rebase when that's done?

iloveeclipse:
> @iloveeclipse
> 
> It looks like all the tests pass and #2130 is still running. Do you still want a rebase when that's done?

No. It seem that tests on jenkins don't include generic editor bundle.

iloveeclipse:
Thanks Ed, really appreciate that you've debugged all that mess with startup init.

merks:
And I really appreciate your constructive suggestions, the review from @jjohnstn, and that this is merged for M1.  Yeah!!!  Thanks thanks!!

Phillipus:
Doe this mean no more red errors at startup? Will test if so... :-)

merks:
I'm not 100% sure if https://github.com/eclipse-pde/eclipse.pde/pull/1711 is also needed to fix the problem but I think so...

Phillipus:
I tested build I20250404-1800. I still see red error markers in Java editors on startup. They go away after a few seconds, but Eclipse never used to be like this. Is this intended behaviour?

merks:
I’m not sure the value in repeating the months long history. Maybe best you dredge through it. 

Quick summary. JDT asks for the classpath on the UI thread really early even when the job manager is not yet enabled to schedule jobs.  PDE now guards against this deferring the computation to a background job. The computation typically takes several seconds if everything is cached. It could take minutes.  There were issues filed complaining about extremely long freezes during startup.  So now the startup does freeze but the editor reports errors during the former freeze period. 

Is it intended that a long running task be done on the UI thread early during startup?  Apparently so because no steps are being taken to avoid that. Of course I think it’s a hard problem so I’m not pointing fingers. I can only say that I’ve invested days of effort making it less horrible. Someone else will need to nip this problem at the source. No classpath initialization on the UI thread. That needs to stop and of course avoiding expensive IO and network traffic on the UI thread is always to be avoided at all cost.  But I think the folks with the skills and insights are busy on other things.

Phillipus:
@merks Thanks for the explanation, I appreciate it. :smile:

I guess I'm still trying to further my understanding of why it all *seemed* to be working before [this](https://github.com/eclipse-pde/eclipse.pde/commit/7e45ebbb7a0efa7bd64d0704df986aa1006f33b6). 

> I can only say that I’ve invested days of effort making it less horrible.

I can see that and I appreciate your hard work (and all the the work that you've put into Eclipse over the years 💯  )

merks:
The code wasn’t just changed for no good reason.   There were issues opened about minute(s) long startup freezes with fingers pointing at PDE.  But PDE cannot make the world spin faster. 

As I said, it’s best to focus on fixing the problem at the origin rather than suggesting that PDE should start yet again resolving a target platform on the the UI thread because that mostly seemed OK except when it wasn’t.   We should move forward. 

It reminds me a bit of the horror of URI encoding where properly encoding a URI that was not being properly encoded previously breaks downstream code which assumed the URL was not encoded.  The temptation is to revert the “breaking” change but it’s a death spiral where the problems just cannot and will not ever be fixed.  This also has cost me more than a day of effort but was well invested nevertheless. 

One really significant thing to note here is that this behavior literally on kicks in if and only if the job manager is suspended. That only is the case during the very earliest stages of bootstrapping the workbench. It’s even been suggested that we stop suspending the job manager to fix the problem.  It seems any solution that doesn’t involve JDT avoiding to do arbitrarily expensive work at what is arguably an obscenely early point during startup is preferable.  We really need to move toward a proper solution. I wish I were made of time. 😩

Phillipus:
@merks Thank-you for taking the time to explain things further, I appreciate it. :-)

merks:
I know who the people are who make Eclipse the great community that I love. 💕  You are high on that list! 🏅

Phillipus:
> I know who the people are who make Eclipse the great community that I love. 💕 You are high on that list! 🏅

Ah, man, that's high praise coming from you! (I've been developing on Eclipse since 2003. BTW - the greatest Eclipse framework **ever** is EMF).

PR #2139, opened 2025-04-07
Changed that single liner in new folding dont get folding button

jakub-suliga:
I’ve modified the behavior so that single-line statements no longer display a folding button. Since there’s nothing to collapse, the button serves no functional purpose and can be visually distracting. For example, if a single method is declared and the next line is just a bracket, clicking the folding button makes no difference. This change is also important for PR #2084.

For example:
Old:
image

New:
image

New with more than 1 line:
image

How to Test:
Use for example this and pasted it in the editor and there shouldnt be a folding button.
```java
if (true) {
	}
```


iloveeclipse:
I've just rebased on master

org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/folding/FoldingTest.java
danthe1st, line 1: I think your changes to the test only ensure that there is a line ensuring the folding regions are present. I think it might be a good idea to add a test that actually tests your PR (ensuring that a consisting of only a single line isn't folded) i.e. actually test your improvement.

danthe1st (changes requested):

org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/folding/FoldingTest.java
jakub-suliga: Done

jakub-suliga:
@jjohnstn @iloveeclipse @fedejeanne Can you look over the code?

This PR needs to be merged befor #2084

iloveeclipse (approved):
LGTM

PR #2158, opened 2025-04-14
[#2157] Describe final field initialization proposal considering type

### LINKED ISSUE #2157: Сorrection proposals in same context have same name but different results
## Summary
When final field is declared, but not yet initialized and the explicit constructor defined, there are several correction proposals suggested:
Image

Both of them have same name, but do behave a bit different. The first one creates "empty" initialization in constructor, while the second one initializes field with constructor parameter, that's why such naming can be very confusing.

## Mitigations
1. Wait for preview to render to see the correction result and then you can distinguish corrections from each other.
2. Corrections often have the same order, so after some time it's generally OK for user to distinguish them by order, but it is still not very discoverable.

## Proposal
Add `with parameter` to the second proposal description to make them different and self-descriptive.

zelenyhleb:
## What it does
`InitializeFinalFieldProposal` now selects more appropriate description using updateType passed to it, which helps better distinguish both types of corrections in UI.

image

## How to test
- create new class
- create final field inside it
- create empty constructor
- invoke correction proposals with `Ctrl+Space` or other configured shortcut
- see the different names

Fixes #2157

### DIFF
--- a/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/text/correction/proposals/InitializeFinalFieldProposal.java
+++ b/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/text/correction/proposals/InitializeFinalFieldProposal.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2019, 2023 IBM Corporation and others.
+ * Copyright (c) 2019, 2025 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -11,6 +11,7 @@
  * Contributors:
  *     IBM Corporation - initial API and implementation
  *     Red Hat Inc - separate core logic from UI images
+ *     Nikifor Fedorov (ArSysOp) - eclipse-jdt/eclipse.jdt.ui#2157 Select proposal description considering type
  *******************************************************************************/
 
 package org.eclipse.jdt.internal.ui.text.correction.proposals;
@@ -23,9 +24,10 @@ import org.eclipse.jdt.core.dom.IVariableBinding;
 
 import org.eclipse.jdt.internal.corext.util.Messages;
 
+import org.eclipse.jdt.ui.text.java.IProblemLocation;
+
 import org.eclipse.jdt.internal.ui.JavaPluginImages;
 import org.eclipse.jdt.internal.ui.text.correction.CorrectionMessages;
-import org.eclipse.jdt.ui.text.java.IProblemLocation;
 
 public class InitializeFinalFieldProposal extends LinkedCorrectionProposal {
 	public InitializeFinalFieldProposal(IProblemLocation problem, ICompilationUnit cu, ASTNode astNode, IVariableBinding variableBinding, int relevance) {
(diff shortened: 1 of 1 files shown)


org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/text/correction/proposals/InitializeFinalFieldAtConstructorDescription.java
```diff
@@ -0,0 +1,37 @@
+/*******************************************************************************
```
jjohnstn, line 1: Hi @zelenyhleb Thanks for contributing.  A new class is not warranted for this.  Just add a conditional using the updateType in the super constructor call in InitializeFinalFieldProposal.

jjohnstn (changes requested):

org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/text/correction/proposals/InitializeFinalFieldAtConstructorDescription.java
zelenyhleb: Hi @jjohnstn. Thank you for your review! I did this initially to improve readability a bit, because with conditionals constructor looked overloaded for me. Fixed this.

org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/text/correction/proposals/InitializeFinalFieldProposal.java
```diff
@@ -34,11 +36,16 @@ public InitializeFinalFieldProposal(IProblemLocation problem, ICompilationUnit c
 	}
 
 	public InitializeFinalFieldProposal(IProblemLocation problem, ICompilationUnit cu, ASTNode astNode, int relevance, int updateType) {
-		super(Messages.format(CorrectionMessages.InitializeFieldInConstructorCorrectionProposal_description, problem.getProblemArguments()[0]), cu, null, relevance,
+		super(Messages.format(
+				updateType == InitializeFinalFieldProposalCore.UPDATE_CONSTRUCTOR_NEW_PARAMETER
+						? CorrectionMessages.InitializeFieldWithConstructorParameterCorrectionProposal_description
+						: CorrectionMessages.InitializeFieldAtDeclarationCorrectionProposal_description,
```
jjohnstn, line 42: This message should be CorrectionMessages.InitializeFieldInConstructorCorrectionProposal_description

jjohnstn (changes requested):

org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/text/correction/proposals/InitializeFinalFieldProposal.java
zelenyhleb: Indeed. Fixed this as well.

zelenyhleb: > Actually, it is the 2nd message that is in error. The first should be ..WithConstructor.... as you have above and the 2nd should be ..InConstructor. There is no case here where ..AtDeclaration will be used - it is done in the separate constructor above.

Yes, I just noticed it myself and the fix is already provided. Sorry for this inconsistency.

PR #2163, opened 2025-04-15
Add check for release less than JRE in JavaSearchPage

### LINKED ISSUE #2017: Prefer not to search for references within JRE when using --release
Reference search operates on sources, but when using `--release X` while only a JDK Y (Y>X) is installed / configured, then no sources of the desired version are available. 

Possible impact:
* finding references which don't exist in the desired JDK version
* not being able to process some sources because they have syntax errors when compiled at the given release level.

For that reason I suggest that the Java Search dialog should detect this situation and when `--release` doesn't match the physical JDK then
* by default disable **Search In > JRE libraries**
* if the above is enable by the user raise a warning about the incompatibility ("cannot show search results exactly matching the configured release version" or similar).

jjohnstn:
- modify JavaSearchPage.updateOKStatus() to check if release set is less than the default JRE in which case, issue a warning message on the page
- fixes #2017

## What it does

Adds warning if searching in JRE libraries when the release is set lower than JRE version.

## How to test

See issue.

## Author checklist


jjohnstn:
Thanks @stephan-herrmann for reviewing and your suggestions.

PR #2169, opened 2025-04-17
Fix for Autocompletion takes too long

### LINKED ISSUE #2125: [Performance] Autocompletion takes too long
When working in our big Workspace (~2000 plugins), the auto-completion of methods (`Ctrl + <space>`) sometimes takes too long (sometimes it doesn't even finish).

The problem _seems_ to have surfaced when updating from 2025-03 **M2** to 2025-03 _i.e._ from **v20250123-1800** to **v20250228-0640** 3 weeks ago. Since it's a short time frame (only 1 month) I can not be 100% sure that that culprit commit occurred exactly in that time frame but I am fairly sure it did (we have 80+ developers and they are pretty eager to notify performance regressions).

I took a peek at this sample but I couldn't find anything obviously suspicious, maybe someone with more knowledge could lend a hand?

Sample: [slowCodeCompletion.zip](https://github.com/user-attachments/files/19549760/slowCodeCompletion.zip)

I only saw that a lot of time goes into this thread:

Image

## Reproducer
None at the moment. I wasn't able to reproduce this in a small WS like the one I use (Platform + PDE + JDT).

## Help wanted
I would appreciate any hints and I can test small code changes (involving 1-2 plugins) in our big WS. 
Thank you in advance!

jakub-suliga:
Fixes #2125

Old (around 1:38 min):
https://github.com/user-attachments/assets/e6173a0f-7e92-4958-aed6-074e837ddd88

New:

### DIFF
--- a/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/text/java/ContentAssistProcessor.java
+++ b/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/text/java/ContentAssistProcessor.java
@@ -23,6 +23,7 @@ import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Set;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import org.eclipse.swt.SWT;
 import org.eclipse.swt.events.SelectionAdapter;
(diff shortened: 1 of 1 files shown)


jakub-suliga:
@fedejeanne @noopur2507 @iloveeclipse Can you look over the code and give feedback. Do you want something else? 

Adding the threads imporoves the peformance a lot. Currently I am testing it if it is all correct and check if there is a cleaner way using Futures.

fedejeanne (commented):
I was testing this because I don't really get where the performance gain is coming from: there is 1 non-UI thread that is now spawning several other threads **and then joining them**. But, as far as I understand, the performance bottleneck is still _in_ that thread. Put otherwise: the `for` loop used to take a lot of time to complete because one of the providers (a `CompletionProposalCategory`) was too slow and it slowed down the execution of the `for`-loop (and therefore, of its thread), but now a slow provider will **not** slow down the _main_ `for`.loop in line 334 but it will definitely slow down the _secondary_ `for`-loop in line 353 (the one that joins to all the threads).

Can it be that the performance gain did not come from your change but instead from some cache that you happened to use when you tested?

Also, I tested this PR by artificially slowing down the instances of `CompletionProposalCategory` _e.g._ by adding this in `org.eclipse.jdt.internal.ui.text.java.CompletionProposalCategory` ...

image

... and I observed that **all** of the completion proposals are delayed until that "category" is done calculating. In my case, the _TemplateProposalComputer_ refers to the 2nd page of the proposals and the Java proposals are the 1st page.

Here's your PR behaves **without** any artificial delay: [completion_proposals_no_delay](https://github.com/user-attachments/assets/f0e78710-8276-48e5-b59b-b2a3be07dd39)
And here **with** an artificial delay: [completion_proposals_with_delay](https://github.com/user-attachments/assets/1d20faf3-65d8-4b12-a0a2-ebabfaf848b6)

As you can see, the artificial delay of 5 seconds impacts both the Java proposals (1st page) and the Template proposals (2nd page).

fedejeanne:
@jakub-suliga I propose to drop this approach (see [why](https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2125#issuecomment-2827490147))

PR #2174, opened 2025-04-22
Add SVGs to org.eclipse.jdt.ui bundles

Michael5601:
This PR adds SVGs for all icons in the bundles `eclipse.jdt.ui` and `eclipse.jdt.ui.unittest.junit` except for the following as these do not yet exist as SVG:

eclipse.jdt.ui/etool16/move_up.svg
eclipse.jdt.ui/wizban/newrecord_wiz.svg

Not all icons can be loaded via SVGs since some icons are made API via `org.eclipse.jdt.ui.ISharedImages`. There is a converter that was used for loading PNGs instead. I adjusted this converter so SVGs will be loaded: `org.eclipse.jdt.internal.ui.JavaPluginImages.createImageDescriptor(Bundle, IPath, boolean)`

I did not change the functionality in `org.eclipse.jdt.internal.ui.viewsupport.ImagesOnFileSystemRegistry.getNewFile()` by changing `.png` to `.svg` as I don't know the side effects.

Icon disablement is not yet removed.


HeikoKlare (approved):
The change is sound.

It adopts the support for SVGs added to SWT in order to provide sharp icons for any, even fractional zoom from a single source: https://github.com/eclipse-platform/eclipse.platform.swt/pull/1638

Disabled icons are not replaced by SVGs but are to be replaced with on-the-fly generation from the enabled version, see:
- https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2190

Issue #2257, opened 2025-06-05
Running JUnit with class selectors bypasses classloading in JUnit discovery phase


holly-cummins:
JUnit tests can be run in (roughly) four ways:

- Right clicking on the test class name and choosing 'run as JUnit test'
- Right clicking on the test method and choosing 'run as JUnit test'
- Right clicking on a package in the Package Explorer and choosing 'run as JUnit test'
- After tests have run, in the JUnit view, choosing a class and right-clicking and choosing 'run'

With the last three, test classloading happens after the `launcherDiscoveryStarted` hook in JUnit's `LauncherDiscoveryListener`. With the first one, classloading happens earlier. 

The reason is that a `DiscoverySelector` can be created either by passing a string, or a class. If a class is passed, JUnit populates the class name, and reads a classloader off the class object. If a string is passed, JUnit only stores the class name. If `ClassSelector::getJavaClass` is called, JUnit will use the current thread context classloader to load the class. 

```
ClassSelector(Class<?> javaClass) {
		this.className = javaClass.getName();
		this.classLoader = javaClass.getClassLoader();
		this.javaClass = javaClass;
	}
```

In `JUnit5TestLoader`, most calls to `DiscoverySelectors.select*` pass a string, but in the case where `Class<?>[] testClasses` has been set, the `selectClass` method which accepts a class is called. This means the normal discovery classloading phase is bypassed. 

The fix is to update the [selector creation](https://github.com/eclipse-jdt/eclipse.jdt.ui/blob/88f32972f22e82fde6fb4d7f056d18961c7c3c77/org.eclipse.jdt.junit5.runtime/src/org/eclipse/jdt/internal/junit5/runner/JUnit5TestLoader.java#L78) to pass `clazz.getName()` rather than `clazz`. This makes it consistent with other paths in the test loader. 

I'm happy to submit a PR for this.

PR #2258, opened 2025-06-05
Use class names for selectors to avoid locking in classloader

holly-cummins:
Fixes https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2257

## What it does

Use the class name in the selector, rather than a class object, to avoid 'locking in' the classloader of the class. Locking in the class prevents thread context classloaders being used in the discovery phase to load classes.

## How to test

To reproduce the original problem, there are two options:
- Write a JUnit DiscoveryListener which sets a thread context classloader; confirm that with the changes, tests are being loaded with the thread context classloader
- Create a minimal Quarkus application using https://code.quarkus.io/, load it into Eclipse, and then run `GreetingResourceTest` by right clicking on the class name. Without the fix the launch fails, with the fix the launch succeeds.

I've used the Oomph SDK and run an Eclipse with my changes and confirmed they fix the problem.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.junit5.runtime/META-INF/MANIFEST.MF
+++ b/org.eclipse.jdt.junit5.runtime/META-INF/MANIFEST.MF
@@ -4,7 +4,7 @@ Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-Vendor: %providerName
 Bundle-SymbolicName: org.eclipse.jdt.junit5.runtime;singleton:=true
-Bundle-Version: 1.1.300.qualifier
+Bundle-Version: 1.1.400.qualifier
 Bundle-Localization: plugin
 Export-Package: org.eclipse.jdt.internal.junit5.runner;x-internal:=true
 Require-Bundle: org.eclipse.jdt.junit.runtime;bundle-version="[3.5.0,4.0.0)",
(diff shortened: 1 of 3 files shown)


iloveeclipse:
@trancexpress : since you were recently active in this area due https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4053, could you please also check this PR?

trancexpress (approved):
According to the PRs description, the change allows further class loaders to be used to discover tests. So I double checked that the special class loader Eclipse PDE uses for plug-in test launches still works: `org.eclipse.pde.internal.junit.runtime.RemotePluginTestRunner.BundleClassLoader`

Plug-in tests can be executed without an issue (I checked plain JUnit tests too just in case, no issue as expected), I don't have other concerns here.

iloveeclipse:
@holly-cummins : the patch is integrated and you can test it in https://download.eclipse.org/eclipse/downloads/drops4/I20250606-0710/.

Note, the Junit in Eclipse is also [upgraded there from 5.12.2 to 5.13](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3067/files), so if you see any regression, please report it here.

holly-cummins:
Thank you, @iloveeclipse and @trancexpress!

PR #2260, opened 2025-06-06
Support configuration of the release for a a source folder

laeubi:
This allows the user to set the new 'release' option for source folders.

See
- https://github.com/eclipse-jdt/eclipse.jdt.core/pull/3900

### DIFF
--- a/org.eclipse.jdt.ui/META-INF/MANIFEST.MF
+++ b/org.eclipse.jdt.ui/META-INF/MANIFEST.MF
@@ -97,7 +97,7 @@ Require-Bundle:
  org.eclipse.core.filesystem;bundle-version="[1.2.0,2.0.0)",
  org.eclipse.core.resources;bundle-version="[3.14.0,4.0.0)",
  org.eclipse.core.variables;bundle-version="[3.2.200,4.0.0)",
- org.eclipse.jdt.core;bundle-version="[3.42.0,4.0.0)",
+ org.eclipse.jdt.core;bundle-version="[3.43.0,4.0.0)",
  org.eclipse.search;bundle-version="[3.10.0,4.0.0)",
  org.eclipse.debug.core;bundle-version="[3.22.0,4.0.0)",
  org.eclipse.debug.ui;bundle-version="[3.11.0,4.0.0)",
--- a/org.eclipse.jdt.ui/icons/full/obj16/release_option.svg
+++ b/org.eclipse.jdt.ui/icons/full/obj16/release_option.svg
@@ -1,0 +1,226 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!-- Created with Inkscape (http://www.inkscape.org/) -->
+
+<svg
+   width="16"
+   height="16"
+   id="svg2"
+   version="1.1"
+   inkscape:version="1.2.2 (b0a8486541, 2022-12-01)"
+   sodipodi:docname="release_option.svg"
+   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
+   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
+   xmlns:xlink="http://www.w3.org/1999/xlink"
+   xmlns="http://www.w3.org/2000/svg"
+   xmlns:svg="http://www.w3.org/2000/svg"
+   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
+   xmlns:cc="http://creativecommons.org/ns#"
+   xmlns:dc="http://purl.org/dc/elements/1.1/">
+  <defs
+     id="defs4">
+    <linearGradient
+       id="linearGradient3827">
+      <stop
+         style="stop-color:#ffffff;stop-opacity:0;"
+         offset="0"
+         id="stop3829" />
+      <stop
+         id="stop3835"
+         offset="0.60149658"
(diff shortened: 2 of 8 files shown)


stephan-herrmann:
Can we have a combo box / dropdown of all supported releases, instead of plain text input field?

laeubi:
> Can we have a combo box / dropdown of all supported releases, instead of plain text input field?

Basically, yes (as there is similar for compliance configuration) but I'm not sure if its a good UI... typing two numbers is most likely faster than searching a number in a (growing) drop down field.

Regarding [your comment here](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/3900#issuecomment-2988528371) I now added this:

grafik

stephan-herrmann:
> Basically, yes (as there is similar for compliance configuration) but I'm not sure if its a good UI... typing two numbers is most likely faster than searching a number in a (growing) drop down field.

Efficiency is one thing, consistency another. Essentially all places in the IDE where users select Java versions are implemented as combos, no?

We minimally need full validation, otherwise you can pass release 10000 into the compiler and then expect what?

:x:  `Compiling for Java version '' is no longer supported. Minimal supported version is '1.8'`

ups?

laeubi:
@stephan-herrmann the dialog now looks like this:

grafik

laeubi:
> I think it would be good to have a new dedicated icon

Maybe @BeckerWdf can help / suggest one, but I think an icon can be changed anytime so I really don't see this as a blocker ... 

> before this feature is released

Could I see this as a sign that we are close to merge this one? :-)

- https://github.com/eclipse-jdt/eclipse.jdt.core/pull/3900

BeckerWdf:
> > I think it would be good to have a new dedicated icon
> 
> Maybe @BeckerWdf can help / suggest one, but I think an icon can be changed anytime so I really don't see this as a blocker ...
> 
> > before this feature is released
> 
> Could I see this as a sign that we are close to merge this one? :-)
> 
> * [Add multi-release compilation support to JDT eclipse.jdt.core#3900](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/3900)

I will have a look once I’m back from vacation.

laeubi:
@stephan-herrmann I now created a dedicated icon based on the `bundleversion` and the `classfile` icon. I think a folder is not really suitable here (the parent element is already a folder icon) and it looks like this now:

grafik

As it is now only a matter of changing one file, maybe this can be merged and the icon be polished afterwards, I'll then wait with writing a N&N until it is finalized.

stephan-herrmann:
OK, I merged this for now, to facilitate testing before the release.

Nitpick: inserting the new classpathAttributeConfiguration between TestAttributeConfiguration and WithoutTestCodeAttributeConfiguration makes no sense :)

As for the icon: having a folder item as a child of another folder item doesn't look unusual to me :)

OTOH, I still don't understand what the class icon is to convey here.

Taking a step back I'm seeing the following difficulty (which also surfaced in our discussion on the implementation PR in jdt.core): The term "release" connects several quite different topics:
* source files are compiled with different build path:
  * using different views of the underlying physical JDK (`--release`)
  * using different slices of the current project
* source files may use java features of different versions (`--source`)
* source files are compiled to byte code of different versions (`--target`)

Actually, the first item suggest that we could use the library icon with version overlay!

Since there is not one single versioned thing, perhaps we should use the [1.0] box not as an overlay but as the main icon (fit to icon size).

stephan-herrmann:
> > As for the icon: having a folder item as a child of another folder item doesn't look unusual to me :)
> 
> But should then not all items in this area have a folder icon as they are related to a source folder? This dies not sounds right to me.

You're twisting my words backwards. Of course any entry that has a suitable visualization will keep that.

BTW, we have a precedent: Ignore optional compile problems has a decorated source-folder icon.

> > OTOH, I still don't understand what the class icon is to convey here.
> 
> Basically on a very high level you control the class file version this classes are compiled with.

If we teach people that this is the meaning of the option, then they will not understand the full power of it.

BTW, "class file" has a different icon, but that is not suitable for decorating with a version number "1.0", because it already contains a "010" glyph to signal "binary" ;-P

Let's try to get at least one more opinion.

danthe1st:
Just providing alternative suggestions for icons:
- Since multi-release builds are mostly used for multi-release _JARs_, using the [JAR icon](https://github.com/eclipse-jdt/eclipse.jdt.ui/blob/master/org.eclipse.jdt.ui/icons/full/obj16/jar_obj.png) (possibly without the 010 glyph) might be appropriate (if applicable with some decoration to indicate the version
- Maybe something like [this](https://github.com/eclipse-jdt/eclipse.jdt.ui/blob/master/org.eclipse.jdt.ui/icons/full/obj16/empty_logical_package_obj.png) which shows multiple packages (empty or nonempty)
- maybe something similar to the [`java_app` icon](https://github.com/eclipse-jdt/eclipse.jdt.ui/blob/master/org.eclipse.jdt.ui/icons/full/dtool16/java_app%402x.png) (but without the `+`)
- Instead of class files, maybe a variant of the [symbol for compilation units](https://github.com/eclipse-jdt/eclipse.jdt.ui/blob/master/org.eclipse.jdt.ui/icons/full/obj16/cu_obj%402x.png) or [that one](https://github.com/eclipse-jdt/eclipse.jdt.ui/blob/master/org.eclipse.jdt.ui/icons/full/obj16/jcu_obj.png) would work?

Suggested icons in image form:
JAR icon packages `java_app` icon compilation unit Java compilation unit

PR #2266, opened 2025-06-12
Bump some versions

stephan-herrmann:
Ensure plugins with changes in the branch have higher version than master (+ .50).

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -21,7 +21,8 @@ pipeline {
 						clean verify --batch-mode --fail-at-end \
 						-Pbree-libs -Papi-check -Pjavadoc -Pbuild-individual-bundles \
 						-Dmaven.test.failure.ignore=true \
-						-Dcompare-version-with-baselines.skip=false
+						-Dcompare-version-with-baselines.skip=true \
+						-DcompilerBaselineMode=disable -DcompilerBaselineReplace=none
 					"""
 				}
 			}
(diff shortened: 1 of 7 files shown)


stephan-herrmann:
This commit combines various fixes to update jdt.ui for the next patch build, and then to get a green PR build in the first place.

@noopur2507 @jarthana  what do you think specifically of https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2266/commits/1565ecb4da571ca9108a8fcf3ccb2286b546b886 for now, i.e., disabling the checks that otherwise would raise alarm as shown in [this comment](https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2266/commits/1565ecb4da571ca9108a8fcf3ccb2286b546b886). The alternative would be to really do a regular merge-from-master *every time* some version bump occurs in master, which currently seems to happen about every other day. We could then re-enable these checks closer to the release when versions are more stable.

My goal is to be able to build P-builds quite frequently already now, to allow for quick feedback loops. Maybe for 25 it's not crucial as this is a low traffic release, but still I'd like to demonstrate that JDT supports early testing with the next Java version.

The final failure is beyond me: the check for merge commits breaks for its own set of problems, not ours. This clearly is false alarm.

jarthana:
> This commit combines various fixes to update jdt.ui for the next patch build, and then to get a green PR build in the first place.
> 
> @noopur2507 @jarthana what do you think specifically of [1565ecb](https://github.com/eclipse-jdt/eclipse.jdt.ui/commit/1565ecb4da571ca9108a8fcf3ccb2286b546b886) for now, i.e., disabling the checks that otherwise would raise alarm as shown in [this comment](https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2266/commits/1565ecb4da571ca9108a8fcf3ccb2286b546b886). The alternative would be to really do a regular merge-from-master _every time_ some version bump occurs in master, which currently seems to happen about every other day. We could then re-enable these checks closer to the release when versions are more stable.

I am fine with this.

PR #2272, opened 2025-06-16
Cleanup SVGs

BeckerWdf:


BeckerWdf:
can somebody of the JDT committers pls. merge this?

BeckerWdf:
it's just deleting invisible stuff from the SVGs. So it's correct that you can't see a difference.

PR #2282, opened 2025-06-20
allow overlapping start and end markers for custom folding regions

danthe1st:
## What it does

In #1825, I contributed custom folding regions but this does not allow using the same marker/prefix for the start and end marker comments.

Looking at classes like [`AddDelegateMethodsAction`](https://github.com/eclipse-jdt/eclipse.jdt.ui/blob/e25880c065ecc1534a94da98dd10f7c957b73719/org.eclipse.jdt.ui/ui/org/eclipse/jdt/ui/actions/AddDelegateMethodsAction.java), I can see comments splitting the code into different regions where one region ending and another one starting is indicated with the same comment.

Assuming that there are also other projects that would like to use that, I am changing custom folding regions to allow using the same text for start and end markers. The same comment is allowed to be both the end and start of a region.

If the region marker texts aren't mutually exclusive (i.e. if one of them starts with another), the last line is not included in the region.

## Questions for reviewers/maintainers (asking for opinions)

In cases like the one I mentioned, there is typically no comment ending the last region. So I guess it would be good to automatically end custom folding regions if there is no corresponding comment/end marker at the end of the surrounding element. Should I add this?
If so, should this be the case only for non-exclusive start/end regions (e.g. `----` as opposed to `startregion`/`endregion`)? Should this be the case only if the start and end markers are the same or also if one has additional text (e.g. `region`/`regionend`)?
In principle, this could also be a preference but to be honest, I guess there's no need to configure this.

For now, I implemented it in a way that
- If the start and end region markers are overlapping (one starts with another):
  - The end marker comment is excluded from folding e.g. if a region starts and ends with `// ---- ...`, both are visible if the regions are collapsed
  - All non-closed regions end automatically at the end of the block

Doing this for all regions ensures that the end comments are either always shown or never shown for a project (or everything with the same configuration (region markers)).

This includes a change of behavior of custom folding regions shown in `CustomFoldingRegionTest#testCustomFoldingRegionsMultipleLevels()` (see the diff of the test) if the extended folding is active. Before, it allowed custom folding regions spanning spanning over different top level types or starting outside of a top level type and ending within that top level type. Since this PR requires tracking top level types anyway, this confusing behavior is not there any more and custom folding regions behave the same way (in this regard) no matter whether enhanced folding is active or not.

## How to test

- Enable Window > Preferences > Java > Editor > Folding > Enable Folding of custom regions
- In the same preference window, set the start and end region comment text to the same string.
- Edit a Java file like the following:
  ```java
  public class Test {
      // region start first
      
      // region end first, start second
      
      // region end second
  }
  ```

This should add folding regions like this:
image

This should work both with and without the "Extended Folding" of control structures.

One simple example where this can be used is a class like the following (using `----` as both the start and end region marker):
```java
package org.example.test;

import java.time.LocalDateTime;
import java.util.Objects;

class Test {
	
	// ---- variables ---------------------------------------------
	
	private int i;
	private String s;
	
	private LocalDateTime currentDateTime = LocalDateTime.now();
	
	// ---- constructors ------------------------------------------
	
	public Test() {
		i = 0;
		s = "";
	}
	
	public Test(int i, String s) {
		this.i = i;
		this.s = s;
	}

	// ---- getters/setters ---------------------------------------

	public String getS() {
		return s;
	}

	public void setS(String s) {
		this.s = s;
	}

	public LocalDateTime getCurrentDateTime() {
		return currentDateTime;
	}

	public void setCurrentDateTime(LocalDateTime currentDateTime) {
		this.currentDateTime = currentDateTime;
	}
	
	// ---- equals, hashCode, toString ----------------------------
	
	@Override
	public String toString() {
		return "Test [i=" + i + ", s=" + s + ", currentDateTime=" + currentDateTime + "]";
	}

	@Override
	public int hashCode() {
		return Objects.hash(currentDateTime, i, s);
	}

	@Override
		if (this == obj)
			return true;
		if (obj == null)
			return false;
		if (getClass() != obj.getClass())
			return false;
		Test other = (Test) obj;
		return Objects.equals(currentDateTime, other.currentDateTime) && i == other.i && Objects.equals(s, other.s);
	} 
	
} 
```

image

## Author checklist

### DIFF
--- a/org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/folding/CustomFoldingRegionTest.java
+++ b/org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/folding/CustomFoldingRegionTest.java
@@ -58,7 +58,7 @@ public class CustomFoldingRegionTest {
 	@Parameter
 	public boolean extendedFoldingActive;
 
-	@Parameters(name = "Experimental folding active: {0}")
+	@Parameters(name = "Extended folding active: {0}")
 	public static Object[] extendedFoldingElements() {
 		return new Object[] { true, false };
 	}
@@ -257,13 +257,7 @@ public class CustomFoldingRegionTest {
 				""";
 		List<IRegion> projectionRanges= getProjectionRangesOfFile(str);
 		assertEquals(4, projectionRanges.size());
-		if (extendedFoldingActive) {
-			// folding with control structures does not consider top level classes
-			// hence the first region ends with the endregion comment inside the Test class
-			FoldingTestUtils.assertContainsRegionUsingStartAndEndLine(projectionRanges, str, 1, 3);
-		}else {
-			FoldingTestUtils.assertContainsRegionUsingStartAndEndLine(projectionRanges, str, 1, 13);//outside class
-		}
+		FoldingTestUtils.assertContainsRegionUsingStartAndEndLine(projectionRanges, str, 1, 13);//outside class
 		FoldingTestUtils.assertContainsRegionUsingStartAndEndLine(projectionRanges, str, 4, 11);//outside method
 		FoldingTestUtils.assertContainsRegionUsingStartAndEndLine(projectionRanges, str, 7, 9);//inside method
 		FoldingTestUtils.assertContainsRegionUsingStartAndEndLine(projectionRanges, str, 5, 9);//void a()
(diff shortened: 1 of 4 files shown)


danthe1st:
Would @iloveeclipse or someone else care to take a look at this PR?

jjohnstn (changes requested):
Really like the feature - especially in the context of being able to fold sections of existing code.

Minor formatting comment to start.

Please follow JDT UI formatting.  ifs, for loops, while statements, etc.  should have a space between the keyword and the opening parenthesis.

jjohnstn:
Thanks @danthe1st, I tagged this as noteworthy so we will remember to add a blurb in the N&N for 2025-09.

PR #2284, opened 2025-06-21
Use package imports for commonmark

merks:


merks:
@stephan-herrmann 

I think the build problem here are not related to the change but need investigation.

---

The BSN of pretty much all the commonmark bundles have changed:

image

So better we just use package imports that don't care about the BSN...

stephan-herrmann:
GitHub informs us:
> -> Some quality gates have been missed: overall result is UNSTABLE

well??

the [linked logs](https://ci.eclipse.org/jdt/job/eclipse.jdt.ui-github/job/PR-2284/1/execution/node/42/log/) don't mention any problem.

Somewhere else on Jenkins I found

> [INFO] --- tycho-p2-plugin:5.0.0-SNAPSHOT:p2-metadata-default (default-p2-metadata-default) @ org.eclipse.jdt.ui ---
[WARNING] MavenProject: org.eclipse.jdt:org.eclipse.jdt.ui:3.35.100-SNAPSHOT @ /home/jenkins/agent/workspace/eclipse.jdt.ui-github_PR-2284/org.eclipse.jdt.ui/pom.xml: baseline and build artifacts have same version but different contents

Is someone telling us that you succeeded to produce modified content in org.eclipse.jdt.ui without even touching a single file in the project?

But what exactly was the reason to mark the build as failed?

danthe1st:
The build of another PR from yesterday (that I wrote) [failed for the same reason just with a different class](https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2282#issuecomment-2993584806) so this is likely unrelated to this PR (maybe because the last `master` build [was unstable as well](https://github.com/eclipse-jdt/eclipse.jdt.ui/runs/44377231143)?)

HannesWell:
> So better we just use package imports that don't care about the BSN...

If bundles provide properly versioned packages (which most Eclipse bundles/plugins don't do (yet), unlike the most third party dependencies we use), the general recommendation is to use `Import-Package` over `Require-Bundle`

> GitHub informs us:
> 
> > -> Some quality gates have been missed: overall result is UNSTABLE
> 
> well??
> 
> the [linked logs](https://ci.eclipse.org/jdt/job/eclipse.jdt.ui-github/job/PR-2284/1/execution/node/42/log/) don't mention any problem.

Unfortunately the GH UI directly links to the `Pipeline Overview` and not the general overview of the latest build. You can reach the latter by clicking on the build number:


While the `Pipeline Overview` looks a bit more fancy, it doesn't provide all the information the general overview of a build provides. I'll try out some Jenkins settings and can ask the IT to change the default link (if possible and you want to).

That being said, if you scroll down the general overview page of the build you get the failing quality gates and get all the details if you click on the number of the new issues:



> Is someone telling us that you succeeded to produce modified content in org.eclipse.jdt.ui without even touching a single file in the project?

Yes, that's right. This could theoretically have different reasons (see below).

> But what exactly was the reason to mark the build as failed?

It seems that the baseline comparison warning counts as compiler warning, but I'm not sure why this is and if that's intended.


> Found something :
> 
> > baseline and build for org/eclipse/jdt/internal/ui/javaeditor/SemanticHighlightingReconciler.class have different contents
> 
> Where would I find the class file from the "baseline"?

The baseline should be the latest I-build, so for new it should be https://download.eclipse.org/eclipse/updates/4.37-I-builds/I20250621-1800/.

One reasons for this could theoretically be that different compiler versions have been used in the latest I-build than is used in this verification build and looking at the logs it says:
```
[INFO] --- tycho-compiler:5.0.0-SNAPSHOT:compile (default-compile) @ org.eclipse.jdt.ui ---
[INFO] Compiling 1721 source files to /home/jenkins/agent/workspace/eclipse.jdt.ui-github_PR-2284/org.eclipse.jdt.ui/target/classes using Eclipse Compiler for Java(TM) 3.42.0.v20250526-2018
```
This is the latest released compiler (I guess) but not the latest snapshot from the master, that the I-build is using. I'll try to investigate why, this is the case.
To confirm that this is the problem, I suggest to also rerun the master branch build (can be done through the Jenkins UI by either clicking `Rebuild` on a specific build or clicking `Build now` at the overview of the `master` branch) and see if the same error happens.

Another cause could be that this change has the effect the JDT classes in question are build against a different (common-mark) API, but as the affected project is not changed at all, I cannot tell how this should happen.
So my guess is currently the first assumption.

stephan-herrmann:
Thanks, @merks !

PR #2293, opened 2025-06-26
Record Javadocs not shown in Content Assist suggestions #4134

jarthana:
## What it does
Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4102

## Author checklist


jarthana:
One thing that is kind of iffy in the PR is, MethodProposalInfo, which is supposed to be dealing only with methods is forced to return a field. I guess it's OK since the doc only says about members in general. If anyone sees a problem, let me know.

srikanth-sankaran:
Once all is green, we can scheduled a walk through and discuss, Thanks

noopur2507:
> One thing that is kind of iffy in the PR is, MethodProposalInfo, which is supposed to be dealing only with methods is forced to return a field. I guess it's OK since the doc only says about members in general. If anyone sees a problem, let me know.

This seems OK.

noopur2507 (approved):
The only concern is related to retrieving the Javadoc of the record type instead of the particular record component for the given accessor method. Other changes look good.

PR #2302, opened 2025-06-29
enable folding with "Only show selected Java element"

### LINKED ISSUE #2264: Eclipse new folding feature does not work when use show only selected element feature
When you put one the show only selected element feature
 (A great killer underrated feature of Eclipse that is somewhat hidden and should put in the front)
in order to focus working on method only in a class
all the folding little icons disappear from the editor and you can not do any folding on the focus method
not until you disable the show only selected menu to the editor
once you  see all the methods in the class the  folding icons appear again in the editor

danthe1st:
Fixes #2264 
Depends on https://github.com/eclipse-platform/eclipse.platform.ui/pull/3074 (Is it necessary to change the version requirements in some way even if I am not depending on any new API?)
~~Depends on #2410 _for this to work correctly with extended folding_~~ That dependency PR has been merged.

I think this should be tested with JDT-UI (with this PR) before https://github.com/eclipse-platform/eclipse.platform.ui/pull/3074 is merged.

## What it does
This PR enables folding when Window > Preferences > Java > Editor > Only show the selected Java element is enabled.

This depends on https://github.com/eclipse-platform/eclipse.platform.ui/pull/3074 to work correctly.

## How to test
- ~~Include https://github.com/eclipse-platform/eclipse.platform.ui/pull/3074~~ not necessary any more since that's merged, just make sure to use the latest I-build of eclipse.platform.ui
- Enable Window > Preferences > Java > Editor > Only show the selected Java element
- Enable folding in Window > Preferences > Java > Editor > Folding
- Open a Java file
- Folding should work
- Select an element in the outline
- Observe that only the selected element is shown in the editor.
- Folding should still work
- All of that should work with and without extended folding or custom folding regions. Editing the file shouldn't break it.

## Author checklist

  - ~~If only a selected region is shown and one enters text at the end of that region (if the curser is _exactly_ after the closing `}`), everything after the selected region is displayed. I am not sure whether that's an issue. See https://github.com/user-attachments/assets/18cc621f-1c69-45e6-a7ce-97d609f489fe~~ This was an issue with my changes in eclipse.platform.ui which should be fixed.
  - I did not do any performance evaluation.

### DIFF
--- a/org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/folding/FoldingWithShowSelectedElementTests.java
+++ b/org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/folding/FoldingWithShowSelectedElementTests.java
@@ -67,7 +67,7 @@ public class FoldingWithShowSelectedElementTests {
 	}
 
 	@After
-	public void tearDown() throws CoreException {
+	public void tearDown() {
 		JavaPlugin.getDefault().getPreferenceStore().setToDefault(PreferenceConstants.EDITOR_SHOW_SEGMENTS);
 		JavaPlugin.getDefault().getPreferenceStore().setToDefault(PreferenceConstants.EDITOR_FOLDING_CUSTOM_REGIONS_ENABLED);
 	}


danthe1st:
I wrote some basic tests for the PR in eclipse.platform.ui. Is it also necessary to add integration tests in JDT-UI for this change?

iloveeclipse:
> I wrote some basic tests for the PR in eclipse.platform.ui. Is it also necessary to add integration tests in JDT-UI for this change?

If possible, regression test for original problem?

danthe1st:
> If possible, regression test for original problem?

I added a test that ensures folding is active (even though I am not sure whether this is actually a bug I'm fixing or more implementing a feature/enhancement).

danthe1st:
The test failure should be because #3074 is not merged.

danthe1st:
The changes I made to `DefaultJavaFoldingStructureProvider` (making sure the `IJavaElement` is present) are necessary because it uses the `IJavaElement` to determine which folding regions are equal. Without this, it tries to update the existing folding projection annotations in a way that caused issues with my change (I think it tried to set existing folding regions to other ones).

org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/folding/FoldingWithShowSelectedElementTests.java
```diff
@@ -0,0 +1,190 @@
...
+		ICompilationUnit cu= packageFragment.createCompilationUnit("A.java", str, true, null);
+		JavaEditor editor= (JavaEditor) EditorUtility.openInEditor(cu);
+
+		try {
+			editor.setSelection(cu.getElementAt(str.indexOf("someMethod")));
+
+			assertEquals("""
+						void someMethod() {
+							// content here
+						}\t\t\t\t
```
danthe1st, line 169: Including trailing whitespace seems to be necessary because otherwise it would hide the closing `}`.

iloveeclipse:
Seem to work fine except the binary types (e.g. java.lang.String etc), I guess similar change like it was in JavaEditor is needed for ClassFileEditor.

danthe1st:
> Seem to work fine except the binary types (e.g. java.lang.String etc), I guess similar change like it was in JavaEditor is needed for ClassFileEditor.

Did the "Only show the selected Java element" option ever work with the Class File Editor? I didn't see that doing anything without my change.

iloveeclipse:
> > Seem to work fine except the binary types (e.g. java.lang.String etc), I guess similar change like it was in JavaEditor is needed for ClassFileEditor.
> 
> Did the "Only show the selected Java element" option ever work with the Class File Editor? I didn't see that doing anything without my change and this PR doesn't seem to impact that.

Honestly, no idea, I never used this "Only show the selected Java element" option, just assumed it should work there too :-)
So if it wasn't ther ebefore, don't care!

danthe1st:
The changes to `DefaultJavaFoldingStructureProvider` (and `CustomFoldingRegionTest`) are necessary to prevent issues with changing the document with enhanced folding in cases like this:
```java
// region outer

class SomeClass {
    // region inner
    void someMethod() {
        // content
    }
    // endregion
}

// endregion
```

That class would have the following issues when the file is modified:
- With the changes to allow folding with "Only show selected Java element" without the changes I made to `DefaultJavaFoldingStructureProvider`, if `someMethod` is selected (and only that is shown), editing the code results in the full `inner` region being shown
- Even without this PR/in the current `master` (or without the changes to `DefaultJavaFoldingStructureProvider`), if `region outer` is collapsed, editing the file expands it

The reason this happens is that it tries to "update" the `outer` region with the `inner` region or vice-versa (it confuses the folding regions with each other) so it re-expands the old folding region.

If wanted, I could factor out this fix to its own PR.

jjohnstn (changes requested):
See main-line comment

danthe1st:
> When I select test() in the outline view, it shows test() only but it also shows the // region comment above it (AST elements get tied to line comments above them and on the same line). It is recognized as a custom folding comment but clicking on it does not hide the test() method. It adds the end-of-line stuff in a normal folding on all of the test() lines. I can confirm that folding within test() works fine.

When writing the Platform-UI patch, I ensured that it is not possible to fold anything outside of the visible region (as it didn't make sense IMO). I think I can provide another PR there that ensures that these annotations are hidden.

danthe1st:
@jjohnstn Would you like to review this now for 2026-03? There are still https://github.com/eclipse-platform/eclipse.platform.ui/pull/3457, https://github.com/eclipse-platform/eclipse.platform.ui/pull/3456 and https://github.com/eclipse-platform/eclipse.platform.ui/pull/3585 but I think these shouldn't block this review.

PR #2331, opened 2025-07-08
Move getCreateInSuperClassProposals(..) to jdt.core.manipulation.

rgrunber:
- Addresses https://github.com/redhat-developer/vscode-java/issues/4064
- Move QuickAssistProcessor.getCreateInSuperClassProposals(..) into jdt.core.manipulation for ease of reuse

The methods being refactored are tested by `testCreateMethodWhenOverrideAnnotation`, `testOverrideAnnotationButNotOverriding` from `ModifierCorrectionsQuickFixTest`, and `testCreateInSuperInGeneric`, `testCreateInSuper` from `AssistQuickFixTest`

- [x] Before merging, need to investigate whether https://github.com/eclipse-jdt/eclipse.jdt.ui/blob/b66e1d4cd08ac520157780a010047c2cd6fc6d97/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/text/correction/proposals/NewDefiningMethodProposal.java#L77-L78 can also be moved as currently we have a parameter `addOverride` being passed around that isn't actually used.

### DIFF
--- a/org.eclipse.jdt.core.manipulation/common/org/eclipse/jdt/internal/ui/text/correction/ModifierCorrectionSubProcessorCore.java
+++ b/org.eclipse.jdt.core.manipulation/common/org/eclipse/jdt/internal/ui/text/correction/ModifierCorrectionSubProcessorCore.java
@@ -20,6 +20,7 @@ package org.eclipse.jdt.internal.ui.text.correction;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashSet;
+import java.util.LinkedList;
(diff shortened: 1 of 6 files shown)


rgrunber:
I tired to move `addOverrideAnnotation` but it gets complicated so I'd prefer to leave it for now. The problem is that the code needs to potentially rewrite a completely different compilation unit (eg. create method in base class, and add "Override" in current class) . I don't see how proposals can modify multiple compilation units at once.

PR #2349, opened 2025-07-17
Do not return null editor input from the ClassFileEditorInputFactory

laeubi:
Currently `ClassFileEditorInputFactory` return null in some cases when restoring an editor input from an `IMemento`. While this is legal in general for `IElementFactory` it is not very useful for factories that restore editor inputs because this will fail to instantiate the editor with a nasty exception without any way for the editor to handle or the user to understand the underlying problem.

This now changes two things:

1) In case of an error, a special `ExceptionalEditorInput` is created, that when later used to gather its underlying data will rethrow the original exception so we get a useful message about the cause.
2) In case of an element not found by the project an `InternalClassFileEditorInput` is returned as we actually have the class data already there.

### DIFF
--- a/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/javaeditor/ClassFileEditorInputFactory.java
+++ b/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/javaeditor/ClassFileEditorInputFactory.java
@@ -18,6 +18,7 @@ import org.eclipse.core.runtime.IAdaptable;
 
 import org.eclipse.ui.IElementFactory;
 import org.eclipse.ui.IMemento;
+import org.eclipse.ui.IPersistableElement;
 
 import org.eclipse.jdt.core.IClassFile;
 import org.eclipse.jdt.core.IJavaElement;
@@ -46,7 +47,18 @@ public class ClassFileEditorInputFactory implements IElementFactory {
 	public IAdaptable createElement(IMemento memento) {
 		String identifier= memento.getString(KEY);
 		if (identifier == null) {
-			return null;
+			return new ExceptionalEditorInput("memento data", new IPersistableElement() { //$NON-NLS-1$
+
+				@Override
+				public void saveState(IMemento m) {
+					//nothing to save...
+				}
+
+				@Override
+				public String getFactoryId() {
+					return ID;
+				}
+			}, new IllegalStateException(String.format("No %s present in memento", KEY))); //$NON-NLS-1$
 		}
 		IJavaElement element= JavaCore.create(identifier);
 		try {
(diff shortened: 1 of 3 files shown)


org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/javaeditor/ClassFileEditorInputFactory.java
```diff
@@ -61,14 +73,25 @@ public IAdaptable createElement(IMemento memento) {
 				if (project != null) {
 					type= project.findType(type.getFullyQualifiedName());
 					if (type == null)
-						return null;
+						return new InternalClassFileEditorInput(cf);
```
laeubi, line 76: If one wants, here we can pass a boolean to Indicate that the type was not found on the project if that's useful for anything later on.

laeubi:
@iloveeclipse would you mind merging this PR as it already improves the situation? I then plan to do another smaller PR and then continue with
- https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2351

org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/javaeditor/ExceptionalEditorInput.java
```diff
@@ -0,0 +1,79 @@
...
+	}
+
+	@Override
+	public IPersistableElement getPersistable() {
+		return fElement;
+	}
+
+	@Override
+	public String getToolTipText() {
+		return String.format("Error while gathering input for %s: %s", fIdentifier, fException); //$NON-NLS-1$
```
mehmet-karaman, line 61: Is there not an externalization necessary?

mehmet-karaman: I don't know if there is a separation inbetween UI and Exception messages. But this could be used in the UI

org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/javaeditor/OrdinaryClassFileEditorInput.java
```diff
@@ -0,0 +1,29 @@
...
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Christoph Läubrich - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.jdt.internal.ui.javaeditor;
+
+import org.eclipse.jdt.core.IOrdinaryClassFile;
+
+public class OrdinaryClassFileEditorInput extends InternalClassFileEditorInput {
```
mehmet-karaman, line 18: Why don't we use directly the InternalClassFileEditorInput and create a new class without any additional functionality or marking (Checked also against the check_incomplete_input branch but couldn't find the use of it)

mehmet-karaman: In overall the solution looks good. The second PR has no issues from my side.. The last PR (load the Sources when it is available) worked only once and then never again.. I will try to understand the problem, did you pushed your latest changes there?

laeubi: Lets first finish this ones 

- https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2357 
- https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2353

before further proceed, it all gets to much complicate to have even more PRs inflight. Once this is done I can rebase and we can easier discuss on possible improvements or increments.

mehmet-karaman: Is the order of the PRs correct? 

1. https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2349
2. https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2357
3. https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2353

mehmet-karaman: 2349 contains already the change from 2353 .. 

The order might be:
2353
2349
2357

laeubi:
Closing as not enough interest from JDT maintainers and 

- https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2353

is simpler.

PR #2357, opened 2025-07-18
If there is no source show the raw class file content

laeubi:
Currently if ClassFileEditor is opened for a classfile that is not on the classpath it fails with an error "The class file is not on the classpath" even though it could just display the raw class.

This now adjust the ClassFileEditor in a way to be more lenient and allows to display class files that are not on the classfile at least with the default raw view.

FYI @iloveeclipse


mehmet-karaman:
It doesn't displays the class file in the default raw view: 
image

iloveeclipse:
@laeubi  & @mehmet-karaman : could you both please provide steps how to reproduce? Maybe you use different use cases?

laeubi:
@iloveeclipse just to make sure you have no other changes in JDT.core or PDE in your workspace, so only JDT.ui?

mehmet-karaman (commented):
Looks good to me.

laeubi:
I'm closing this in favor of
- https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2380

Issue #2365, opened 2025-07-21
org.eclipse.jdt.internal.ui.text.JavaReconciler manages fIsEditorActive poorly.


merks:
The state of the field determines the result of isEditorActive():

https://github.com/eclipse-jdt/eclipse.jdt.ui/blob/25537d682407cbc66dcbeb4b1b90d41b679fc0b2/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/text/JavaReconciler.java#L491-L493

That method is used to guard calls to forceReconciling().

https://github.com/eclipse-jdt/eclipse.jdt.ui/blob/25537d682407cbc66dcbeb4b1b90d41b679fc0b2/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/text/JavaReconciler.java#L173-L174

But I think the concept is a bit confused and mismanaged.  It essentially manages a single active editor even where my workbench looks like this where there are 4 visible editors across 3 shells and 2 workbench pages:

<img width="674" height="1131" alt="Image" src="https://github.com/user-attachments/assets/c210d926-8184-4857-b160-bea00eb33129" />

I think the logic is there mostly to avoid reconciling an editor that isn't visible and visible versus active have become conflated.

One side-effect of this is when the classpath is updated after the editor has been opened, the even is ignored.  We see above that the editor in the bottom window has errors, and with the tracing in the PR I'll create, we can see the follow thing happening.
```
TMException.java : setEditorActive=true -> true
TMException.java : setJavaModelChanged=false -> false
TMException.java : setEditorActive=true -> false   <!-- This will ignore events now.  This happens when the second window is activated.
Injection.java : setEditorActive=true -> true
Injection.java : setEditorActive=true -> true
TMException.java : setJavaModelChanged=false -> true  <--- This event should reconcile but does not.
IToken.java : setJavaModelChanged=false -> true
ITokenizeLineResult.java : setJavaModelChanged=false -> true
Injection.java : setJavaModelChanged=false -> true
```

It's interesting to consider that shell activation behaves differently and will set active all visible editors:

https://github.com/eclipse-jdt/eclipse.jdt.ui/blob/25537d682407cbc66dcbeb4b1b90d41b679fc0b2/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/text/JavaReconciler.java#L136

[...]

PR #2366, opened 2025-07-21
Improve management of JavaReconciler.fIsEditorActive

merks:
https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2365

### DIFF
--- a/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/text/JavaReconciler.java
+++ b/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/text/JavaReconciler.java
@@ -41,8 +41,8 @@ import org.eclipse.jface.text.reconciler.MonoReconciler;
 
 import org.eclipse.ui.IEditorInput;
 import org.eclipse.ui.IFileEditorInput;
-import org.eclipse.ui.IPartListener;
-import org.eclipse.ui.IWorkbenchPart;
+import org.eclipse.ui.IPartListener2;
+import org.eclipse.ui.IWorkbenchPartReference;
 import org.eclipse.ui.IWorkbenchPartSite;
 import org.eclipse.ui.IWorkbenchWindow;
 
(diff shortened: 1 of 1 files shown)


iloveeclipse:
Interestingly, **now** github shows me different picture 
image

I don't know what's going on, just undraft, I can block PR by negative review.

iloveeclipse:
@merks : I see the code still uses "old" `IPartListener` while we already have `IPartListener2` that has probably what you want (partVisible/partHidden). Do you want to check that?

merks:
That's definitely cleaner, simpler, and more obvious what it's doing!! 

I've removed all the content that made it a draft and improve the commit message.

iloveeclipse:
> I still do occasionally see that the call in this to set the classpath
> 
> ```
> org.eclipse.pde.internal.core.RequiredPluginsInitializer.DeferredClasspathContainerInitializerJob.run(IProgressMonitor)
> ```
> 
> does not result in a call to
> 
> ```
> org.eclipse.jdt.internal.ui.text.JavaReconciler.ElementChangedListener.elementChanged(ElementChangedEvent)
> ```

See https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2351/files#r2222825448

I hope to find time to debug this further.

> So I fear there are still timing issues, but at least with this change, the editor will definitely update if/when it has been notified. 

I believe that's the main issue that we don't get notification because the delta is empty for some reason.

> So there still appears to be small window of time when JDT "just doesn't listen".

It listens but the empty delta is not sent to listeners (which makes perfect sense).

merks:
I did work on a hack that really works, but it was ugly which involves joining to block the reconcilation:
```
	protected void initialProcess() {
		try {
			Job.getJobManager().join(JavaUI.ID_PLUGIN, new NullProgressMonitor());
		} catch (OperationCanceledException | InterruptedException e) {
			// TODO Auto-generated catch block
			e.printStackTrace();
		}
```

And mangling PDE so that the join waits for the classpath job:

image

🤮

Issue #2028, opened 2025-02-17
Copying class freezes UI


fedejeanne:
> [!IMPORTANT]
> This happened only sporadically in an older version of Eclipse. It's not clear if it still happens in the newest Eclipse version so feel free to handle it with a lower priority for now
>

Eclipse version: **2024-06**
OS: **Windows**

When copying a class to the clipboard (`Ctrl + c`), the UI froze. I noticed that the UI thread was blocked, waiting for something to be finished (probably the `DeduplicationUtil::intern`):

![Image](https://github.com/user-attachments/assets/cbf20db2-5d40-4f1c-9d41-7781b3980896)

This is in the thread `Worker-1413: Semantic Highlighting Job`

![Image](https://github.com/user-attachments/assets/b4a8e2ab-551a-4d6b-b35e-f8b7a9951946)

I see that:
- ZIP files are being cached but the calls to `JavaModelManager.getZipFile` need more than 2 minutes anyway
- `DeduplicationUtil.intern` needs 17 seconds

Here's the sample: [snapshot-1738739823412.zip](https://github.com/user-attachments/files/18827102/snapshot-1738739823412.zip)

jukzi:
DeduplicationUtil.intern is certainly not the problem, as it returns within fixed time. Could only be that it is called too frequently.
It is not normal that opening zip files takes 2 min. Even if you have thousands of zipfiles in your workspace. Probably that is only called too often (multiple times for the same file). 
One might try/think about caching the result of `Util.getJdkLevel` in JarPackageFragmentRoot or analyze why same JAR is opened multiple times.

fedejeanne:
> It is not normal that opening zip files takes 2 min. Even if you have thousands of zipfiles in your workspace. Probably that is only called too often (multiple times for the same file).

Thank you for the pointer, I activated some debugging through the `.options` file, hopefully it will yield some useful information:
```properties
#org.eclipse.jdt.internal.core.JavaModelManager.ZIP_ACCESS_WARNING
org.eclipse.jdt.core/debug/zipaccessWarning=true

#org.eclipse.jdt.internal.core.JavaModelManager.ZIP_ACCESS_VERBOSE
org.eclipse.jdt.core/debug/zipaccess=true
```

> One might try/think about caching the result of `Util.getJdkLevel` in `JarPackageFragmentRoot` or analyze why same JAR is opened multiple times.

I'll come back to you in case the traces show something to go on with :-)

[...]

Issue #2323, opened 2025-07-04
Freeze / slow mark occurrences on selection changes in Java editor


iloveeclipse:
I observe significant regression in recent SDK builds while editing Java code.

Selecting a method, variable of field in Java editor is not immediately updating the occurrences int he editor but with a significant delay (5 seconds?).

At same time, sometimes I also see UI monitoring reporting short UI freezes in the code below
```
eclipse.buildId=4.37.0.I20250703-1800
java.version=21.0.6-7
java.vendor=N/A
BootLoader constants: OS=linux, ARCH=x86_64, WS=gtk, NL=en_US
Command-line arguments:  -os linux -ws gtk -arch x86_64 -data /data/4x_platform_workspace

org.eclipse.ui.monitoring
Info
Fri Jul 04 13:43:20 CEST 2025
Sample at 13:43:18.524 (+0.666s)
Thread 'main' tid=1 (TIMED_WAITING)

Stack Trace
	at java.base@21.0.6-7/java.lang.Object.wait0(Native Method)
	at java.base@21.0.6-7/java.lang.Object.wait(Object.java:366)
	at org.eclipse.jdt.core.manipulation.CoreASTProvider.getAST(CoreASTProvider.java:176)
	at org.eclipse.jdt.core.manipulation.SharedASTProviderCore.getAST(SharedASTProviderCore.java:138)
	at org.eclipse.jdt.internal.corext.refactoring.RefactoringAvailabilityTesterCore.isCanonicalConstructor(RefactoringAvailabilityTesterCore.java:313)
	at org.eclipse.jdt.internal.corext.refactoring.RefactoringAvailabilityTester.isCanonicalConstructor(RefactoringAvailabilityTester.java:159)
	at org.eclipse.jdt.internal.corext.refactoring.RefactoringAvailabilityTester.isIntroduceParameterObjectAvailable(RefactoringAvailabilityTester.java:877)
	at org.eclipse.jdt.internal.ui.actions.IntroduceParameterObjectAction.selectionChanged(IntroduceParameterObjectAction.java:74)
	at org.eclipse.jdt.ui.actions.SelectionDispatchAction.dispatchSelectionChanged(SelectionDispatchAction.java:262)
	at org.eclipse.jdt.ui.actions.SelectionDispatchAction.selectionChanged(SelectionDispatchAction.java:257)
	at org.eclipse.jface.viewers.Viewer$1.run(Viewer.java:151)
[...]

PR #2382, opened 2025-07-25
Show progress when calculating AST during copy (Ctrl + c) operation #2028

fedejeanne:
## What it does
If the copy operation requires an AST, then the AST will be calculated by using forking to a non-UI thread and using a `ProgressMonitorDialog`. This gives the user a chance to cancel the operation while also showing the current progress.

Fixes https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2028

## How to test
Sadly I don't know of any clean way to recreate the UI freeze described in #2028 , but one could emulate the behavior with a small hack (see below).

Once the lengthy operation is reproducible (or emulated via the hack), do this:
- Create these 2 classes in 2 different files

```java
import java.io.InputStream;
import java.util.ArrayList;

class A {
	private void a() {
		ArrayList<InputStream> l; // Copy this line
	}
}
```

```java
import java.io.BufferedReader;
import java.util.List;

class B {
	private void b() {
		List<BufferedReader> l2;
		// paste it in this method
	}
	
}
```
- Select and copy (`Ctrl + c`) the line `ArrayList<InputStream> l;` inside `A`. You should see this dialog:

image

- Paste the line in the method `b` in class `B`

**Expected outcome**
If you did **not** cancel the progress when copying the line from `A` then `B` should compile since the imports where obtained and added to the clipboard contents

If you **did** cancel the operation then `B` should not compile since some imports are missing.

### Small hack to help testing
Add this to `org.eclipse.jdt.core.manipulation.CoreASTProvider.getAST(ITypeRoot, WAIT_FLAG, IProgressMonitor)`, which resembles (but it is **not** identical to) the newly added method `waitForLock`:

```java
public CompilationUnit getAST(final ITypeRoot input, WAIT_FLAG waitFlag, IProgressMonitor progressMonitor) {
	if (input == null || waitFlag == null)
		throw new IllegalArgumentException("input or wait flag are null"); //$NON-NLS-1$
	if (progressMonitor != null && progressMonitor.isCanceled())
		return null;

	// Add this
	if (getThreadName().equals("ModalContext")) { //$NON-NLS-1$
		synchronized (fWaitLock) {
			int timeoutInSeconds= 3;
			if (progressMonitor != null)
				progressMonitor.beginTask("Waiting for AST for: " + input.getElementName(), timeoutInSeconds); //$NON-NLS-1$
			for (int i= timeoutInSeconds; i > 0; i--) {
				if (progressMonitor != null && progressMonitor.isCanceled())
					return null; // imports will NOT be copied in the copy operation
				try {
					fWaitLock.wait(1_000);
				} catch (InterruptedException e) { }
				if (progressMonitor != null)
					progressMonitor.worked(1);
			}
		}
	}
	
	// Leave the rest unchanged...
}
```

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.manipulation/common/org/eclipse/jdt/core/manipulation/CoreASTProvider.java
+++ b/org.eclipse.jdt.core.manipulation/common/org/eclipse/jdt/core/manipulation/CoreASTProvider.java
@@ -171,9 +171,11 @@ public final class CoreASTProvider {
 				// Wait for AST
 				synchronized (fWaitLock) {
 					if (isReconciling(input)) {
+						String details= "Waiting for AST for: " + input.getElementName(); //$NON-NLS-1$
 						if (JavaManipulationPlugin.DEBUG_AST_PROVIDER)
-							System.out.println(getThreadName() + " - " + DEBUG_PREFIX + "waiting for AST for: " + input.getElementName()); //$NON-NLS-1$ //$NON-NLS-2$
-						fWaitLock.wait(30000); // XXX: The 30 seconds timeout is an attempt to at least avoid a deadlock. See https://bugs.eclipse.org/366048#c21
+							System.out.println(getThreadName() + " - " + DEBUG_PREFIX + details); //$NON-NLS-1$
+
+						waitForLock(progressMonitor, details);
 					}
 				}
 
@@ -215,6 +217,23 @@ public final class CoreASTProvider {
 			}
 		}
 		return ast;
+	}
+
+	private void waitForLock(IProgressMonitor progressMonitor, String details) throws InterruptedException {
+		// XXX: The 30 seconds timeout is an attempt to at least avoid a deadlock. See https://bugs.eclipse.org/366048#c21
+		int timeoutInSeconds= 30;
+		if (progressMonitor != null)
+			progressMonitor.beginTask(details, timeoutInSeconds);
+
+		for (int i=timeoutInSeconds; i>0; i--) {
+			if (progressMonitor != null && progressMonitor.isCanceled())
+				break;
+
+			fWaitLock.wait(1_000);
+
+			if (progressMonitor != null)
+				progressMonitor.worked(1);
+		}
 	}
 
 	private void notifyReconciler() {
--- a/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/javaeditor/ClipboardOperationAction.java
+++ b/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/javaeditor/ClipboardOperationAction.java
@@ -24,6 +24,7 @@ import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
 import java.util.ResourceBundle;
+import java.util.concurrent.atomic.AtomicReference;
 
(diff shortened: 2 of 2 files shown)


fedejeanne:
Would it be possible to have this one reviewed and merged for M3? 

Would someone kindly take a look? :-)

fedejeanne:
Ping! It would be **highly appreciated** if someone could take a look at this PR. 

The current `master` implementation was responsible for **4+ Hours of UI freezes** in the last month for our developer team. Providing them with a simple fix like this one would save not only time but also really improve the UX of our developers.

Maybe @jjohnstn / @vogella would kindly like to take a look at this PR?

jjohnstn:
Tried it out.  I get a momentary busy cursor when I select the whole line or multiple lines but it is fine.  When I add the delay code, it pops up after one second.  Interestingly, if I try to cancel after 2 of 3 seconds but before it is finished, it copies the imports.  I guess the delay of the dialog is probably in there somewhere.  I think it is ok to merge.

jjohnstn (approved):

laeubi:
@fedejeanne I think this "waiting" code is actually broken, I made an attempt here to improve it but it has not getting much interest, maybe it can give you some help to improve the code:
- https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2324

fedejeanne:
Thank you for the hint @laeubi . I've been reading some of the PRs linked to yours (the ones about mark occurrences and the reconciler running/starting in a Job) and what stuck was your https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2323#issuecomment-3036575541, more precisely:

> So what we see is missing notification and a spurious wakeup of the thread after some seconds...

This seems like the issue I experienced (#2028) were IMO the deadlock is broken by a mere coincidence. Do you think that this is still the case? I mean if I understood correctly, the underlying problem is: one reaches the call to _wait_ but no one is ever going to call _notify_ (which is what the timeout of 30 seconds tried to "fix"), correct?

https://github.com/eclipse-jdt/eclipse.jdt.ui/blob/3a6d9228ceaa45137f34283c663d8ab28ec18605/org.eclipse.jdt.core.manipulation/common/org/eclipse/jdt/core/manipulation/CoreASTProvider.java#L176

In any case, I will roll out the next release to our development team in the coming weeks and see if the situation improves or if we still see lots of UI freezes when copying code. If the UI freezes recede, I'd dare saying that the issue "is solved"...

laeubi:
> This seems like the issue I experienced (https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2028) were IMO the deadlock is broken by a mere coincidence.

I have no clue what kind of "deadlock" the wait wanted to fix, but as described a wait timeout can never fix an deadlock anyways. If Thread A and Thread B are in a deadlock situation, non of the threads can ever issue any wait or notify call.

> I mean if I understood correctly, the underlying problem is: one reaches the call to wait but no one is ever going to call notify (which is what the timeout of 30 seconds tried to "fix"), correct?

So I can only guess that not a _deadlock_ but a _missing notification_ is here "fixed" even though it is simply done wrong, and yes it might work often because the actual thing that is tried to fix do not occur regularly enough to be noticed and is hard to reproduce due to the exact timing.

PR #2392, opened 2025-07-30
Revert "Wait for deferred PDE classpath initializer before initial reconcile"

iloveeclipse:
This reverts commit 1f01b4197f57f636144e5a5059b9af06f14d79e5.

The code in PDE that used ClasspathContainerInitializer.class family is gone, so this code is meaningless now and only increases code complexity.

See https://github.com/eclipse-pde/eclipse.pde/pull/1898


merks:
Thanks for the heads up BTW!

Issue #2403, opened 2025-08-02
JavaLeakTest failures since I20250801-1800


iloveeclipse:
Probably related to https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2382 

```
[...]

PR #2406, opened 2025-08-04
Do not wait for AST when copying code in ClipboarOperationAction

fedejeanne:
## What it does
Does not let the call to `SharedASTProviderCore.getAST(...)` in `ClipboarOperation` wait for the AST of the class. This keeps the call from landing in this line...

https://github.com/eclipse-jdt/eclipse.jdt.ui/blob/3a6d9228ceaa45137f34283c663d8ab28ec18605/org.eclipse.jdt.core.manipulation/common/org/eclipse/jdt/core/manipulation/CoreASTProvider.java#L176

... which is known to cause UI freezes when the call is done in the UI thread (which is the case here, #2028 ).

**TL;DR**
Fixes https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2028 without causing the leak reported in #2403 (I triggered the tests locally by adding the **Extra VM Arguments** [mentioned here](https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2403#issuecomment-3146705909))

## How to test
Copy code from one class to the other (you may use the classes provided in https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2382#issue-3263253258).
If the AST was present at the moment of the copy operation then the imports will be present in the target class. If the AST was **not** ready at the time of the copy operation then the imports will **not** be copied in the target class.

**Disclaimer**
This PR aims to avoid the UI freeze and it does so at the cost of _maybe_ (just _maybe_) not copying the imports when copying code. This is a small tradeoff that I consider acceptable since UI freezes are IMO the absolute worst. I plan to come back to this complex issue (the "deadlock" that happens when waiting in the UI freeze) but for the moment, this PR does the absolute minimum change possible that gets rid of the UI freeze.

It would be nice to get it merged for the next release so (sorry for the spam but) I'll take the liberty of asking for your input directly @jjohnstn , @iloveeclipse and @laeubi .

## Author checklist

- [x] I have thoroughly tested my changes, but manually **and by triggering the `JavaLeakTest` locally**

### DIFF
--- a/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/javaeditor/ClipboardOperationAction.java
+++ b/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/javaeditor/ClipboardOperationAction.java
@@ -484,7 +484,11 @@ public final class ClipboardOperationAction extends TextEditorAction {
 
 
 	private ClipboardData getClipboardData(ITypeRoot inputElement, int offset, int length) {
-		CompilationUnit astRoot= SharedASTProviderCore.getAST(inputElement, SharedASTProviderCore.WAIT_ACTIVE_ONLY, null);
+		// Since this call happens in the UI thread, pass WAIT_NO as parameter
+		// so that the call can not reach the wait call that causes UI freezes.
+		// The cost of doing so is that the method could return null from time
+		// to time and therefore the imports may not be copied to the clipboard
+		CompilationUnit astRoot= SharedASTProviderCore.getAST(inputElement, SharedASTProviderCore.WAIT_NO, null);
 		if (astRoot == null) {
 			return null;
 		}


fedejeanne:
All test failures are unrelated since none of the failing tests even runs into the code that this PR modifies.

fedejeanne:
Any chances of getting this merged by EOD today?

This change is pretty safe and it does prevent UI freezes.

Other methods already take the same approach, _.e.g_ `CleanUpPostSaveListener::getAst`

https://github.com/eclipse-jdt/eclipse.jdt.ui/blob/ef74c4430d96c065c7785c7eb4b6835517a080b2/org.eclipse.jdt.ui/core%20extension/org/eclipse/jdt/internal/corext/fix/CleanUpPostSaveListener.java#L589

jjohnstn:
> Any chances of getting this merged by EOD today?
> 
> This change is pretty safe and it does prevent UI freezes.
> 
> Other methods already take the same approach, _.e.g_ `CleanUpPostSaveListener::getAst`
> 
> https://github.com/eclipse-jdt/eclipse.jdt.ui/blob/ef74c4430d96c065c7785c7eb4b6835517a080b2/org.eclipse.jdt.ui/core%20extension/org/eclipse/jdt/internal/corext/fix/CleanUpPostSaveListener.java#L589

I'm not sure that is an entirely accurate statement.  In the code cited above, it indeed doesn't wait on the getAST call and sees if the AST is available but if it isn't, it parses the unit and creates an AST.  Could that be done in this scenario as well perhaps in a non-UI thread?

fedejeanne:
You're right, I picked a bad example because `CleanupPostSaveListener::createAst` has a fallback mechanism. And yes, that can be done in another thread (see my other PR #2415 ), but this is beside the point.

The point I wanted to make is that the code is ready to handle `null` as the result of calling `SharedASTProviderCore.getAST(...)`. In the case of **this** PR, getting `null` back would mean that the imports didn't get copied to the clipboard and therefore won't be added to the class where you paste the code (you then probably have to manually run "Organize imports" to get them).

Perhaps a better example of not waiting for the AST is in `org.eclipse.jdt.internal.ui.text.java.hover.JavaSourceHover.getBracketHoverInfo(ITextViewer, IRegion)`:

https://github.com/eclipse-jdt/eclipse.jdt.ui/blob/ef74c4430d96c065c7785c7eb4b6835517a080b2/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/text/java/hover/JavaSourceHover.java#L504-L506

In such case, if you get `null` back you simply do **not** display information about the beginning of a block when you hover over a closing bracket (this popup would be missing).

image

Sorry for the confusion.

fedejeanne:
> @fedejeanne I'm ok with this temporary fix. I'll wait a bit to see if there any objections from @iloveeclipse or @laeubi.

Thank you, much appreciated :-)

fedejeanne:
@iloveeclipse I found the culprit: _Semantic Highlighting Reconciler_. See my https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2028#issuecomment-3184139051 . The important part regarding the degradation of the functionality (imports not being copied) is that the imports would be copied once the AST has been calculated.

> This would mean that the copy operation would only get the imports (should there be any necessary) iff the AST for the currently opened editor has already been calculated.

In other words: even after merging this PR, the loss of functionality (missing imports) would only happen while the AST is being calculated. Imaging that the currently opened editor contains a huge class and it takes the highlighting reconciler 2 minutes to get the AST, any copy operation **before** those 2 minutes have elapsed would have no imports (functionality degradation) but every copy operation **after** the 2 minutes have elapsed would have the imports (no degradation).

I hope this level of detail in the analysis is acceptable and the tradeoff is acceptable for you. Let me know if there are any questions :-)

Looking forward to your answer.

iloveeclipse:
> In other words: even after merging this PR, the loss of functionality (missing imports) would only happen while the AST is being calculated. Imaging that the currently opened editor contains a huge class and it takes the highlighting reconciler 2 minutes to get the AST, any copy operation **before** those 2 minutes have elapsed would have no imports (functionality degradation) but every copy operation **after** the 2 minutes have elapsed would have the imports (no degradation).

Sounds reasonable. 
Please rebase on master before merge. I see 21 test fails, this doesn't look nice.

iloveeclipse:
> @iloveeclipse rebase done. Same result.

Could it be, you run into https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2423?
Please double check, if the fails are different, please open new ticket.

PR #2407, opened 2025-08-04
[NewTypeWizardPage] Add support for multiple type creation (#1251)

### LINKED ISSUE #1251: Consider to add support to create multiple types at once through the Dialog Window
Hello 

At the `src/main/java` location for any custom package, when we arrive to the `New Java Class` dialog window we can put the name of the class in the `Name` field. Could you consider to enable to put multiple names to create a set of classes together at once? I mean something like:

```
Car,Computer,Airplane,Book
```

It to avoid open 3 more times the mentioned dialog window. Currently is not possible through neither `,` nor `;`

Consider the same request for Interfaces, Enums and .properties (it at `src/main/resources` and `src/test/resources`)

Of course considering that all these types should be created within the same package.

Thanks for your understanding.

alexcoover:
## What it does

Adds support for creating multiple types at once through the New Type dialog (#1251).

Multiple type names can be entered, separated by semicolons (with or without spaces).

Only the first file is opened in the editor.

Duplicate names are ignored.

## How to test

Open the dialog for creating a new class, enum, or other types.

In the type name field, enter multiple names separated by semicolons.

Closes #1251


iloveeclipse:
- Please squash 34 commits to one
- Rebase this ome on master
- Never do merge commits again
- Provide appropriate commit message

alexcoover:
Is there anything else I need to do? @iloveeclipse

PR #2422, opened 2025-08-09
Support configuration of the release for a source folder - alternative icon

stephan-herrmann:
Here's my proposal for #2260: 

jar_mr_obj@2x


laeubi:
Looks good to me

PR #2425, opened 2025-08-11
Don't log InterruptedExceptions in CompletionProposalComputerDescriptor

fedejeanne:
## What it does
Avoid logging a `RuntimeException` that wraps an `InterruptedException`. These are thrown way down in the `Parser` if the user cancels (clicks outside of) the completion proposal popup.

## Requires
- https://github.com/eclipse-platform/eclipse.platform.ui/pull/3157
- https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4302

## How to test
Once this PR and the requires PRs are merged, it can be tested as specified in https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4302


jjohnstn (approved):
LGTM

PR #2432, opened 2025-08-14
Stop using jface.text.tests DisplayHelper

akurtakov:
Use the JDT testplugin one.
JFace.text.tests one is to be removed.


akurtakov:
Thanks Jeff!

PR #2436, opened 2025-08-14
Add logic to OrganizeImports to support module imports

### LINKED ISSUE #2421: Organize imports for Java 24 replaces existing import module that covers types used
Applying organize imports to the following code using Java 24 and preview option on:

```
package test24;

import module java.base;

public class Blah {
	BigDecimal x = new BigDecimal(4);
}

```

results in replacing the import module statement with an import of java.math.BigDecimal.  The operation should leave the existing import module.

jjohnstn:
- add new ImportOrganizeTest24
- fixes #2421

## What it does

See commit comment.

## How to test

See new test.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.manipulation/common/org/eclipse/jdt/core/manipulation/OrganizeImportsOperation.java
+++ b/org.eclipse.jdt.core.manipulation/common/org/eclipse/jdt/core/manipulation/OrganizeImportsOperation.java
@@ -39,6 +39,7 @@ import org.eclipse.jdt.core.Flags;
 import org.eclipse.jdt.core.ICompilationUnit;
 import org.eclipse.jdt.core.IJavaElement;
 import org.eclipse.jdt.core.IJavaProject;
+import org.eclipse.jdt.core.IModuleDescription;
 import org.eclipse.jdt.core.IPackageFragment;
 import org.eclipse.jdt.core.IPackageFragmentRoot;
 import org.eclipse.jdt.core.ISourceRange;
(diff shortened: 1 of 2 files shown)


stephan-herrmann:
> @stephan-herrmann This is the JDT UI part that requires the ImportRewrite patch. I have a test but I added it for Java 24 with preview. I was able to generate an RT_STUBS24 library. I don't know what the protocol is but the test could be modified post Java 25 to add an RT_STUBS25 library and make it Java 25.

Now looking at this for end-to-end testing.

Unfortunately in BETA_JAVA25 the compiler no longer accepts 24 with preview flag, which causes all new tests to fail. So please update to 25 now. And since no other test required the setup for 24, you could directly modify all 24-additions to say 25 (without preview option in this case).

org.eclipse.jdt.core.manipulation/common/org/eclipse/jdt/core/manipulation/OrganizeImportsOperation.java
```diff
@@ -323,6 +370,27 @@ public void add(SimpleName ref) {
 				}
 				typeBinding= typeBinding.getTypeDeclaration();
 				if (!typeBinding.isRecovered()) {
+					String qualifiedTypeName= typeBinding.getQualifiedName();
+					if (qualifiedTypeName.indexOf('.') > 0) {
+						String qualifier= qualifiedTypeName.substring(0, qualifiedTypeName.lastIndexOf('.'));
+						if (moduleImportsContains(qualifier)) {
+							for (Entry<String, Set<String>> entry : fOldModuleImports.entrySet()) {
+								if (entry.getValue().contains(qualifier)) {
```
stephan-herrmann, line 378: Just a random observation: the outer if seems redundant, given that we are doing the same iteration with contains check again.

stephan-herrmann:
For testing along with https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4313 I created a bunch of modular projects: [testmodules.zip](https://github.com/user-attachments/files/22147231/testmodules.zip)

When invoking organize imports in main.Client, actually two imports should be removed:
* `d.D` is obviously covered by `module d` -- this is correctly removed
* `a.c.C` is covered by `module a` due to its `requires transitive a.c` -- this import is not getting removed

Note, that `a.b.B` is still necessary because that requires is not transitive.

I didn't check whether the problem is in jdt.core or ui.

stephan-herrmann:
> I have verified the test works as desired.

I can confirm, thanks.

Playing around with the new Organize Imports led me to file https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4374

Feel free to comment.

PR #2448, opened 2025-08-22
Sort Profile settings by key when exporting

FlorianKroiss:
## What it does
When the exporting a profile, e.g. for cleanup or formatting, the order of the key/value pairs may be undefined, as they come from a HashMap. This change sorts the entries by key before creating the XML Profile Element.

## How to test
1. Edit default Cleanup Profile
2. Give it a different name
3. Export to XML file
4. The settings in the XML file should be sorted by key

## Author checklist


FlorianKroiss:
Without this change it's hard to see which settings were changed when looking at a git diff of an exported Profile

PR #2460, opened 2025-09-01
New Class dialog: fix cropped Interfaces field on mixed-DPI monitors

ShahzaibIbrahim:
### What it does

The Interfaces field in the New Class dialog was clipped when moving dialog between monitors with different scaling factors (100–300% DPI). This was caused by GridData.heightHint being set to convertHeightInCharsToPixels(3), which can resolve too small after a DPI switch and result in the text field and its adjacent button being cut off.

This change increases the height hint from 3 to 4 character rows when fTypeKind == CLASS_TYPE, ensuring the control renders correctly across all tested DPI scales.

## How to test

- Run the runtime workspace with following arguments
  ```
  -Dswt.autoScale=quarter
  -Dswt.autoScale.updateOnRuntime=true
  ```
- Primary monitor at 100% zoom, secondary monitor at 200% zoom (scaling)
- Start Eclipse from the file explorer on the 200% monitor.
- Open the "New >>Class" dialog.
- Move the dialog to the 100% monitor.

### Results
image

## Author checklist

### DIFF
--- a/org.eclipse.jdt.ui/ui/org/eclipse/jdt/ui/wizards/NewTypeWizardPage.java
+++ b/org.eclipse.jdt.ui/ui/org/eclipse/jdt/ui/wizards/NewTypeWizardPage.java
@@ -1177,7 +1177,7 @@ public abstract class NewTypeWizardPage extends NewContainerWizardPage {
 		});
 		GridData gd= (GridData) fSuperInterfacesDialogField.getListControl(null).getLayoutData();
 		if (fTypeKind == CLASS_TYPE) {
-			gd.heightHint= convertHeightInCharsToPixels(3);
+			gd.heightHint= convertHeightInCharsToPixels(4);
 		} else {
 			gd.heightHint= convertHeightInCharsToPixels(6);
 		}


ShahzaibIbrahim:
> I’ve verified that this change resolves the reported issue. That said, it essentially replaces one magic number with another, so it’s difficult to review the code. I don’t know why the height hint was set to 3 originally, nor why it now needs to be 4.
> 
> Just noting an observation, I noticed that the position of the “Remove” button shifts relative to the interfaces table when moving the dialog across monitors. I believe this was already the case since the height was always defined in pixels. If I skip setting the size hint entirely, the button’s position seems to remain consistent with the textbox.

With heightHint you provide the preferred height in points. If you don't provide it then you hand the control to windows and it decides whatever the height it wants to set, which relatively looks fine for this case if we don't set it. My understanding of setting some value here is about we want to control the size ourself and not rely on windows. Value 3 seemed to work when we had only 100% monitor and now with different sizes (as we tested), the value 4 seem to work well.

jjohnstn (approved):
LGTM

PR #2471, opened 2025-09-06
fold markdown Javadoc comments with extended folding

### LINKED ISSUE #2442: Extended folding does not allow folding of markdown comments
```java
    /// comments
    /// go
    /// here
    public MyClass() {
        super();
    }
```
When enabling Extended Folding, the markdown comments can't be folded. With it disabled, folding works as expected.

Version: 2025-09 (4.37)
Build id: I20250806-1800

danthe1st:
Fixes #2442

## What it does
Extended folding currently doesn't fold [Markdown Javadoc comments](https://openjdk.org/jeps/467) in places where it would fold regular block comments. This PR fixes that.

I also used this opportunity to remove `JavaProjectHelper.delete(jProject);` from the test because `ProjectTestSetup` already does that.

## How to test

- Create a Java project using JDK 23 or later
- Enable extended folding in Window > Preferences > Java > Editor > Folding
- Add the following class:
```java
/// This is
/// a class
/// using markdown comments
public class MyClass {
    /// This is
    /// a method
    /// using markdown comments
    void x(){}
}
```
- The Javadoc comments should be folded

## Author checklist

### DIFF
--- a/org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/folding/MarkdownJavadocFoldingTest.java
+++ b/org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/folding/MarkdownJavadocFoldingTest.java
@@ -1,0 +1,116 @@
+/*******************************************************************************
+ * Copyright (c) 2025 Daniel Schmid and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Daniel Schmid - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.jdt.text.tests.folding;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import java.util.List;
+
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import org.junit.runners.Parameterized.Parameters;
+
+import org.eclipse.jdt.testplugin.JavaProjectHelper;
+
+import org.eclipse.core.runtime.CoreException;
+
+import org.eclipse.jface.preference.IPreferenceStore;
+
+import org.eclipse.jface.text.IRegion;
+
+import org.eclipse.jdt.core.IJavaProject;
+import org.eclipse.jdt.core.IPackageFragment;
+import org.eclipse.jdt.core.IPackageFragmentRoot;
+
+import org.eclipse.jdt.ui.PreferenceConstants;
+import org.eclipse.jdt.ui.tests.core.rules.Java23ProjectTestSetup;
+import org.eclipse.jdt.ui.tests.core.rules.ProjectTestSetup;
+
+import org.eclipse.jdt.internal.ui.JavaPlugin;
+
+@RunWith(Parameterized.class)
+public class MarkdownJavadocFoldingTest {
+	@Rule
+	public ProjectTestSetup projectSetup = new Java23ProjectTestSetup(false);
+
+	private IJavaProject fJProject1;
+
+	private IPackageFragmentRoot fSourceFolder;
+
+	private IPackageFragment fPackageFragment;
+
+	private final boolean extendedFoldingActive;
+
+	public MarkdownJavadocFoldingTest(boolean extendedFoldingActive) {
+		this.extendedFoldingActive= extendedFoldingActive;
+	}
+
+	@Parameters(name= "Extended folding active: {0}")
+	public static Object[] parameters() {
+		return new Object[] { true, false };
+	}
+
+	@Before
+	public void setUp() throws CoreException {
+		fJProject1= projectSetup.getProject();
+		fSourceFolder= JavaProjectHelper.addSourceContainer(fJProject1, "src");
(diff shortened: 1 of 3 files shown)


carstenartur:
Usually tests make use of a Rule declaring the right project test configuration at the start of the test file. So you can reuse test configurations without repeating all the test setup for different java versions.

danthe1st:
It would be possible to do that, I'm not sure whether it's worth it hence I'm not committing this yet.

Since I only need the JDK compilance and not a runtime with the JDK, that rule is sufficient:
```java
public class JavaNCompilanceProjectTestSetup extends ProjectTestSetup {

	private final int javaVersion;

	public JavaNCompilanceProjectTestSetup(int javaVersion) {
		super("TestSetupProjectN", JavaProjectHelper.RT_STUBS_15);
		this.javaVersion= javaVersion;
	}

	@Override
	protected IJavaProject createAndInitializeProject() throws CoreException {
		IJavaProject javaProject= super.createAndInitializeProject();

		Map<String,String> options= javaProject.getOptions(false);
		JavaCore.setComplianceOptions(String.valueOf(javaVersion), options);
		javaProject.setOptions(options);

		return javaProject;
	}
}
```
However, I don't think using that makes the test simpler in any way:
```java
@RunWith(Parameterized.class)
public class MarkdownJavadocFoldingTest {
	@Rule
	public ProjectTestSetup projectSetup;

	private IJavaProject fJProject1;

	private IPackageFragmentRoot fSourceFolder;

	private IPackageFragment fPackageFragment;

	private final boolean extendedFoldingActive;
	private final boolean hasMarkdownJavadocs;

	public MarkdownJavadocFoldingTest(boolean extendedFoldingActive, int jdkCompilance) {
		this.extendedFoldingActive= extendedFoldingActive;
		this.projectSetup= new JavaNCompilanceProjectTestSetup(jdkCompilance);
		hasMarkdownJavadocs = jdkCompilance >= 23;

	}

	@Parameters(name= "Extended folding active: {0}, JDK compilance: {1}")
	public static List<Object[]> parameters() {
		return List.of(
				new Object[] { true, 22 },
				new Object[] { true, 23 },
				new Object[] { false, 22 },
				new Object[] { false, 23 });
	}

	@Before
	public void setUp() throws CoreException {
		fJProject1= projectSetup.getProject();
		fSourceFolder= JavaProjectHelper.addSourceContainer(fJProject1, "src");
		fPackageFragment= fSourceFolder.createPackageFragment("org.example.test", false, null);
		IPreferenceStore store= JavaPlugin.getDefault().getPreferenceStore();
		store.setValue(PreferenceConstants.EDITOR_NEW_FOLDING_ENABLED, extendedFoldingActive);
	}

	@After
	public void tearDown() {
		IPreferenceStore store= JavaPlugin.getDefault().getPreferenceStore();
		store.setToDefault(PreferenceConstants.EDITOR_NEW_FOLDING_ENABLED);
	}

	@Test
	public void testClassWithMarkdownJavadocAsHeaderComment() throws Exception {
		String str= """
				package org.example.test;
				/// Javadoc							//here should be an annotation
				/// comment
				/// here
				public class HeaderCommentTest {
				}
				""";
		if (hasMarkdownJavadocs) {
			FoldingTestUtils.assertCodeHasRegions(fPackageFragment, "TestFolding.java", str, 1);
			List<IRegion> regions= FoldingTestUtils.getProjectionRangesOfFile(fPackageFragment, "TestFolding.java", str);
			FoldingTestUtils.assertContainsRegionUsingStartAndEndLine(regions, str, 1, 3); // Javadoc
		} else {
			FoldingTestUtils.assertCodeHasRegions(fPackageFragment, "TestFolding.java", str, 0);
		}
	}

	@Test
	public void testSingleMethodWithMarkdownJavadoc() throws Exception {
		String str= """
				package org.example.test;
				public class SingleMethodTest {
				    /// Javadoc							//here should be an annotation
				    /// comment
				    /// here
				    public void foo() {					//here should be an annotation
				        System.out.println("Hello");
				    }
				}
				""";
		List<IRegion> regions= FoldingTestUtils.getProjectionRangesOfFile(fPackageFragment, "TestFolding.java", str);
		if (hasMarkdownJavadocs) {
			assertEquals(2, regions.size());
			FoldingTestUtils.assertContainsRegionUsingStartAndEndLine(regions, str, 2, 4); // Javadoc
		} else {
			assertEquals(1, regions.size());
		}
		FoldingTestUtils.assertContainsRegionUsingStartAndEndLine(regions, str, 5, 6); // foo method
	}
}
```

Alternatively, it would be possible to create two classes for these tests - but I don't feel like there's much of a point in doing that.

carstenartur:
We have that already 
https://github.com/eclipse-jdt/eclipse.jdt.ui/tree/master/org.eclipse.jdt.ui.tests/ui/org/eclipse/jdt/ui/tests/core/rules

danthe1st:
> We have that already

Thanks, I didn't really notice that ones for JDK 22 and 23 already existed. I updated my PR accordingly.

org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/folding/MarkdownJavadocFoldingTest.java
```diff
@@ -0,0 +1,124 @@
...
+		this.extendedFoldingActive= extendedFoldingActive;
+		this.hasMarkdownJavadocs= hasMarkdownJavadocs;
+		if (hasMarkdownJavadocs) {
+			projectSetup= new Java23ProjectTestSetup(false);
+		} else {
+			projectSetup= new Java22ProjectTestSetup();
+		}
+	}
+
+	@Parameters(name= "Extended folding active: {0}, JDK compilance with markdown Javadocs: {1}")
```
jjohnstn: What is the purpose of the first parameter?  The tests don't seem to do anything different based on whether the preference is set or not.

PR #2410, opened 2025-08-05
preserve contextual information with enhanced folding

danthe1st:
## What it does
Currently, the "enhanced folding" does not add information like the Java element the folding regions are associated with to them. Because of this, it is possible that Eclipse confuses/swaps two folding regions when editing the file which resets their expanded/collapsed state (because it considers them moved or doesn't find the folding regions any more).

The information of which element a folding region is associated to is used [in `update()` to correlate "old" folding regions with "new" folding regions](https://github.com/eclipse-jdt/eclipse.jdt.ui/blob/master/org.eclipse.jdt.ui/ui/org/eclipse/jdt/ui/text/folding/DefaultJavaFoldingStructureProvider.java#L1317).

With #2302 (that's how I noticed it), the issue solved by this PR causes can also cause parts outside of the visible region to become visible hence this PR is a precondition to make #2302 work properly.

## How to test

- Enable "Enhanced Folding" and "Custom Folding regions" (with the default region markers) in Preferences > Java Editor > Folding
- Create the following class:
```java
public class TestClass {

	// region

	void test() {

	}

	//endregion

}
```
- Collapse the `test` method
- Edit the file by adding a line break somewhere (or perform a similar edit)
- Without this PR, the whole `//region` folding region is collapsed. With it, the `test()` method stays collapsed and the region stays expanded.

## Author checklist

- [X] I have thoroughly tested my changes
- [X] The change is following the [coding conventions](https://wiki.eclipse.org/Platform/How_to_Contribute#Coding_Conventions)
- [X] I have signed the [Eclipse Contributor Agreement (ECA)](https://www.eclipse.org/legal/ECA.php)

## Other notes

If wanted, I can make the integration test more extensive like this:
```java
	@Test
	public void testFoldingUpdateWithMultipleCustomRegionsDoesNotSwitchRegions() throws Exception {
		String code= """
				package org.example.test;

				// region outer

				public class Test {
					// region middle
					void someMethod() {
						boolean b = true;
						// region inner
						if (b) {
							// content
						}
						// endregion
						for(;;) {
							// something else
						}
					}
					// endregion
				}

				// endregion outer
				""";
		ICompilationUnit cu= fPackageFragment.createCompilationUnit("Test.java", code, true, null);
		JavaEditor editor= (JavaEditor) EditorUtility.openInEditor(cu);
		try {
			ProjectionAnnotationModel model= editor.getAdapter(ProjectionAnnotationModel.class);

[...]

PR #2488, opened 2025-09-13
sort folding regions to avoid confusing different custom regions (prevent random collapsing/expanding of custom folding regions when editing a file)

danthe1st:
## What it does

(This is the technical description of why and how, for a quick introduction of what this fixes, view "How to test")

If extended folding is enabled, custom folding regions (if enabled) are not necessarily created in the order they are located in the file (because the file is processed sequentially and the regions can be created only once a `// endregion` (or similar) comment has been found).

To keep collapsed folding collapsed and expanded regions expanded, the folding logic [tries to match the old folding regions with the new ones](https://github.com/eclipse-jdt/eclipse.jdt.ui/blob/6ae4ee7f4f9cb880f70c42e392865ee04b656a66/org.eclipse.jdt.ui/ui/org/eclipse/jdt/ui/text/folding/DefaultJavaFoldingStructureProvider.java#L1349-L1369). For this, it uses a combination of information about which `IJavaElement`s are associated to the folding regions, whether the folding regions correspond to comments or not and (if it is still ambiguous) the order of folding regions.

If two custom folding regions are associated with the same `IJavaElement`, it can incorrectly associate them hence it will collapse a previously expanded region when editing a file (or vice-versa) because they are in the wrong order.

This PR fixes that issue for custom folding regions by sorting them by their offset. The existing (old) folding regions are already sorted [here](https://github.com/eclipse-jdt/eclipse.jdt.ui/blob/6ae4ee7f4f9cb880f70c42e392865ee04b656a66/org.eclipse.jdt.ui/ui/org/eclipse/jdt/ui/text/folding/DefaultJavaFoldingStructureProvider.java#L2302). Sorting folding regions (once more) shouldn't be much an issue when it comes to performance because Java uses TimSort which is pretty optimized in general and O(n) if it's already sorted (and to confirm this, I checked with VisualVM and didn't see the sorting show up when profiling - neither with nor without extended folding, see below).

This is very similar to #2410 and this PR would actually fix the reproducer from #2410 but there are cases where both changes are necessary.

For example, this PR fixes the issue in "How to test"

On the other hand, it's also possible to construct a case where this PR isn't sufficient and #2410 is still necessary. For example, consider the following class:
```java
public class SomeClass {
    void a(){
         // ...
    }

    void b(){
         // ...
    }
}
```

If
- extended folding is active
- `a()` is collapsed while `b()` isn't
- the `a()` method is removed

then `b()` shouldn't be collapsed. This is an example where #2410 is still necessary.

#2410 enabled the use of `IJavaElement`s (which are also used for the "old" Java folding) with extended folding for properly detecting the `IJavaElement`s. On the other hand, this PR improves the situation for multiple custom folding regions associated to the same `IJavaElement`.

### is there anything left

Even with both of these fixes, there are some possibilities where folding regions are confused. This may happen if custom folding regions are added or removed.

For example, imagine the following class:
```java
public class Hello1234 {
	// region abc
	
	// endregion abc
	
	// region def
	
	// endregion def
	
	// region gih
	
	// endregion gih
}
```
If the `def` region is collapsed and the `abc` region is removed (e.g. by deleting the `// region abc` line), all custom regions get expanded.

## How to test
- Enable extended folding
- Create the following class:
```java
public class SimpleNestedFolding {
	// region outer

	//region inner

	// endregion

	//endregion
}
```
- Collapse the inner regions
- Edit the file
- observe the outer region being collapsed

I have also done some checks with VisualVM. I created a file with 12k inner classes (or technically I reused it) and edited it while profiling `org.eclipse.jdt.ui.text.folding.**` with VisualVM.

With extended folding, I got the following results:
image

And these are my results with the "old" folding:
image

As you can see, the sorting isn't relevant for performance here, it doesn't even appear in the profiling output (it's part of the "Self time" in `update()`).

## Author checklist

### DIFF
--- a/org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/folding/CustomFoldingRegionTest.java
+++ b/org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/folding/CustomFoldingRegionTest.java
@@ -691,10 +691,13 @@ public class CustomFoldingRegionTest {
 				// region outer
 
 				public class Test {
+					// region middle
+
 					// region inner
 					void someMethod() {
 						// content
 					}
+					// endregion
 					// endregion
 				}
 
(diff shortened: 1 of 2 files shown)


jjohnstn (approved):
Looks good.

PR #2502, opened 2025-09-19
Add ambiguous type checking for import module

### LINKED ISSUE #2490: [25] Organize Imports with multiple import module statements does not check for ambiguous classes
For the following modules a, a.b, and a.c:

```
module a {
	requires a.b;
	requires transitive a.c;
	exports a;
}
```

```
module a.b {
	exports a.b;
}
```

```
module a.c {
	exports a.c;
}
```

where there an a.b.C and a.c.C class, the following code:

```
package main;

import module a;
import module a.b;
import a.c.C;

public class Example {
	A a;
	B b;
	C c;
}
```

which has the following module-info.java:

```
module main {
	requires a;
	requires a.b;
}
```

when Organize Imports CTRL+SHFT+O is entered, will remove the import a.c.C statement which is needed because without it, there is an ambiguous reference to C (there a.b.C and a.c.C).

jjohnstn:
- add new logic to OrganizeImportsOperation to look for the case where a simple type name is found in multiple import modules and the user has added an explicit import statement which should not be removed
- add proper ambiguous test in ImportRewrite.processTypeInfo for modules
- add new test to ImportOrganizeTest25 for preserving an import that is otherwise ambiguous
- fixes #2490

## What it does

See issue.

## How to test

See issue or new test.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.manipulation/common/org/eclipse/jdt/core/manipulation/OrganizeImportsOperation.java
+++ b/org.eclipse.jdt.core.manipulation/common/org/eclipse/jdt/core/manipulation/OrganizeImportsOperation.java
@@ -367,7 +367,7 @@ public class OrganizeImportsOperation implements IWorkspaceRunnable {
 								fImpStructure.addModuleImport(entry.getKey(), new ArrayList<>(entry.getValue()));
 								fModuleImportsAdded.add(entry.getKey());
 							}
-							if (fOldModuleImports.size() > 0 && typeNameAmbiguousForImportedModules(typeName)) {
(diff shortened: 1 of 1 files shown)


jjohnstn:
This requires: https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4415 merged to I-builds

stephan-herrmann:
> ... the case where a simple type name is found in multiple import modules and the user has added an explicit import statement which should not be removed

Does the fix also address the similar situation where ambiguity is produced just by one module import?
```java
import module java.se;
import java.awt.List;

...
List l = ...
```
When the explicit import is removed, ambiguity exists between `java.awt.List` and `java.util.List`.

Issue #2505, opened 2025-09-20
"Open Type" dialog suggest wrong type when using fully qualified name


ptziegler:
This is a regression introduced with https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2420. When searching for a type using its fully qualified name, it might not be the first result in the list.

Before:

<img width="596" height="501" alt="Image" src="https://github.com/user-attachments/assets/b7f52212-dc3d-4a76-96d2-c6cfa7be3581" />

After:

<img width="596" height="501" alt="Image" src="https://github.com/user-attachments/assets/a5cf391e-e8de-40eb-ab6e-2393e6f91005" />

I would expect an exact match to always be the first suggestion.

ptziegler:
@jjohnstn Based on https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2387#issuecomment-3275313754 it sounds like you're aware of the ordering issue?

This new behavior is causing a failure in our automatic tests because now the wrong type is selected.

ptziegler:
I've created https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2506.

PR #2506, opened 2025-09-20
Make '*' injection in "Open Type" dialog more restrictive #2505

ptziegler:
## What it does

This adapts the modification of the pattern text, that was introduced with 02958ce874164f2a61fcfba52e975988f3704e89 to only add '*' between two camel-case words. Otherwise the search might return wrong or unexpected results when e.g. searching for fully-qualified types.

## How to test

1) Search for "java.lang.String". The Java class should be the first match, as opposed to e.g. "java.lang.AbstractStringBuilder", which is suggested when searching for "java.lang.*String".

2) Search for "ComAct", which should return the same results as "Com*Act". More specifically, it should also include types such as the "ComboBoxActionPropertyChangeListener", which was previously excluded because the "Box" was not part of the search query.

Closes https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2505

## Author checklist

### DIFF
--- a/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/dialogs/FilteredTypesSelectionDialog.java
+++ b/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/dialogs/FilteredTypesSelectionDialog.java
@@ -446,16 +446,20 @@ public class FilteredTypesSelectionDialog extends FilteredItemsSelectionDialog i
 		String text= super.getPatternText();
 		fTypeItemsComparator.setOriginalPattern(text);
 		StringBuilder builder= new StringBuilder();
-		boolean lastCharIsUpperCase= true;
+		boolean canAddAnyStringNext= true;
 		for (int i= 0; i < text.length(); ++i) {
 			char ch= text.charAt(i);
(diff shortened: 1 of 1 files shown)


org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/dialogs/FilteredTypesSelectionDialog.java
```diff
@@ -446,18 +446,22 @@ protected String getPatternText() {
...
-		boolean lastCharIsUpperCase= true;
+		boolean canAddAnyStringNext= true;
 		for (int i= 0; i < text.length(); ++i) {
 			char ch= text.charAt(i);
-			if (!lastCharIsUpperCase && Character.isUpperCase(ch)) {
+			if (canAddAnyStringNext && Character.isUpperCase(ch)) {
 				builder.append('*');
 			}
 			builder.append(ch);
-			lastCharIsUpperCase= Character.isUpperCase(ch);
```
ptziegler, line 456: There is also an edge-case that was missed in the previous implementation.  If the original pattern is e.g. `A*B`, then this method would convert it to `A**B`, which is excessive.

PR #2535, opened 2025-10-10
Add org.eclipse.jdt.junit6.runtime

merks:
- It's a copy of org.eclipse.jdt.junit5.runtime but with everything renamed for junit 6 includes, packages, classes, the bundle name, and references to them.
- This bundle requires Java 17.


HannesWell:
> We might want to mention this maybe on `platform-dev` or `eclipse-dev`? Since anyone running plug-in tests will have to restrict their JUnit 5 version to actually use 5 and not accidentally use 6.

I would be in favour of making the error check and it's message very clear (and ideally still simple) so that also those users that don't follow the mailing lists are capable to fix that problem properly, as discussed in:
- https://github.com/eclipse-pde/eclipse.pde/issues/2045

> And I'm not sure if we need to add anything in PDE to support JUnit 6... Probably at least the quick-fix for "Add JUnit 5 to plug-in requirements (e.g. on hover of `@Test` that cannot be resolved) needs the equivalent for JUnit 6.

In general each occurance of `Junit5` should be checked if it needs a junit6 pendant:
- https://github.com/search?q=repo%3Aeclipse-pde%2Feclipse.pde+junit5&type=code

Some might not need adjustments, but I assume many do.

> It should be fine to merge this, or are there objections? @merks , @laeubi , @HannesWell ? Technically the new plug-in is not yet used anywhere, so this PR is harmless. But the next one will be adding JUnit 6 to the SDK, with this change:

Please not that non of us there has power here.
Therefore I ask the JDT committers here to judge:
@iloveeclipse, @akurtakov, @jjohnstn or @noopur2507 please judge or mention those that can judge here.

iloveeclipse:
One the build is green, I plan to merge & start new SDK build, as otherwise the code does not compile in the IDE without JUnit 6 in the platform.

iloveeclipse:
> please not merge this now! 

OK

> We need some more steps and possibly can use a simpler approach here!

Could you provide more details which additional steps needed?

trancexpress:
IMO its time we merge this.

@merks , @HannesWell , @laeubi any objections?

After that we can add the dependency to it: https://github.com/eclipse-jdt/eclipse.jdt/pull/145

trancexpress:
> so I wonder how anyone ever would have been able to test the new runtime here...

Let me check this then.

When I import the new plug-in, there are 2 problems:

```
Unsatisfied version constraint: 'junit-jupiter-migrationsupport: [6.0.0,7.0.0)'	MANIFEST.MF	/org.eclipse.jdt.junit6.runtime/META-INF	line 13	Plug-in Problem
Unsatisfied version constraint: 'junit-vintage-engine: [6.0.0,7.0.0)'	MANIFEST.MF	/org.eclipse.jdt.junit6.runtime/META-INF	line 15	Plug-in Problem
```

Do we add those to the JDT feature requirements @laeubi? Currently those are:

```
   <requires>
      <import plugin="junit-jupiter-api" version="6" match="compatible"/>
      <import plugin="junit-jupiter-engine" version="6" match="compatible"/>
      <import plugin="junit-jupiter-params" version="6" match="compatible"/>
      <import plugin="junit-platform-commons" version="6" match="compatible"/>
      <import plugin="junit-platform-engine" version="6" match="compatible"/>
      <import plugin="junit-platform-launcher" version="6" match="compatible"/>
      <import plugin="junit-platform-suite-api" version="6" match="compatible"/>
      <import plugin="junit-platform-suite-engine" version="6" match="compatible"/>
   </requires>
```

trancexpress:
Any objections to merging this?

iloveeclipse (approved):
LGTM

iloveeclipse:
The code compilation with Java 17 showed one new warning which I will fix now.

Issue #2540, opened 2025-10-15
Sticky Lines hang forever on opening DirectByteBuffer


iloveeclipse:
Using latest I20251014-1810 build on Java 25.
I've tried to open type "DirectByteBuffer" from java 25 and the UI hangs with 100% CPU load for munites now. I have to kill it.

Stack :
```
"main" #3 [3831173] prio=6 os_prio=0 cpu=558814,23ms elapsed=7180,99s tid=0x00007f9538030980 nid=3831173 runnable  [0x00007f953f1fc000]
   java.lang.Thread.State: RUNNABLE
	at org.eclipse.jdt.core.dom.ASTNode.accept(ASTNode.java:3310)
	at org.eclipse.jdt.core.dom.ASTNode.acceptChildren(ASTNode.java:3383)
	at org.eclipse.jdt.core.dom.TypeDeclaration.accept0(TypeDeclaration.java:497)
	at org.eclipse.jdt.core.dom.ASTNode.accept(ASTNode.java:3312)
	at org.eclipse.jdt.core.dom.ASTNode.acceptChildren(ASTNode.java:3383)
	at org.eclipse.jdt.core.dom.CompilationUnit.accept0(CompilationUnit.java:257)
	at org.eclipse.jdt.core.dom.ASTNode.accept(ASTNode.java:3312)
	at org.eclipse.jdt.core.dom.NodeFinder.<init>(NodeFinder.java:196)
	at org.eclipse.jdt.internal.ui.javaeditor.JavaStickyLinesProvider.getASTNode(JavaStickyLinesProvider.java:327)
	at org.eclipse.jdt.internal.ui.javaeditor.JavaStickyLinesProvider.getStickyLines(JavaStickyLinesProvider.java:300)
	at org.eclipse.ui.internal.texteditor.stickyscroll.StickyScrollingHandler.calculateAndShowStickyLines(StickyScrollingHandler.java:153)
	at org.eclipse.ui.internal.texteditor.stickyscroll.StickyScrollingHandler$$Lambda/0x0000000090c70000.run(Unknown Source)
	at org.eclipse.jface.util.Throttler.lambda$0(Throttler.java:66)
	at org.eclipse.jface.util.Throttler$$Lambda/0x0000000090523c00.run(Unknown Source)
	at org.eclipse.jface.util.Throttler.lambda$1(Throttler.java:79)
[...]

Issue #2533, opened 2025-10-09
MarkdownCommentTests.testSpec03Tags fails since I20251007-1800


iloveeclipse:
See for example https://download.eclipse.org/eclipse/downloads/drops4/I20251007-1800/testresults/html/org.eclipse.jdt.ui.tests_ep438I-unit-linux-x86_64-java21_linux.gtk.x86_64_21.html

```
expected: <<p>super doc In addition, this methods calls <code><a href='eclipse-javadoc:%E2%98%82=TestSetupProject/src%3Cp%7BSpec03Tags.java%E2%98%83Spec03Tags~m~I%E2%98%82%E2%98%82wait%E2%98%82'>wait()</a></code>.<div><b>Overrides:</b> <a href='eclipse-javadoc:%E2%98%82=TestSetupProject/src%3Cp%7BSpec03Tags.java%E2%98%83Super~m~I'>m(...)</a> in <a href='eclipse-javadoc:%E2%98%82=TestSetupProject/src%3Cp%7BSpec03Tags.java%E2%98%83Super'>Super</a></div></p> <dl><dt>Parameters:</dt><dd><b>i</b> the index</dd></dl> > but was: <<p>super doc} In addition, this methods calls <code><a href='eclipse-javadoc:%E2%98%82=TestSetupProject/src%3Cp%7BSpec03Tags.java%E2%98%83Spec03Tags~m~I%E2%98%82%E2%98%82wait%E2%98%82'>wait()</a></code>.<div><b>Overrides:</b> <a href='eclipse-javadoc:%E2%98%82=TestSetupProject/src%3Cp%7BSpec03Tags.java%E2%98%83Super~m~I'>m(...)</a> in <a href='eclipse-javadoc:%E2%98%82=TestSetupProject/src%3Cp%7BSpec03Tags.java%E2%98%83Super'>Super</a></div></p> <dl><dt>Parameters:</dt><dd><b>i</b> the index</dd></dl> >

org.opentest4j.AssertionFailedError: expected: <<p>super doc
[...]

PR #2541, opened 2025-10-15
Break the loop in getStickyLines() if same AST node was found

iloveeclipse:
Fixes https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2540


iloveeclipse:
Test failure is unrelated, see https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2533

Issue #1830, opened 2024-12-04
Slow (Quick-)Type hierarchy for large trees.


jukzi:
Simple reproducer:
```
class LargeTree {

	interface I0 {}
	interface I1 extends I0 {}
	interface I2 extends I1, I0 {}
	interface I3 extends I2, I1 {}
	interface I4 extends I3, I2 {}
	interface I5 extends I4, I3 {}
	interface I6 extends I5, I4 {}
	interface I7 extends I6, I5 {}
	interface I8 extends I7, I6 {}
	interface I9 extends I8, I7 {}
	interface I10 extends I9, I8 {}
	interface I11 extends I10, I9 {}
	interface I12 extends I11, I10 {}
	interface I13 extends I12, I11 {}
	interface I14 extends I13, I12 {}
	interface I15 extends I14, I13 {}
	interface I16 extends I15, I14 {}
	interface I17 extends I16, I15 {}
	interface I18 extends I17, I16 {}
	interface I19 extends I18, I17 {}
	interface I20 extends I19, I18 {}
//	interface I21 extends I20, I19 {}
//	interface I22 extends I21, I20 {}
//	interface I23 extends I22, I21 {}
//	interface I24 extends I23, I22 {}
//	interface I25 extends I24, I23 {}
//	interface I26 extends I25, I24 {}
//	interface I27 extends I26, I25 {}
//	interface I28 extends I27, I26 {}
//	interface I29 extends I28, I27 {}
//	interface I30 extends I29, I28 {}
}

```
![image](https://github.com/user-attachments/assets/45c6054d-6d47-4b00-b5ff-11275d373a5a)

On windows ~ 12 seconds to crtl-t `I20`.  Seems to be a non ending story for already I30 - O(2^n) ?

![image](https://github.com/user-attachments/assets/4aa09ac7-b12e-4bd2-918d-8d24701e00bd)
![image](https://github.com/user-attachments/assets/8f110e5e-a5ff-465d-8884-2fb1d64f26df)

My Vote: by default JDT should not try to expand more then constant number of  nodes.
Probably Initial maximum number of elements would be a usefull limit:
![image](https://github.com/user-attachments/assets/9cdff5e7-4e02-4148-9547-31d8729c037d)

jukzi:
@raghucssit WDYT?

jukzi:
Or just do not auto-expand nodes that are already expanded elsewhere in the tree? @fedejeanne

[...]

PR #2550, opened 2025-10-22
Limit initial quick type hierarchy view expansion #2549

### LINKED ISSUE #2549: Slow opening of quick type hierarchy view for lots of classes
In our project (about ~2000 plugins and features) we have very large and deep inheritance hierarchies. When opening the quick type hierarchy view (`ctrl+t`) on some methods, it takes a very long time to open the view and Eclipse ui freezes while the view is being opened.

See this sample of opening a method in our project, about 120s to render the view on `4.38.0.v20251020-1800` on a windows machine:

Image

This issue boils down to the fact that the view expands an extremely large inheritance tree (about 17000 classes) when opening.
The fix from #1954 for related issue #1830 does not apply to this situation because there are no duplicate subtrees involved.

SeineEloquenz:
## What it does

This PR fixes #2549.
When opening the quick type hierarchy view, there is no limit on the amount of items expanded initially, leading to possibly long loading times and ui freezes.
With this change, the amount of items in the tree expanded initially is limited to the `largeViewLimit` setting.

For an especially bad example in our code base (see #2549) this change reduces the waiting time for the view dramatically, while not impacting use cases with a more _reasonable_ amount of classes (less than the `largeViewLimit`):
image

## How to test

Sadly I could not find a simple reproducer that does not contain our product code or is a big zip bundle of thousands of classes.
The issue occurs if there is a very large amount of classes in the type hierarchy view that do not get mitigated by the change from #1830.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/typehierarchy/HierarchyInformationControl.java
+++ b/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/typehierarchy/HierarchyInformationControl.java
@@ -134,6 +134,7 @@ public class HierarchyInformationControl extends AbstractInformationControl {
 		tree.setLayoutData(gd);
 
 		TreeViewer treeViewer= new TreeViewer(tree) {
+			private static final int EXPANSION_LIMIT = 10;
 			private Set<Object> visited;
 
 			@Override
@@ -153,7 +154,7 @@ public class HierarchyInformationControl extends AbstractInformationControl {
 			}
 
 			private boolean shouldExpand(Widget widget) {
-				if (widget == null) {
+				if (widget == null || visited.size() > EXPANSION_LIMIT) {
 					return false;
 				}
 				Object data= widget.getData();


iloveeclipse:
What is the status of this PR, why it is in a draft state? Something missing?

The idea to limit the view is OK, assuming it works (I haven't tested) :-)

Ideally you could provide steps to reproduce, and of course you can set the viewer limit preference to "5" to see the effect. 

I assume askung for type hierarchy of `Runnable` could be one of the candidates? It is slow but dosn't freeze anything for me on Linux.

iloveeclipse:
Why not just do that:
```
diff --git a/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/typehierarchy/HierarchyInformationControl.java b/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/typehierarchy/HierarchyInformationControl.java
index c9247b7..76018fc 100644
--- a/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/typehierarchy/HierarchyInformationControl.java
+++ b/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/typehierarchy/HierarchyInformationControl.java
@@ -37,2 +37,4 @@
 
+import org.eclipse.ui.views.WorkbenchViewerSetup;
+
 import org.eclipse.jdt.core.ICompilationUnit;
@@ -162,2 +164,3 @@
 		};
+		WorkbenchViewerSetup.setupViewer(treeViewer);
 		treeViewer.addFilter(new ViewerFilter() {
```

SeineEloquenz:
I agree now that the `WorkbenchViewerSetup#getItemsLimit` is not the correct value for this use-case.
Hardcoding a value like `10` (or anything else `< 1000` really) would certainly work for our situation.

Regarding the auto expand level; I do not know why `ALL_LEVELS` was used in the first place (maybe no-one thought there'd be such large trees that it would become a problem?).
For our codebase, however, we would need a limit less than `8` because the hierarchy is not deep but rather extremely wide. I'm skeptical whether that would be a good value to impose on all eclipse users.

Would it be OK for you if I update the PR with a hard-coded limit of `10` instead of `WorkbenchViewerSetup#getItemsLimit`?

PR #2552, opened 2025-10-22
[Performance] Do not calculate the folding structure for uninteresting events

fedejeanne:
> [!IMPORTANT]
> This issue is especially relevant when using the **new** folding mechanism introduced in 4.36 ([N&N](https://eclipse.dev/eclipse/markdown/?f=news/4.36/jdt.md)) because that one requires an AST and this can be very expensive for large classes. Avoiding some of the unnecessary calculations reduces the risk of UI freezes.
 
## What it does
Avoids recalculating the folding structures for changes in preferences that either
a) Do not actually affect folding or
b) The preferences haven't changed since the last time they were set.

> [!NOTE]
> w.r.t. point _a)_: activating/deactivating the breadcrumbs is easily done (no need to open the preferences) and is one such case.

> [!NOTE]
> w.r.t. point _b)_: this happens when the project has the _project specific settings_ active and the **General** preferences are changed. In such case, the general preferences trigger an event and the method `DefaultJavaFoldingStructureProvider::initializePreferences` pulls the "new" preferences from the project-specific settings, which haven't changed at all.

## How to test
- Create 2 Java projects and add some classes to them
- Activate the _Project specific settings_ for one of the projects under `Preferences > Java Editor > Folding`
- Open a class in one project
- Change the folding preferences of the other project

**Corner case**
Open one class of each project side-by-side and play around with the preferences.

**Expected result**
If you are currently looking at a Java File that is not tied to the preference you changed (either because the preference is not related to _folding_ or because you changed the _General_ preferences and the project has _Project specific settings_ activated) then the method `initialize()` inside the listener should never be reached.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.ui/ui/org/eclipse/jdt/ui/text/folding/DefaultJavaFoldingStructureProvider.java
+++ b/org.eclipse.jdt.ui/ui/org/eclipse/jdt/ui/text/folding/DefaultJavaFoldingStructureProvider.java
@@ -26,6 +26,7 @@ import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 
 import org.eclipse.core.runtime.Assert;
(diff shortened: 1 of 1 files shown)


fedejeanne:
I fixed the issue with the project-specific settings. This PR is ready to be reviewed.

jjohnstn (approved):
A sensible optimization.

fedejeanne:
Failing test is unrelated: #2533

PR #2558, opened 2025-10-24
Add support for selecting JUnit 6 in the UI

laeubi:
Currently the user can not select JUnit 6 as a version in the classpath

This enhances the wizard so the user can choose.

grafik

FYI @merks @akurtakov @iloveeclipse @trancexpress


trancexpress:
Sorry @laeubi I got swamped with other things at work.

PR #2569, opened 2025-10-26
Include last line during folding

### LINKED ISSUE #2439: Folding misbehave since 2025-03
In earlier version, folding a method or class included the last line containing the closing '}', but since 2025-03 it doesn't, the last line remains visible with the first line. Is this intentional ? And if yes I'm begging you to reverse it.

### LINKED ISSUE #2596: "old" folding cannot expand annotated enum constants with more elements
When creating an enum constant with an annotation in another line (and another enum constant later), a folding region is created (assuming the "old" non-extended folding).
When this folding region is collapsed, its icon disappears and it is no longer possible to expand it.

### Steps to reproduce
- Disable Extended folding in Window > Preferences > Java > Editor > Folding but keep folding enabled
- Create an enum like this:
```java
enum Test {
	@Deprecated
	A,
	B
}
```
- The `A` enum constant can be collapsed.
Image
- Upon collapsing it, the indicator for the folding region disappears and it can no longer be unfolded without "Collapse All".
Image

This also happens if the enum constant has multiple annotations.

danthe1st:
Fixes #2439
Fixes #2596

## What it does
When extended folding was introduced, both the existing and extended folding mechanisms were changed to generally not include the last line (with a few exceptions).

If I remember correctly, this was done to prevent hiding the `b()` method in situations like this:
```java
public class Test {
	void a() {
		
	}void b() {
		
	}
}
```
This PR fixes this regression by including the last line in the folding regions where appropriate. If there is anything additional in the same line as the end of the folding region (e.g. the `b()` method above), the last line won't be included.

This change applies to both the "old" and "extended" folding mechanisms but the extended folding will **not** include the last line in the folding regions in control structures other than lambdas (unless someone tells me when exactly these should be included).

## How to test
- Make sure folding is enabled
- Create a Java class like the one from https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2439#issuecomment-3304535151
- If the preference Java > Editor > Folding > Include last line when folding Java elements is enabled, the `}` should be hidden
- This should work with both the "old" and extended folding.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/folding/CustomFoldingRegionTest.java
+++ b/org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/folding/CustomFoldingRegionTest.java
@@ -171,7 +171,7 @@ public class CustomFoldingRegionTest {
 				""";
 		List<IRegion> projectionRanges= getProjectionRangesOfFile(str);
 		assertEquals(1, projectionRanges.size());
-		FoldingTestUtils.assertContainsRegionUsingStartAndEndLine(projectionRanges, str, 2, 3);
+		FoldingTestUtils.assertContainsRegionUsingStartAndEndLine(projectionRanges, str, 2, 4);
 	}
 
 	@Test
@@ -188,7 +188,7 @@ public class CustomFoldingRegionTest {
 				""";
 		List<IRegion> projectionRanges= getProjectionRangesOfFile(str);
 		assertEquals(2, projectionRanges.size());
-		FoldingTestUtils.assertContainsRegionUsingStartAndEndLine(projectionRanges, str, 2, 5);
+		FoldingTestUtils.assertContainsRegionUsingStartAndEndLine(projectionRanges, str, 2, 6);
 		FoldingTestUtils.assertContainsRegionUsingStartAndEndLine(projectionRanges, str, 3, 5);
 	}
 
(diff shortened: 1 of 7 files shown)


fedejeanne:
Looks good, I tested only with the snippet provided in https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2439#issuecomment-3304535151 and I like what I see. I would get rid of the parameter though, it's unnecessary complexity.

Thank you, @danthe1st !

danthe1st:
>  In my testing I also found some other issues (regressions that I think came from the introduction of extended folding) but I think these can be worked on independently.

Nevermind, that was #2596 and I was able to fix it by by treating commas in the same way as semicolons so it makes more sense to do it in this PR.

However, one thing that still needs to be done is adapting all the tests since they expect the last line to be excluded.
image

fedejeanne:
Thank you for looking into this, @danthe1st 💪

danthe1st:
Note the following (change of behavior with extended folding/folding of control structures): My current approach (not fully committed yet) also folds the last line with some extended folding control structures where there is something after the region ends:
```java
try { // not collapsed (start)
    ... // collapsed, region end
} catch(SomeException e) { // not collapsed by first region, second region starts here
    // collapsed
} // folding region end
```

```java
do { // not collapsed (start)
    ... // collapsed, region end
} while(...); // Not collapsed
```

For methods and other control structures, the last line is included:
```java
void something() { // not collapsed, region start
    ... // collapsed
} // collapsed, region end
```
```java
while(...) { // not collapsed, region start
    ... // collapsed
} // collapsed, region end
```

Furthermore, anything that consists of only two lines is now folded:
```java
void a() { // folding region here
} // this would be collapsed now because it is a region consisting of 2 lines
```

Also, should `break;` (and `yield`?) statements be included in the folding of `case` blocks or should it stop before? Currently, these are excluded. For consistency, I guess it makes sense to also collapse these?

Also, for things like `while` and `for`, it is based on indentation of the next line (there is an explicit check for that (`FoldingVisitor#createFoldingRegionForStatement`) - why???).
See this:
```java
    while(...) {
        ...
    } // this is not collapsed
    // indentation ends here
    while(...) {
        ...
    } // this is collapsed???
        // indentation ends here
```
Should I remove that behavior as well (it _should_ be as simple as just removing the whole `if (!(node instanceof Block)) {`)?

Note: With my (current) approach, comments would prevent the last line from being folded.

PR #2578, opened 2025-10-28
Adjust highlighting of deprecation

stephan-herrmann:
After https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4564 deprecation no longer propagates from a type to its members.

Still highlight as deprecated constructors of anonymous classes and default constructors.

See https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4553#issuecomment-3448344936


stephan-herrmann:
@srikanth-sankaran I wonder if record constructors need special treatment, too? Here it's about their visualization as deprecated (strike-through), but then I'm not sure if those are actually correctly wired into deprecation warnings in the first place?

stephan-herrmann:
> @srikanth-sankaran I wonder if record constructors need special treatment, too? Here it's about their visualization as deprecated (strike-through), but then I'm not sure if those are actually correctly wired into deprecation warnings in the first place?

Compiler-wise things look good. I have some tests to be pushed soon.

In the editor I see strike through missing in line 5 here:
image

All else looks good to me.

PR #2597, opened 2025-11-02
QuickFix : Navigate to Existing Variable Declaration from duplicate entry

SougandhS:
This PR adds a navigation link in the Quick Fix hover and dialog to navigate to the first declaration of a variable when a duplicate is detected. Users can click the link to jump directly to the original declaration, making it easier to understand the variable’s context and reducing time spent searching through code.

image

https://github.com/user-attachments/assets/c5d166bc-688f-4d08-b1c2-94ff8f785432

NB : label was updated

## Author checklist


SougandhS:
Test failure was expected 👍 
I can update the tests accordingly.. WDYT @noopur2507 ?

SougandhS:
Updated label text : 
image

noopur2507 (changes requested):
The link here is called a "Quick Fix". Please make the above changes to the implementation and update the PR title as well.

SougandhS:
> I tried the PR today. The functionality is the same as the "Open Declaration" option that we get when we press Cmd and hover on an element. So you can use the same name in the quick fix.
> 
> I tried a few scenarios with this example:
> 
> ```
> package pkg;
> 
> public class ZZ {
> 	public static boolean foo(Integer n) {
> 		boolean b = switch (n) {
> 		case Integer x when x.equals(10) -> {
> 			int n; // error
> 			yield true;
> 		}
> 		case Comparable<?> y -> {
> 			int y, b; // errors
> 			yield false;
> 		}
> 		};
> 		return b;
> 	}
> 
> 	public static void main(String[] argv) {
> 		System.out.println(foo(Integer.valueOf(0)));
> 		System.out.println(foo(Integer.valueOf(10)));
> 	}
> }
> ```
> 
> * Using this new quick fix on "n" selects "Integer n", i.e. it selects the type also.
> * Using this new quick fix on "b" selects the entire variable declaration fragment (the entire initializer).
> * No quick fix on the error at "y".
> 
> It should select only the variable declaration name. You can refer to the implementation of "Open Declaration" option and reuse the same for this quick fix if possible.

Thanks.. will check this scenarios

PR #2638, opened 2025-11-12
Fix inconsistent project settings regarding java version

laeubi:
The org.eclipse.jdt.junit.runtime currently declares a BREE with Java-11 and therefore also gets this as a JRE11 on the classpath, but in the build.properties it declares jre.compilation.profile = JavaSE-1.8 and should actually work to run tests on older JREs as well. Furthermore it has previously declared source/compliance/target level of 1.8 already so it actually was targeted at 1.8

To prevent confusion in an OSGi environment (where it will only resolve for Java 11+) this now fixes the manifest and classpath to align with that as well.

This is a regression from 8fc47ee24e6af110bc194b291586674df7378ac5 where it was claimed to bum the bundle to java 11 but it was not and likely should not unless we support java 1.8 in JDT.

FYI @trancexpress @iloveeclipse


laeubi:
Fixed junit4/5 runtime as well, junit6 looks fine.

PR #2653, opened 2025-11-20
Making the dialog area of NewTypeWizard pages scrollable

arunjose696:
When the wizard dialog is manually resized , text on NewTypeWizard pages can become partially cut off without any visual indication. This update makes the page area scrollable, so any content that would otherwise be cut off remains accessible to the user through scrollbars.

Steps to reproduce

1)Open a new class dialog(or any other new type dialogs) in eclipse IDE
2)Resize it to make it smaller to see the scrollbars appear in the dialog area with this change


jjohnstn:
Hi @arunjose696 We are just in 4.38 RC1 and heading into 4.38 RC2, then 4.38 release, so this will have to wait until 4.39 M1.  It is policy that you should not merge eclipse-jdt into your branch, rather you should always rebase on eclipse-jdt:master.  The update button in github offers to merge and no one seems to be able to turn off the merge option, but you can use the pull down to select the 2nd option which is update with rebase.

Issue #2667, opened 2025-11-26
Avoid parsing jar name when determining JUnit 5 or 6 version


trancexpress:
See: #2665

Right now we are ignoring the parsing exception. We should do something better for Eclipse 4.39 - e.g. check for a `MANIFEST.MF` in the jar, then check the version and symbolic name from it.

iloveeclipse:
Good idea

laeubi:
I won't use any names at all and also OSGi info's are more optional than mandatory and might change, but I'm not deep into this code and what it try to archive...

For the formal part see here so java even has an API for that:

- https://docs.oracle.com/javase/tutorial/deployment/jar/packageman.html 
- https://docs.oracle.com/javase/8/docs/technotes/guides/versioning/spec/versioning2.html#wp89936

and JUnit follow the specification with the following headers in its jar:

```
Implementation-Title: junit-jupiter-api
Implementation-Vendor: junit.org
Implementation-Version: 6.0.1
Specification-Title: junit-jupiter-api
Specification-Vendor: junit.org
Specification-Version: 6.0.1
```

so for the JDT part I would lean to use this API over any guessing of jar/class/osgi names...

trancexpress:
There are some internal methods we could use from JDT core:

```
					try {
						PackageFragmentRoot root= (PackageFragmentRoot) type.getAncestor(IJavaElement.PACKAGE_FRAGMENT_ROOT);
						Manifest manifest= root.getManifest();
						if (manifest != null) {
							Attributes attributes= manifest.getMainAttributes();
							String versionString= attributes.getValue("Specification-Version"); //$NON-NLS-1$
							if (versionString != null) {
								Version version= Version.parseVersion(versionString);
								if (version.getMajor() != junitMajorVersion) {
									return false;
								}
							}
						}
					} catch (Throwable e) {
						JUnitCorePlugin.log(Status.warning("Failed to determine JUnit version", e)); //$NON-NLS-1$
					}
```

Unfortunately it requires a JDT core manifest change, so not something for 4.38:

```
diff --git a/org.eclipse.jdt.core/META-INF/MANIFEST.MF b/org.eclipse.jdt.core/META-INF/MANIFEST.MF
index f63fc9f5bf..a73db01de8 100644
--- a/org.eclipse.jdt.core/META-INF/MANIFEST.MF
+++ b/org.eclipse.jdt.core/META-INF/MANIFEST.MF
[...]

PR #2671, opened 2025-11-28
Parse jar manifest when trying to determine JUnit 5 or 6 version

trancexpress:
Fixes: #2667

## Author checklist

### DIFF
--- a/org.eclipse.jdt.junit.core/META-INF/MANIFEST.MF
+++ b/org.eclipse.jdt.junit.core/META-INF/MANIFEST.MF
@@ -3,7 +3,7 @@ Automatic-Module-Name: org.eclipse.jdt.junit.core
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.jdt.junit.core;singleton:=true
-Bundle-Version: 3.14.0.qualifier
+Bundle-Version: 3.14.100.qualifier
 Bundle-Activator: org.eclipse.jdt.internal.junit.JUnitCorePlugin
 Bundle-ActivationPolicy: lazy
 Bundle-Vendor: %providerName
--- a/org.eclipse.jdt.ui.tests/META-INF/MANIFEST.MF
+++ b/org.eclipse.jdt.ui.tests/META-INF/MANIFEST.MF
@@ -3,7 +3,7 @@ Automatic-Module-Name: org.eclipse.jdt.ui.tests
 Bundle-ManifestVersion: 2
 Bundle-Name: %Plugin.name
 Bundle-SymbolicName: org.eclipse.jdt.ui.tests; singleton:=true
-Bundle-Version: 3.16.300.qualifier
+Bundle-Version: 3.16.400.qualifier
 Bundle-Activator: org.eclipse.jdt.testplugin.JavaTestPlugin
 Bundle-Vendor: %Plugin.providerName
 Bundle-Localization: plugin
--- a/org.eclipse.jdt.ui.tests/pom.xml
+++ b/org.eclipse.jdt.ui.tests/pom.xml
@@ -20,7 +20,7 @@
   </parent>
   <groupId>org.eclipse.jdt</groupId>
   <artifactId>org.eclipse.jdt.ui.tests</artifactId>
-  <version>3.16.300-SNAPSHOT</version>
+  <version>3.16.400-SNAPSHOT</version>
   <packaging>eclipse-test-plugin</packaging>
   <properties>
     <defaultSigning-excludeInnerJars>true</defaultSigning-excludeInnerJars>


org.eclipse.jdt.junit.core/src/org/eclipse/jdt/internal/junit/util/CoreTestSearchEngine.java
```diff
@@ -173,19 +176,39 @@ private static boolean hasJUnitJupiterTestAnnotation(IJavaProject project, int j
...
-							JUnitCorePlugin.log(Status.error("Failed to parse determine JUnit version: " + filename, e)); //$NON-NLS-1$
+							/*
+							 * Some JUnit distributions use different naming, ignore those. See: https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2665
+							 * We will try parsing a jar manifest below.
+							 */
 						}
 					}
+					if (isJar && version == null) {
+						IPath typePathWithoutProject= type.getPath().removeFirstSegments(1);
+						IPath absolutePath = project.getProject().getLocation().append(typePathWithoutProject);
```
iloveeclipse: Sure that the jar is always located inside the project and not part of the dependency chain (and so located somewhere else or even links to external location)?

Javadoc says

> If this element is not included in an external library, the path returned is the full, absolute path to the underlying resource, relative to the workbench. If this element is included in an external library, the path returned is the absolute path to the archive or to the folder in the file system. 

I would first ask `IWorkspaceRoot.getFileForLocation(IPath)` and if there is nothing, would check if `Files.get(type.getPath().toOSString())` exists

trancexpress:
> I would move detection into the catch block. I see if there is a launch config, we check ~10 times per config in the dialog.

Maybe its validation once per tab? E.g.:

```
checking for: /data/workspaces/runtime-Eclipse/junitest3/lib/junit-platform-commons.jar
	at org.eclipse.jdt.internal.junit.util.CoreTestSearchEngine.getVersion(CoreTestSearchEngine.java:337)
	at org.eclipse.jdt.internal.junit.util.CoreTestSearchEngine.hasJUnitJupiterTestAnnotation(CoreTestSearchEngine.java:183)
	at org.eclipse.jdt.internal.junit.util.CoreTestSearchEngine.hasJUnit5TestAnnotation(CoreTestSearchEngine.java:163)
	at org.eclipse.jdt.junit.launcher.JUnitLaunchConfigurationTab.validateJavaProject(JUnitLaunchConfigurationTab.java:1019)
	at org.eclipse.jdt.junit.launcher.JUnitLaunchConfigurationTab.validatePage(JUnitLaunchConfigurationTab.java:971)
	at org.eclipse.jdt.junit.launcher.JUnitLaunchConfigurationTab.isValid(JUnitLaunchConfigurationTab.java:905)
	at org.eclipse.debug.internal.ui.launchConfigurations.LaunchConfigurationTabGroupViewer.refresh(LaunchConfigurationTabGroupViewer.java:552)
	at org.eclipse.debug.internal.ui.launchConfigurations.LaunchConfigurationsDialog.updateButtons(LaunchConfigurationsDialog.java:1648)
	at org.eclipse.debug.ui.AbstractLaunchConfigurationTab.updateLaunchConfigurationDialog(AbstractLaunchConfigurationTab.java:155)
	at org.eclipse.debug.ui.AbstractLaunchConfigurationTab$1.runInUIThread(AbstractLaunchConfigurationTab.java:513)
	at org.eclipse.ui.progress.UIJob.lambda$0(UIJob.java:148)
	at org.eclipse.swt.widgets.RunnableLock.run(RunnableLock.java:40)
	at org.eclipse.swt.widgets.Synchronizer.runAsyncMessages(Synchronizer.java:132)
	at org.eclipse.swt.widgets.Display.runAsyncMessages(Display.java:5084)
	at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:4549)
	at org.eclipse.jface.window.Window.runEventLoop(Window.java:824)
	at org.eclipse.jface.window.Window.open(Window.java:804)
	at org.eclipse.debug.internal.ui.launchConfigurations.LaunchConfigurationsDialog.open(LaunchConfigurationsDialog.java:1238)
	at org.eclipse.debug.ui.DebugUITools.lambda$1(DebugUITools.java:631)
	at org.eclipse.swt.custom.BusyIndicator.showWhile(BusyIndicator.java:67)
	at org.eclipse.debug.ui.DebugUITools.openLaunchConfigurationDialogOnGroup(DebugUITools.java:637)
	at org.eclipse.debug.ui.DebugUITools.openLaunchConfigurationDialogOnGroup(DebugUITools.java:575)
	at org.eclipse.debug.ui.actions.OpenLaunchDialogAction.run(OpenLaunchDialogAction.java:85)
```

trancexpress:
> I would move detection into the catch block. I see if there is a launch config, we check ~10 times per config in the dialog.

Thinking about this, does that change anything?

Lets assume we have `junit-platform-commons.jar` on the classpath. In that case, we don't parse anything, as we don't match `junit-platform-commons-*.jar` and `junit-platform-commons_*.jar`. We determine the wrong type of launch (like in our case, version-less names for JUnit 5 jars).

So yes, we spare effort, but we launch the wrong thing (e.g. in our case).

Lets assume we change the matching to `junit-platform-commons*.jar`. We fail the version parsing, so we are in the exception block. We determine the right version from the manifest - but still we do this many times. If the jar with the annotation has some other name (e.g. `junit-commons.jar`) we launch with the wrong type - we save effort but the launch is wrong. If the user wants to fix this, we tell the user to rename their jar.

And then lets assume we keep the change as is - we handle cases where the jar containing the annotation has any name.

Worst case scenario, some network file on the classpath is now slow to work with... On the first launch.

Other than some cache, I don't see how we can correctly determine the JUnit version here without depending on jar names, which will in all likelihood be odd for users.

trancexpress:
Hi @merks ,

really sorry to bother you.

But it seems we underestimated #2667. In our case we have symlinks to JUnit libraries, the versions are removed from the symlink names. This is so that we don't have to update `.classpath` files on version updates.

As a result, we launch with JUnit 6 and run into exceptions, since our projects are on JUnit 5 still (but don't use the standard JDT JUnit 5 container, they refer each necessary JUnit library).

Can we have this change in a 4.38RC2a build?

It might well be that the problem in #2667 is something users will run into and not rarely, as I assumed. 

We ran late with our testing of 4.38 as an IDE for our developers and so found the problem only today.

Best regards,
Simeon

trancexpress:
> As mentioned on the issue, I would rather use
> 
> ```
> Specification-Vendor: junit.org
> Specification-Version: 6.0.1
> ```
> 
> from the manifest and do not look at filenames at all.

IMO the filename checks will avoid some zip I/O.

Is your concern with wrongly named libraries?

merks:
For what it's worth, I trust your judgement and give my PMC +1 for the change and an RC2a.

@HannesWell 

Can you please advise how best to ensure this ends up in an 4.38 build?

trancexpress:
> > Can you please advise how best to ensure this ends up in an 4.38 build?
> 
> We are currently in the process of preparing the next dev-cycle.
> 
> Please await unitl that effort is completed. Then you can submit this to master and then back-port this to the 4.38 maintenance branch. And from the maintanance branch we can then start a new I-build and that can then be promoted to RC2a. But I suggest to wait for a few days if there is more content for an RC2a.

OK, enough time to examine the proposal from @laeubi .

org.eclipse.jdt.junit.core/src/org/eclipse/jdt/internal/junit/util/CoreTestSearchEngine.java
```diff
@@ -165,25 +169,36 @@ public static boolean hasJUnit6TestAnnotation(IJavaProject project) {
...
 				if (type != null) {
 					// check if we have the right JUnit JUpiter version
 					String filename= type.getPath().lastSegment();
-					if ((filename.startsWith(junitBundlePrefix + "_") || filename.startsWith(junitBundlePrefix + "-")) && filename.endsWith(JAR_EXTENSION)) { //$NON-NLS-1$ //$NON-NLS-2$
-						String versionString = filename.substring(junitBundlePrefix.length() + 1, filename.length() - JAR_EXTENSION.length());
-						try {
-							Version version = Version.parseVersion(versionString);
-							if (version.getMajor() != junitMajorVersion) {
-								return false;
+					if (filename.endsWith(JAR_EXTENSION)) {
```
iloveeclipse: Why don't we check first for the "trivial" jar name match now? It would avoid IO for probably most users.

trancexpress:
Well, the impact is in worst case running a test and nothing happens. The user then has to guess that they must change the JUnit launch type from 6 to 5.

I'm guessing we'll have other things to fix too, early in 4.39 cycle. Unfortunately we were able to start testing only late last week.

If we are not merging this for 4.38, then please don't merge the current state. I'd like to make somewhat bigger changes, see: https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2667#issuecomment-3590493124

iloveeclipse (approved):

Issue #2693, opened 2025-12-11
[Javadoc] New JDT.Core Behavior for Curly Brackets in Multiline @code Tags Triggers JDT.UI Regression


subyssurendran666:
### Brief history about the Issue

There were some Javadoc parsing optimisations done in [JDT.Core](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4626) to handle the `@code` and `@literal` tags, which caused a regression in [JDT.UI](https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2685). Therefore, a new PR has been proposed in [JDT.Core](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4686), and a corresponding PR will be proposed for JDT.UI.

### What is the route cause

https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4626 enables the `@code` tag to support multiple curly brackets (`{}`) and multiline content. The older implementation did not support multiline usage and failed when multiple curly brackets were present.
<img width="1954" height="828" alt="Image" src="https://github.com/user-attachments/assets/d0f0da54-1699-4cfc-9973-33ddd3ca075d" />
Here, the entire `@code` tag is not treated as a single TagElement. This was the old behaviour.

===========================================

<img width="1898" height="838" alt="Image" src="https://github.com/user-attachments/assets/c31aa5b0-a191-47f0-97d4-3d77ce66c2ee" />
In the new behaviour, the entire `@code` tag is treated as a single TagElement. However, I suspect there is an issue in this implementation. The opening and closing curly brackets inside the `@code` tag are excluded from the AST view.

Therefore, the new approach cause regression in the [JDT.UI](https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2685). So, temporarily , the tests were disabled [via](https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2689).

PR #2694, opened 2025-12-11
Update JDT.UI to support curly brackets in multiline tags from JDT.Core

subyssurendran666:
Aligns JDT.UI with the updated JDT.Core behavior where multiline @code tags may contain curly brackets, ensuring the AST/UI rendering remains consistent.

Fix: https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2693

NB: This PR can be merged only after the corresponding PR in [JDT.core](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4686)

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.manipulation/common/org/eclipse/jdt/core/manipulation/internal/javadoc/CoreJavadocAccessImpl.java
+++ b/org.eclipse.jdt.core.manipulation/common/org/eclipse/jdt/core/manipulation/internal/javadoc/CoreJavadocAccessImpl.java
@@ -978,11 +978,14 @@ public class CoreJavadocAccessImpl implements IJavadocAccess {
 		if (isCode || (isLink && addCodeTagOnLink())) {
 			if (isCode) {
 				ASTNode sibling = getNextSiblingElement((TagElement)node.getParent(), node);
-				if (sibling != null)
+				if (sibling != null) {
 					fBuf.append("\n </code>"); //$NON-NLS-1$
-			}
-				else
+				} else {
 					fBuf.append("</code>"); //$NON-NLS-1$
+				}
+			} else {
+				fBuf.append("</code>"); //$NON-NLS-1$
+			}
 		} else
 		if (isSnippet)
 			fBuf.append("</code></pre>"); //$NON-NLS-1$


org.eclipse.jdt.core.manipulation/common/org/eclipse/jdt/core/manipulation/internal/javadoc/CoreJavadocAccessImpl.java
```diff
@@ -975,15 +975,31 @@ else if (isSnippet) {
 
 		if (isReturn)
 			fBuf.append(JavaDocMessages.JavadocContentAccess2_returns_post);
-		if (isCode || (isLink && addCodeTagOnLink()))
-			fBuf.append("</code>"); //$NON-NLS-1$
+		if (isCode || (isLink && addCodeTagOnLink())) {
+			if (isCode) {
+				ASTNode sibling = getNextSiblingElement((TagElement)node.getParent(), node);
```
jjohnstn: Do you have a test scenario where getNextSibling is null and so you don't add the closing </code> tag?  It should be added. 

FWIW: I did work on pre code tag support and nested inline tags in an inline return tag.

Minor nit: the code indentation is off.

subyssurendran666: I do not have a valid test scenario where `getNextSibling` is null, but I can mimic one.

```
/**
	 * Performs:
	 * <pre>{@code
	 *    for (String s : strings) {
	 *        if (s.equals(value)) {
	 *            return \u0030;
	 *        }
	 *        if (s.startsWith(value)) {
	 *            return 1;
	 *        }
	 *    }
	 *    return -1;
	 * }
	 */
```
Here, I removed the closing `pre` tag. In this case, getNextSibling becomes null.

You are right, it is better to add an `else` for the `if (sibling != null)` condition

jjohnstn: I tried a test with no <pre> tags and it works as expected.

Issue #2696, opened 2025-12-12
After termination of suspended JUnit test "Update JUnit" job starts every second


iloveeclipse:
JUnit view has a bug, it *seems* that it was always there, haven't verified.

1) Put a breakpoint at the first line of any test method 
2) Start JUnit test
3) If debugger stops at the breakpoint, "stop" JUnit test execution from JUnit view
4) Resume debugger 
5) **JUnit 5:** After launch terminates there will be "Update JUnit" job remaining, which runs every 200ms till next execution will start. JUnit view shows "Stopping..." status forever. **Works on JUnit 6**.

Variation 1:

4) Terminate debugger
5) **Both JUnit 5 & 6:** after launch terminates there will be "Update JUnit" job remaining, which runs every 200ms till next execution will start. JUnit view shows "Stopping..." status forever.

Variation 2:
1) Start JUnit test
2) Terminate launch from debugger
3) **Both JUnit 5 & 6:** after launch terminates there will be "Update JUnit" job remaining, which runs every 200ms till next execution will start. JUnit view shows "Test XYT..." status. After "Stopping" the test from JUnit view the view shows "Stopping..." status forever.

The "Update JUnit" job is supposed only to run while JUnit test is being executed

For validation, I've used the test case below:
```
import org.junit.jupiter.api.Assumptions;
import org.junit.jupiter.api.Test;

class TestCancellation {

	@Test
	void test1() {
//		Assumptions.assumeTrue(false);
		System.out.println("1");
		waitAMoment();
	}
	@Test
	void test2() {
		System.out.println("2"); // breakpoint here
		waitAMoment();
	}
	@Test
	void test3() {
		System.out.println("3");
		waitAMoment();
	}
	private void waitAMoment() {
		try {
			Thread.sleep(1000);
		} catch (InterruptedException e) {
			e.printStackTrace();
		}
	}
}
```

trancexpress:
Ah, the JUnit update job is a system job. Those are not shown in the progress view per default. I see it now.

PR #2700, opened 2025-12-17
Properly react on launch termination during test execution

iloveeclipse:
If test execution is halted in debugger on a breakpoint and "stopped" in JUnit view, the test execution properly shows terminated status for JUnit 5 now (JUnit 6 was OK).

If test execution terminated from debugger (independently on stop in JUnit view or debugger state) the test execution properly shows terminated status for JUnit 5 and 6 now.

In both cases the TestRunnerViewPart.UpdateUIJob is also properly terminated and doesn't reschedule itself every 200 ms forever.

Fixes https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2696


trancexpress:
From my POV this looks fine.

Issue #2711, opened 2025-12-22
NPE in "Externalize Strings" wizard when class already contains externalized strings


ptziegler:
Given a Java project with `org.eclipse.osgi` in the classpath and the following files:

1) `Test.java`
```java
package test;

public class Test {
	String A = Messages.Test_0;
	String B = "B";
}
```

2) `Messages.java`
```java
package test;

import org.eclipse.osgi.util.NLS;

public class Messages extends NLS {
	private static final String BUNDLE_NAME = Messages.class.getPackageName() + ".messages"; //$NON-NLS-1$
	public static String Test_0;
	static {
		// initialize resource bundle
		NLS.initializeMessages(BUNDLE_NAME, Messages.class);
	}

	private Messages() {
	}
}
```

3) `messages.properties`
```
Test_0=A
```

When opening the wizard and switching to the second page, following exception is logged, preventing me from finishing the operation:

```
Caused by: java.lang.NullPointerException: Cannot invoke "String.length()" because "s" is null
	at org.eclipse.jdt.internal.corext.refactoring.nls.PropertyFileDocumentModel.escape(PropertyFileDocumentModel.java:217)
	at org.eclipse.jdt.internal.corext.refactoring.nls.NLSPropertyFileModifier.addReplaceEdits(NLSPropertyFileModifier.java:186)
	at org.eclipse.jdt.internal.corext.refactoring.nls.NLSPropertyFileModifier.addChanges(NLSPropertyFileModifier.java:119)
	at org.eclipse.jdt.internal.corext.refactoring.nls.NLSPropertyFileModifier.create(NLSPropertyFileModifier.java:79)
	at org.eclipse.jdt.internal.corext.refactoring.nls.NLSRefactoring.createChange(NLSRefactoring.java:247)
	at org.eclipse.ltk.core.refactoring.CreateChangeOperation.run(CreateChangeOperation.java:119)
	at org.eclipse.core.internal.resources.Workspace.run(Workspace.java:2505)
	at org.eclipse.core.internal.resources.Workspace.run(Workspace.java:2533)
	at org.eclipse.ltk.internal.ui.refactoring.WorkbenchRunnableAdapter.run(WorkbenchRunnableAdapter.java:90)
	at org.eclipse.jface.operation.ModalContext$ModalContextThread.run(ModalContext.java:123)
```

The problem seems to be that the wizard includes the already externalized `Test_0` entry without loading its value.

[...]

PR #2712, opened 2025-12-23
Improve handling of Eclipse-style NLS accessors #2711

ptziegler:
## What it does

There are two issues when using the Eclipse-style accessors for resources bundles:

1) Modern accessors initialize the `RESOURCE_BUNDLE` field with `Messages.class.getPackageName() + .properties`. This is not supported by JDT, which either expect a fully qualified string or `Messages.class.getName()`. As a result, the property file is not read and the wizard only partially initialized.

2) If a property in the wizard is only partially initialized (i.e. if the value is `null`), a NullPointerException may be thrown. To avoid this, the call to `NLSSubstitution.getValue()` has been replaced with a null-safe call to `NLSSubstitution.getValueNonEmpty()`.

Closes https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2711

## How to test

Create a new project with the following files:

1) `Test.java`
```java
package test;

public class Test {
	String A = Messages.Test_0;
	String B = "B";
}
```

2) `Messages.java`
```java
package test;

import org.eclipse.osgi.util.NLS;

public class Messages extends NLS {
	private static final String BUNDLE_NAME = Messages.class.getPackageName() + ".messages"; //$NON-NLS-1$
	public static String Test_0;
	static {
		// initialize resource bundle
		NLS.initializeMessages(BUNDLE_NAME, Messages.class);
	}

	private Messages() {
	}
}
```

3) `messages.properties`
```
Test_0=A
```

Open the "Externlize Strings" wizard to externalize `B`.

The wizard should look like this:

image

instead of this:

image

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.manipulation/refactoring/org/eclipse/jdt/internal/corext/refactoring/nls/NLSHintHelper.java
+++ b/org.eclipse.jdt.core.manipulation/refactoring/org/eclipse/jdt/internal/corext/refactoring/nls/NLSHintHelper.java
@@ -288,8 +288,21 @@ public class NLSHintHelper {
 					if (InfixExpression.Operator.PLUS.equals(op)) {
 						String leftOperand = getBundleName(infix.getLeftOperand());
 						String rightOperand = getBundleName(infix.getRightOperand());
+						String result= null;
 						if (leftOperand != null && rightOperand != null) {
-							return leftOperand + rightOperand;
+							result= leftOperand + rightOperand;
+							if (infix.hasExtendedOperands()) {
+								List<Expression> extendedOperands= infix.extendedOperands();
+								for (Expression extendedOperand : extendedOperands) {
+									String extendedString= getBundleName(extendedOperand);
+									if (extendedString != null) {
+										result += extendedString;
+									} else {
+										break;
+									}
+								}
+							}
+							return result;
 						}
 						return null;
 					}
@@ -303,8 +316,9 @@ public class NLSHintHelper {
 						if (typeBinding != null) {
 							if ("getPackageName".equals(methInvocation.getName().getFullyQualifiedName())) { //$NON-NLS-1$
 								return typeBinding.getPackage().getName();
+							} else if ("getName".equals(methInvocation.getName().getFullyQualifiedName())) { //$NON-NLS-1$
+								return typeBinding.getQualifiedName();
 							}
-							return typeBinding.getQualifiedName();
 						}
 					}
 				}


org.eclipse.jdt.core.manipulation/refactoring/org/eclipse/jdt/internal/corext/refactoring/nls/NLSHintHelper.java
```diff
@@ -283,14 +283,29 @@ private String getBundleName(Expression initializer) {
...
 					MethodInvocation methInvocation= (MethodInvocation)initializer;
 					Expression exp= methInvocation.getExpression();
 					if ((exp != null) && (exp instanceof TypeLiteral)) {
 						SimpleType simple= (SimpleType)((TypeLiteral) exp).getType();
 						ITypeBinding typeBinding= simple.resolveBinding();
-						if (typeBinding != null)
+						if (typeBinding != null) {
+							if ("getPackageName".equals(methInvocation.getName().getFullyQualifiedName())) { //$NON-NLS-1$
+								return typeBinding.getPackage().getName();
+							}
```
ptziegler: I don't know whether JDT has a better way to check the method signature. The previous implementation wasn't much better, because it would just return the class name for _every_ method, not just `getName()`.

jjohnstn: You can verify it is java.lang.Class.getPackageName() via the method binding and IMethodBinding.getDeclaringClass() if you want to limit it to the normal method, but if you want to open it up to a possible local getPackageName(), this logic is fine.  You should verify there are no arguments by verifying methodInvocation.arguments().isEmpty(), otherwise, you can't guess that the result will be just the package name of the type binding.

org.eclipse.jdt.core.manipulation/refactoring/org/eclipse/jdt/internal/corext/refactoring/nls/NLSHintHelper.java
```diff
@@ -283,14 +283,29 @@ private String getBundleName(Expression initializer) {
 				if (initializer instanceof StringLiteral)
 					return ((StringLiteral)initializer).getLiteralValue();
 
+				if (initializer instanceof InfixExpression infix) {
+					InfixExpression.Operator op = infix.getOperator();
+					if (InfixExpression.Operator.PLUS.equals(op)) {
+						String leftOperand = getBundleName(infix.getLeftOperand());
+						String rightOperand = getBundleName(infix.getRightOperand());
+						if (leftOperand != null && rightOperand != null) {
```
jjohnstn: To be thorough, you need to also check for extended operands (infix.hasExtendedOperands() and then infix.extendedOperands()) which occurs if there are 3 or more items in the infix plus operation that are all the same type.

jjohnstn (changes requested):
Works well.  Just a few comments to address.

org.eclipse.jdt.core.manipulation/refactoring/org/eclipse/jdt/internal/corext/refactoring/nls/NLSHintHelper.java
jjohnstn: My bad.  We have already established we have a TypeLiteral so we don't need to check the arguments as we know we have Xxxxxx.class as expression.  There should have been a test for getName() so I have added it.

PR #2714, opened 2025-12-23
Fix regression on very fast test execution

iloveeclipse:
After the fix for #2696 I've observed (on Windows) regression in the JUnit view: if the test was terminated (regularly) very fast, the test results would be sometimes not shown at all, and JUnit view would show "Stopped" status.

I see (at least on Windows) that test launch termination is reported almost immediately, *before* any test process output is processed by ServerConnection.

The correct place to evaluate "stopped or not" condition would be after the output processing is finished, so moved the new block of code from #2696 to the ServerConnection.shutDown().

See https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2696


trancexpress:
Likely the test should be changed as well, right now it "thinks" that the stop flag will be set when a session is stopped. I didn't get session terminate notifications in the test though... So I'm not sure when we should stop waiting. Maybe in the session end of the other new listener.

trancexpress:
> This change doesn't change stop flag, so tests shouldn't be affected? Or do you mean the tests in general?

It changes the point in time at which the flag is set.

https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2709/commits/b8bacd3718d2c780f6e0c54618727b65b15fab93

Here we wait until the session is stopped, I think this is now too early to stop waiting. Since the new code will do a stop at shut down. Those are two different phases of the session.

Issue #2811, opened 2026-02-26
Failure on AssistQuickFixTest


trancexpress:
See: https://download.eclipse.org/eclipse/downloads/drops4/I20260225-1800/reports.html#tests

https://download.eclipse.org/eclipse/downloads/drops4/I20260225-1800/testresults/html/org.eclipse.jdt.ui.tests_ep439I-unit-win32-x86_64-java21_win32.win32.x86_64_21.html

```
org.eclipse.jdt.ui.tests.quickfix.AssistQuickFixTest	testAssignParamToFieldInGeneric	Failure	Content not as expected: Content is: package test1; import java.util.Vector; public class E<T> { private Vector<String>[] vec; public E(int count, Vector<String>[] vec) { super(); this.vec = vec; } } package test1; import java.util.Vector; public class E<T> { private int count; private Vector<String>[] vec; public E(int count, Vector<String>[] vec) { super(); this.count = count; this.vec = vec; } } Expected contents: package test1; import java.util.Vector; public class E<T> { private Vector<String>[] fVec; public E(int count, Vector<String>[] vec) { super(); fVec = vec; } } ==> expected: <package test1; import java.util.Vector; public class E<T> { private Vector<String>[] fVec; public E(int count, Vector<String>[] vec) { super(); fVec = vec; } } > but was: <package test1; import java.util.Vector; public class E<T> { private Vector<String>[] vec; public E(int count, Vector<String>[] vec) { super(); this.vec = vec; } } package test1; import java.util.Vector; public class E<T> { private int count; private Vector<String>[] vec; public E(int count, Vector<String>[] vec) { super(); this.count = count; this.vec = vec; } } >

org.opentest4j.AssertionFailedError: Content not as expected: Content is:
package test1;
import java.util.Vector;
public class E<T> {
private Vector<String>[] vec;

public E(int count, Vector<String>[] vec) {
super();
this.vec = vec;
}
}

package test1;
import java.util.Vector;
public class E<T> {
private int count;
private Vector<String>[] vec;

public E(int count, Vector<String>[] vec) {
super();
this.count = count;
this.vec = vec;
}
}


Expected contents:
package test1;
import java.util.Vector;
public class E<T> {
private Vector<String>[] fVec;

public E(int count, Vector<String>[] vec) {
super();
fVec = vec;
}
}

==> expected: <package test1;
import java.util.Vector;
public class E<T> {
private Vector<String>[] fVec;

[...]

PR #2812, opened 2026-02-26
Wait for indexer at end of AssistQuickFixTest.setUp()

trancexpress:
Currently the JDT indexer potentially runs in parallel to tests in `AssistQuickFixTest`. This causes a race condition in `JavaModelManager.optionsCache`, which can result in unexpected Java options during the tests.

This change adds a wait for the indexer in `AssistQuickFixTest.setUp()`, to hopefully prevent such race conditions.

Fixes: #2811


trancexpress:
See: https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2811#issuecomment-3967356378

I'm not sure this fixes the test fail, since I couldn't reproduce the same fail. Its also hard to say if this happens when all the tests are running and not just when debugging this one test method.

Anyway, it never hurts to wait for indexing to finish before running tests. Maybe we are lucky and this change is enough.

Issue #2287, opened 2025-06-24
[Test] Several tests failing since (at least) I20250620-1800


fedejeanne:
1482 failures for Mac: https://download.eclipse.org/eclipse/downloads/drops4/I20250620-1800/testresults/html/org.eclipse.jdt.ui.tests_ep437I-unit-macosx-x86_64-java21_macosx.cocoa.x86_64_21.html

1481 for **Linux** 3 days later: https://download.eclipse.org/eclipse/downloads/drops4/I20250623-1800/testresults/html/org.eclipse.jdt.ui.tests_ep437I-unit-linux-x86_64-java24_linux.gtk.x86_64_24.html

Most (if not all) of the failures show something like this:

```java
Multiple Failures (2 failures) org.eclipse.core.internal.resources.ResourceException: Resource '/TestSetupProject' does not exist. org.eclipse.jdt.core.JavaModelException: TestSetupProject does not exist

org.opentest4j.MultipleFailuresError: Multiple Failures (2 failures)
org.eclipse.core.internal.resources.ResourceException: Resource '/TestSetupProject' does not exist.
org.eclipse.jdt.core.JavaModelException: TestSetupProject does not exist
Suppressed: org.eclipse.core.internal.resources.ResourceException: Resource '/TestSetupProject' does not exist.
at org.eclipse.core.internal.resources.Resource.checkExists(Resource.java:364)
at org.eclipse.core.internal.resources.Resource.checkAccessible(Resource.java:238)
at org.eclipse.core.internal.resources.Project.checkAccessible(Project.java:183)
at org.eclipse.core.internal.resources.Folder.assertCreateRequirements(Folder.java:43)
at org.eclipse.core.internal.resources.Folder.create(Folder.java:101)
at org.eclipse.core.internal.resources.Folder.create(Folder.java:129)
at org.eclipse.jdt.internal.ui.util.ResourcesUtility.createFolder(ResourcesUtility.java:77)
at org.eclipse.jdt.internal.ui.util.CoreUtility.createFolder(CoreUtility.java:64)
at org.eclipse.jdt.testplugin.JavaProjectHelper.addSourceContainer(JavaProjectHelper.java:754)
at org.eclipse.jdt.testplugin.JavaProjectHelper.addSourceContainer(JavaProjectHelper.java:732)
at org.eclipse.jdt.testplugin.JavaProjectHelper.addSourceContainer(JavaProjectHelper.java:718)
[...]

PR #2824, opened 2026-03-04
Recreate test project if it doesnt exist

trancexpress:
JDT UI tests sporadically run into multiple failures in `CleanUpTestCase.setUp`:

```
Resource '/TestSetupProject' does not exist.
```

The error is thrown when a Java project exists in the JDT Model, but doesn't exist as a project in the resources model.

This change recreates the test project in this case, removing the project from the JDT model cache beforehand.

See: #2287


trancexpress:
Maybe we have some inconsistency between the Java model and the Eclipse resource state. `ProjectTestSetup.projectExists()` calls `IJavaElement.exists()`...

I'm not sure what the result here will be though, if that is the case. Whether recreating the project will fix the Java model state.

Issue #2830, opened 2026-03-05
JUnit 4 test case is launched with JUnit 6 so no tests are ran


trancexpress:
I'm trying to run this test case: `org.eclipse.jdt.debug.tests.ui.OpenFromClipboardTests`

It has many JUnit 4 annotations:

```
import org.junit.After;
import org.junit.AfterClass;
import org.junit.Assert;
import org.junit.Before;
import org.junit.BeforeClass;
import org.junit.Rule;
import org.junit.Test;
```

When I try to run it, the JUnit 6 launch type is used. But we decided to not include the vintage engine, since its deprecated in JUnit 6. So the launch finds no tests:

```
null
org.eclipse.jdt.junit
Error
Thu Mar 05 14:06:27 EET 2026
No tests found with test runner 'JUnit 6'.
```

So for this case we will have to explicitly set JUnit 4 for the launch type, otherwise no tests are ran. I see this both with the plain JUnit test launch and the JUnit plug-in test launch. Hopefully the code is only in JDT UI.

trancexpress:
> Why? Platform has both libraries, or did I misunderstood the querstion?

The platform does, but `org.eclipse.jdt.debug.tests` doesn't have the entire platform in its `MANIFEST.MF`. JDT should be configured to see what the plug-in has in the `MANIFEST.MF`, plus exported dependencies? It shouldn't have both JUnit 5 and JUnit 6 on the classpath or am I misunderstanding how JDT handles plug-ins via the PDE container?

PR #2831, opened 2026-03-05
Avoid detecting JUnit version based on restricted annotations

trancexpress:
`CoreTestSearchEngine.hasJUnit6TestAnnotation()` will detect JUnit 6 if a classpath entry for `junit-platform-commons` 6 exists, but has forbidden access rule restriction.

This change adjusts the search for the JUnit 5 and 6 annotation types to also consider classpath access rules.

Fixes: #2830


iloveeclipse:
> It doesn't seem like [eclipse-pde/eclipse.pde#2244](https://github.com/eclipse-pde/eclipse.pde/issues/2244) will go anywhere soon

It will, ASAP.

Issue #2840, opened 2026-03-08
Compile error in CompilationUnitDocumentProviderTest


trancexpress:
https://github.com/eclipse-platform/eclipse.platform.ui/pull/3763 caused a compile error in `CompilationUnitDocumentProviderTest`:

```
Description	Resource	Location	Path	Type
The method createLinkedProject(String, JdtTextTestPlugin, Path) is undefined for the type ResourceHelper	CompilationUnitDocumentProviderTest.java	line 107	/org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests	Java Problem
```

PR #2841, opened 2026-03-08
Fix compile error in CompilationUnitDocumentProviderTest

trancexpress:
Fixes: #2840


akurtakov (approved):
This is the right path IMO. Do not introduce dependencies for tests sake.

PR #2871, opened 2026-03-18
Add auto-closing support for Markdown fenced code blocks

### LINKED ISSUE #2529: Typing 3 backticts to start a codeblock and pressing enter does not automatically add a closing 3 backticts
Split from #1800.

In Markdown comments, typing \``` to start a codeblock and pressing enter doesn't automatically add a closing ``` at the end. This might or might not be desirable to users, so it can be considered an enhancement request. Perhaps a setting is needed, like in
Image

subyssurendran666:
image

Adds support for automatically inserting a closing triple backtick when typing ``` in Markdown comments, improving the typing experience. This behavior is controlled by a new preference so users can enable or disable it as needed.

Fix: https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2529

## Author checklist

### DIFF
--- a/org.eclipse.jdt.ui.tests/ui/org/eclipse/jdt/ui/tests/editor/MarkdownTypingTest.java
+++ b/org.eclipse.jdt.ui.tests/ui/org/eclipse/jdt/ui/tests/editor/MarkdownTypingTest.java
@@ -13,6 +13,8 @@
  *******************************************************************************/
 package org.eclipse.jdt.ui.tests.editor;
 
+import java.util.Map;
+
 import org.eclipse.jdt.testplugin.JavaProjectHelper;
 
 import org.eclipse.core.runtime.NullProgressMonitor;
@@ -26,6 +28,9 @@ import org.eclipse.ui.IEditorPart;
 import org.eclipse.jdt.core.ICompilationUnit;
 import org.eclipse.jdt.core.IJavaProject;
 import org.eclipse.jdt.core.IPackageFragment;
+import org.eclipse.jdt.core.JavaCore;
+
+import org.eclipse.jdt.ui.PreferenceConstants;
 
 import org.eclipse.jdt.internal.ui.javaeditor.CompilationUnitEditor;
 import org.eclipse.jdt.internal.ui.javaeditor.EditorUtility;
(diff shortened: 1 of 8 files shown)


org.eclipse.jdt.ui.tests/ui/org/eclipse/jdt/ui/tests/editor/MarkdownTypingTest.java
```diff
@@ -95,10 +98,10 @@ private void typeCharacter(CompilationUnitEditor editr, char c) throws Exception
 	}
 
 	public void testSingleQuoteTypingInMarkdown() throws Exception {
-		ICompilationUnit cu= createCompilationUnit();
+		ICompilationUnit cu= createCompilationUnit(null);
 		editor = openEditor(cu);
 
-		typeCharacter(editor, '\'');
+		typeCharacter(editor, '\'', 21);
```
jjohnstn: Using constant offsets is not a good idea.  I suggest having two typeCharacter() methods.  One takes a third parameter and the other one with just two parameters calculates the offset as it did before and then calls the new method with the offset.

org.eclipse.jdt.ui.tests/ui/org/eclipse/jdt/ui/tests/editor/MarkdownTypingTest.java
```diff
@@ -110,10 +113,10 @@ public class Test {}
 	}
 
 	public void testDoubleQuoteTypingInMarkdown() throws Exception {
-		ICompilationUnit cu= createCompilationUnit();
+		ICompilationUnit cu= createCompilationUnit(null);
```
jjohnstn: You can do the same with createCompilationUnit().  Have one with no parameters and one with the String parameter.  Have the no parameter method call the new method with the default String.  Therefore, in the end the original tests remain untouched.

jjohnstn (changes requested):

jjohnstn:
Hi @subyssurendran666 In trying this out, I also have the following comments:

The fencing does not seem to check the context.  For example, if I create a String x = ""; and type triple single quotes within the string, it completes them.  I don't think this should occur outside of a markdown Javadoc comment.
As well, if I have Java 21 as my compliance level, the additional quotes still get added.  If you add checks for a Javadoc markdown region, this would solve both problems.

org.eclipse.jdt.ui.tests/ui/org/eclipse/jdt/ui/tests/editor/MarkdownTypingTest.java
subyssurendran666: Done

org.eclipse.jdt.ui.tests/ui/org/eclipse/jdt/ui/tests/editor/MarkdownTypingTest.java
subyssurendran666: Done

subyssurendran666:
> Hi @subyssurendran666 In trying this out, I also have the following comments:
> 
> The fencing does not seem to check the context. For example, if I create a String x = ""; and type triple single quotes within the string, it completes them. I don't think this should occur outside of a markdown Javadoc comment. As well, if I have Java 21 as my compliance level, the additional quotes still get added. If you add checks for a Javadoc markdown region, this would solve both problems.

Thanks for the insights Jeff. Since the Markdown was introduced in Java 23, this feature will not be available for lower Java versions. 
In addition, the following scenarios has been addressed.
1. Traditional Javadoc comments will not have auto-closing support.
2. Inside double quotes, for example `String x = "<cursor is here>";`, auto-closing will not happen.
3. Inside a method/Parentheses as well, auto-closing will not happen, for `example abc(<cursor is here>);`

jjohnstn (approved):
Looks good.

PR #2881, opened 2026-03-21
Use contributed search participants in call hierarchy

arcivanov:
## Summary

Enable cross-language call hierarchy by using contributed search participants in both incoming and outgoing call analysis.

### Incoming calls (`CallerMethodWrapper`)

Replace hardcoded `new SearchParticipant[] { getDefaultSearchParticipant() }` with `SearchEngine.getSearchParticipants()` so that incoming call searches include results from contributed search participants registered via the `org.eclipse.jdt.core.searchParticipant` extension point. For contributed (non-Java) elements, accept `A_INACCURATE` matches since JDT's `MatchLocator` cannot resolve type bindings for types not compiled by ECJ.

### Outgoing calls (`CalleeMethodWrapper`)

When the member has no Java AST (i.e. `getCompilationUnitNode()` returns null), fall back to `SearchParticipant.locateCallees()` on all contributed participants. Each participant can report call sites within the member's body. Returned callees are resolved to full declarations via `resolveCallee()`, which handles `METHOD`, `FIELD`, and `TYPE` element types.

### Outgoing calls from Java to non-Java targets (`CalleeAnalyzerVisitor`)

When a Java method invocation has a null ECJ binding (receiver type is from a contributed language, not compiled by ECJ), fall back to `resolveViaSearch()` which uses `SearchEngine.getSearchParticipants()` to find `METHOD` `DECLARATIONS` matching the method name across all contributed participants.

### Declaration resolution with filtering (`CallHierarchyCore.findFirstDeclaration`)

Shared utility method `findFirstDeclaration()` searches for the first matching `IMember` declaration by name, with optional post-filters applied in `acceptSearchMatch()`:
- **Argument count** — reject candidates with wrong parameter count, with varargs support: if the last parameter is an array type, accept any expected argument count ≥ paramCount − 1 (O(1))
- **Receiver type** — reject candidates whose declaring type is not the receiver type or a supertype
- **Declaring type candidates** — reject candidates not in the candidate list
- **Argument types** — reject candidates with incompatible parameter types (FQN comparison with type resolution; for non-Java elements without a declaring type, FQN parameters are still matched while unresolvable simple names are skipped)

### Non-standard `ICompilationUnit` handling (`CallHierarchyCore`)

- Guard `getCompilationUnitNode()` with `isJavaLikeFileName()` check so non-Java source files (e.g. `.kt`) are not parsed as Java AST
- Catch `ClassCastException` for non-standard `ICompilationUnit` implementations that do not implement the internal compiler interface required by `ASTParser.createAST()`

### Dependency

Bump `org.eclipse.jdt.core` dependency to `[3.46.0,4.0.0)` for the `SearchEngine.getSearchParticipants()` and `SearchParticipant.locateCallees()` APIs.

### Related PRs

- **Depends on** eclipse-jdt/eclipse.jdt.core#4938 — adds `searchParticipant` extension point, `SearchParticipantRegistry`, `SearchEngine.getSearchParticipants()`, and `SearchParticipant.locateCallees()`
- **Companion to** eclipse-jdtls/eclipse.jdt.ls#3772 — updates jdtls search call sites to use `SearchEngine.getSearchParticipants()`

## Test plan

- [x] 6 tests in `CallHierarchyParticipantTest`:
  - `searchParticipantsIncludesContributed` — verifies contributed participant is returned by `getSearchParticipants()`
  - `outgoingCallsJavaASTPathUnchanged` — regression: Java AST callee analysis still works
  - `outgoingCallsFallBackToParticipants` — `locateCallees()` called for binary members without source, returns 2 callees
  - `outgoingCallsResolvesExistingCallees` — existing `IMember` callees resolved directly
  - `outgoingCallsFallBackOnClassCastException` — non-standard `ICompilationUnit` triggers `ClassCastException`, falls back to participant path
  - `incomingCallsUsesAllParticipants` — regression: incoming call search with all participants finds Java callers correctly
- [x] All existing `CallHierarchyTest` tests pass (regression)


org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/callhierarchy/CallHierarchyCore.java
```diff
@@ -297,18 +307,220 @@ private static StringMatcher[] parseList(String listString) {
...
+        return true;
+    }
+
+    /**
+     * Resolves a JDT parameter type signature to a fully qualified
+     * name using the declaring type's context.
+     */
+    private static String resolveParameterType(String paramSig,
+            IType declaringType) {
+        if (declaringType == null) {
```
jjohnstn: Is an IMethod in a derived source file expected to return null for getDeclaringType()?  If yes, are you going to just return a method found with same name and same number of parameters?  That seems rather inaccurate.  If you don't expect the type returned to be null, then perhaps the match method above should have bailed immediately.  Regardless, you are able to do the check in lines 455-459 whether the declaringType is null or not.

org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/callhierarchy/CallHierarchyCore.java
```diff
@@ -297,18 +307,220 @@ private static StringMatcher[] parseList(String listString) {
...
+
+    /**
+     * Checks if a candidate method matches the expected argument count
+     * and types from the call site.
+     */
+    private static boolean matchesMethod(IMethod method,
+            int expectedArgCount, String[] expectedArgTypes) {
+        int paramCount= method.getNumberOfParameters();
+        // Argument count filter
+        if (expectedArgCount >= 0
```
jjohnstn, line 416: What about methods using varargs?  A check should be made for the last argument of the method to allow more arguments than paramCount.

jjohnstn (changes requested):

org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/callhierarchy/CallHierarchyCore.java
arcivanov: Fixed — the argument count check now allows more arguments than `paramCount` when the last parameter is an array type (varargs candidate). The condition rejects if: no params, or last param is not array, or too few args for the non-varargs params (`expectedArgCount < paramCount - 1`).

Note that this entire code path (`findFirstDeclaration` / `matchesMethod` / `resolveParameterType`) is only reachable from `CalleeMethodWrapper.findCalleesFromParticipants`, which is gated by `getCompilationUnitNode()` returning null — i.e., it only runs for non-Java source files. The original Java callee analysis  (`CalleeAnalyzerVisitor`) is unaffected.

org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/callhierarchy/CallHierarchyCore.java
arcivanov: `getDeclaringType()` is not expected to return null — participants are required to provide a declaring type for all method elements (e.g., a synthetic file-facade type for top-level functions). The null guard was overly defensive.

Restructured: erasure and simple-name extraction now run before the null check, so FQN signatures (containing `.`) are returned directly regardless of declaring type. The null check is moved down to guard only the `declaringType.resolveType()` call that actually needs it.

When `resolveType()` cannot resolve a simple name (returns null), the type check for that parameter is skipped (`continue`), and matching falls back to name + argument count. This is the best the general API can do — precision beyond that is controlled by the participant's `IType.resolveType()` implementation.

This entire resolution path is exclusively non-Java (gated by `getCompilationUnitNode() == null` in `CalleeMethodWrapper.findChildren`).

org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/callhierarchy/CallHierarchyCore.java
jjohnstn: Ok

org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/callhierarchy/CallHierarchyCore.java
jjohnstn: Ok

arcivanov:
I don't understand how suppression isn't necessary as it's a `List` being assigned to `List<String>`. I thought it was a false positive.

jjohnstn (approved):
Looks good.  Getting this in will allow us to properly exercise the code via jdt.ls testing.

jjohnstn:
> I don't understand how suppression isn't necessary as it's a `List` being assigned to `List<String>`. I thought it was a false positive.

There is a compiler option to ignore assigning raw types to parameterized ones.  This option is set by default for jdt.core because this happens a lot due to legacy code.

arcivanov:
> Looks good. Getting this in will allow us to properly exercise the code via jdt.ls testing.

Thanks!

iloveeclipse:
@arcivanov , please check https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/3015, this PR seem to be related to this regression.

Issue #2774, opened 2026-01-31
Feature Request: JUnit 5 Parameterized Test Filtering and Test Enable/Disable Functionality


carstenartur:
## Summary

I would like to propose an enhancement to the JUnit view in Eclipse that significantly simplifies working with JUnit 5 parameterized tests (especially `@EnumSource`) and additionally provides a quick way to disable/enable tests directly from the JUnit view.

## Motivation

When working with parameterized tests in JUnit 5, there is currently a cumbersome workflow:

1. A single test case (e.g., a specific enum value) fails
2. To temporarily exclude it, you have to manually open the source file
3. The `@EnumSource` annotation must be manually adjusted with `mode = Mode.EXCLUDE` and `names = {...}`
4. The import for `Mode` must be added

This process is error-prone and time-consuming, especially when you want to iterate quickly.

### 1. Exclude Failing Test Cases (Context Menu)
Right-click on a failed parameterized test case in the JUnit view and select **"Exclude This Test Case"**. The `@EnumSource` annotation is automatically updated:

**Before:**
```java
@ParameterizedTest
@EnumSource(TestEnum.class)
void testWithEnum(TestEnum value) {
    // test implementation
}
```

**After:**
```java
@ParameterizedTest
@EnumSource(value = TestEnum.class, mode = Mode.EXCLUDE, names = { "VALUE2" })
void testWithEnum(TestEnum value) {
    // test implementation
}
```

The `Mode` import is automatically added to the file.

### 2. Test Disable/Enable (Context Menu)
- Right-click on any test in the JUnit view
- **"Disable This Test"** automatically adds `@Disabled` (JUnit 5) or `@Ignore` (JUnit 4)
- **"Enable This Test"** removes the annotation

This works for:
- Regular JUnit 5 tests (`@Test`)
- Regular JUnit 4 tests (`@Test`)
- Parameterized tests (`@ParameterizedTest`)
- Repeated tests (`@RepeatedTest`)
- Test factories (`@TestFactory`)
- Test templates (`@TestTemplate`)

### 3. Re-include Excluded Values (Submenu)
If you've previously excluded enum values and want to bring them back:
1. Right-click on the parameterized test method
2. Hover over **"Re-include Excluded Values"** submenu
3. Choose a specific value or **"Re-include All"**

The unused `Mode` import is automatically removed if no longer needed.

### 4. Quick Assist in Editor (Ctrl+1)
[...]

PR #2744, opened 2026-01-17
add context menu to testviewer for enable/disable

carstenartur:
#### Overview

This feature adds new capabilities to the Eclipse JUnit view that make it easier to work with JUnit 5 parameterized tests, especially those using `@EnumSource`. When a specific test case fails, you can now quickly exclude it without manually editing the source code.

##### 1. Exclude a Failing Test Case

When running a parameterized test with `@EnumSource`, each enum value appears as a separate test case in the JUnit view.  If one of these test cases fails, you can now: 

1. **Right-click** on the failed test case (e.g., `testWithEnum[VALUE2]`)
2. Select **"Exclude This Test Case"**
3. The source code is automatically updated: 

**Before:**
```java
@ParameterizedTest
@EnumSource(TestEnum.class)
void testWithEnum(TestEnum value) {
    // test implementation
}
```

**After:**
```java
@ParameterizedTest
@EnumSource(value = TestEnum.class, mode = Mode. EXCLUDE, names = { "VALUE2" })
void testWithEnum(TestEnum value) {
    // test implementation
}
```

The `Mode` import is automatically added to your file. 

---

##### 2. Disable/Enable Test Methods

You can disable any test method directly from the JUnit view:

**For regular JUnit tests:**
1. **Right-click** on a test case (e.g., `testSomething`)
2. Select **"Disable This Test"**
3. A `@Disabled` (JUnit 5) or `@Ignore` (JUnit 4) annotation is added

**For parameterized tests:**
1. **Right-click** on the test method (e.g., `testWithEnum(TestEnum)`)
2. Select **"Disable This Test"**
3. The entire parameterized test is disabled

If the test is already disabled, the menu shows **"Enable This Test"** instead, which removes the annotation.

This works for: 
- Regular JUnit 5 tests (`@Test`)
- Regular JUnit 4 tests (`@Test`)
- Parameterized tests (`@ParameterizedTest`)
- Repeated tests (`@RepeatedTest`)
- Test factories (`@TestFactory`)
- Test templates (`@TestTemplate`)

---

##### 3. Re-include Excluded Values

If you've previously excluded enum values and want to bring them back:

1. **Right-click** on the parameterized test method
2. Hover over **"Re-include Excluded Values"** submenu
3. Choose:
   - A specific value (e. g., `VALUE2`) to re-include just that one
   - **"Re-include All"** to remove all exclusions

**Before:**
```java
[...]

PR #2907, opened 2026-03-29
"Add disable/enable test functionality" from issue  #2744

carstenartur:
Related issue:
https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2774

This change contains a small part that originally came from:
https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2744

Right-clicking a test in the JUnit view now shows **"Disable This Test"** (or **"Enable This Test"** if already disabled), which adds/removes `@Disabled` (JUnit 5) or `@Ignore` (JUnit 4) to the test method in source and opens the editor.

## New files

- **`DisableTestAction`** — Context menu `Action` that toggles disabled state. Works on `TestCaseElement` (plain tests) and `TestSuiteElement` (parameterized test method nodes). Checks actual source annotation state and also treats `Result.IGNORED` as disabled.
- **`TestAnnotationModifier`** — AST-rewrite utility for `addDisabledAnnotation`, `removeDisabledAnnotation`, `isDisabled`, `isJUnit5TestMethod`, `hasAnnotation`. Handles import add/remove automatically.
- **`TestMethodFinder`** — Resolves `IMethod` from a `TestSuiteElement` whose test name follows the `"methodName(ParamType)"` pattern.

## Modified files

- **`TestViewer`** — Adds `fDisableTestAction` field, initializes it in the constructor, and conditionally appends it to the context menu for both `TestSuiteElement` and `TestCaseElement` selections.
- **`JUnitMessages` / `.properties`** — Adds `DisableTestAction_label` and `DisableTestAction_enable_label`.

## Tests

`JUnitContextMenuTest` covers:
- Toggle round-trip (`@Disabled` added then removed, annotation import cleaned up)
- No duplicate annotations on repeated invocation
- JUnit 4 `@Ignore` path

@HannesWell As far as I remember, the original change itself was rejected because it duplicated functionality that already exists in another plugin. However, I think it would still be helpful to add a comment to https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2774 explaining how users can access this functionality today, so that the solution is easier to discover.

Personally, I would still prefer to have this functionality available as part of the standard tooling, regardless of where it comes from. 

## How to test

See screenshot below:

image

## Author checklist


jjohnstn (changes requested):
I like the feature.  I have the following usage comments.

If I have a test case with two tests - in my particular case one is parameterized, the other isn't.  Disabling both tests and then enabling removes the import of Disabled annotation so the 2nd test is marked in error.  The way to do this properly is to use an ImportRemover and register the Disabled annotation node as being removed.  It will figure out if the annotation import can be removed or is still needed.  A new test for this should be added.  Also, there aren't any  parameterized test scenarios in your new menu test.

carstenartur:
> I like the feature. I have the following usage comments.
> 
> If I have a test case with two tests - in my particular case one is parameterized, the other isn't. Disabling both tests and then enabling removes the import of Disabled annotation so the 2nd test is marked in error. The way to do this properly is to use an ImportRemover and register the Disabled annotation node as being removed. It will figure out if the annotation import can be removed or is still needed. A new test for this should be added. Also, there aren't any parameterized test scenarios in your new menu test.

hi @jjohnstn,
Parameterized Support is in https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2744 . But that is too big to be reviewed and it has been pointed to another plugin where part of the functionality is already available (in a community session a few weeks ago). So I have taken that as rejection. 
Maybe it should be first clarified whether I have misstaken that?

jjohnstn:
Hi @carstenartur My take on #2744 was that it was going to be broken down into smaller chunks - not rejected.  For this patch, parameterized tests do allow toggling.  If I run a parameterized test and right click in the Junit view, I get the option to enable/disable the test and this works so it should have a test added.  The import remove problem needs to be fixed and with a test added to verify this, this patch can be merged.  Once this patch is merged, more functionality from #2744 can be added without too many lines of code to review.

jjohnstn (approved):
Looks good.

Issue #2903, opened 2026-03-27
TestKindRegistry.getContainerTestKindId detects the wrong JUnit type


trancexpress:
When launching JUnit tests, `org.eclipse.jdt.internal.junit.launcher.TestKindRegistry.getContainerTestKindId(IJavaElement)` is used to determine which JUnit version should be used, so far 3, 4, 5 or 6.

Currently the method looks like this:

```
	public static String getContainerTestKindId(IJavaElement element) {
		new Exception().printStackTrace(System.out);
		if (element != null) {
			IJavaProject project= element.getJavaProject();
			if (isRunWithJUnitPlatform(element)) { // check if we are using a JUnit 4 Runner
				return JUNIT4_TEST_KIND_ID;
			}
			if (CoreTestSearchEngine.hasJUnit4TestAnnotation(project)) { // check if we can find @Testable on the classpath, from JUnit 6
				return JUNIT6_TEST_KIND_ID;
			}
			if (CoreTestSearchEngine.hasJUnit6TestAnnotation(project)) { // check if we can find @Testable on the classpath, from JUnit 5
				return JUNIT5_TEST_KIND_ID;
			}
			if (CoreTestSearchEngine.hasJUnit5TestAnnotation(project)) { // check if we can find @org.junit.Test on the classpath
				return JUNIT4_TEST_KIND_ID;
			}
		}
		return JUNIT3_TEST_KIND_ID;
	}
```

The checks for `@org.junti.Test` and `@org.junit.platform.commons.annotation.Testable` look for the respective type on the projects classpath.

Currently those checks don't respect access restrictions on the classpath entries, which after https://github.com/eclipse-pde/eclipse.pde/issues/2244 results in problems like this one: https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2830

As soon as a plug-in depends on `org.apache.ant` (e.g. due to depending on `org.eclipse.tests`), the plug-in PDE container will contain also JUnit 5 and JUnit 6 bundles (as `org.apache.ant` optionally depends on `junit-platform-launcher` 6). As a result, all tests are considered JUnit 6 tests.

[...]

PR #2910, opened 2026-03-30
Use junit-jupiter-api test annotations to detect JUnit 5 and 6

trancexpress:
This change adds the following annotations to the search performed in `CoreTestSearchEngine.hasJUnitJupiterTestAnnotation()`:

* `org.junit.jupiter.api.Test`
* `org.junit.jupiter.api.TestTemplate`
* `org.junit.jupiter.api.ClassTemplate`

Additionally access restrictions on the found annotation type are applied, in order to handle JUnit 5 or 6 as transitive classapth entries of a plug-in. Such classpath entries come from the PDE classpath container and are have forbidden access pattern for all classes.

Applying the access restrictions prevents detecting the wrong JUnit version for plug-in tests, but may also fail to detect the correct JUnit version. For this reason we search also for the mentioned annotation types.

The change doesn't handle cases where the JUnit annotation `Testable` is extended, but neither `Testable` nor the annotations above are found on the accessible classpath. In such cases the version detection defaults to JUnit 3, which is the current default.

Fixes: https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2903


iloveeclipse:
Let merge this one for now, it fixes regression.

PR #2975, opened 2026-05-06
feat(junit.runtime): support multi-method launches in a single JVM

wenytang-ms:
Allow a `-testNameFile` to contain lines of the form `ClassName:methodName` so that an arbitrary set of methods (potentially across multiple classes) can be discovered and executed inside a single test JVM. This avoids the per-method setup/teardown cycles that currently 
occur when a user runs more than one method via Eclipse's Test Explorer (or via `vscode-java-test`, which sits on top of this protocol).
 
 * `ITestLoader` gains a default `loadTests(LinkedHashMap<Class<?>, List<String>>, ...)` method that pair-wise delegates to the legacy single-method overload, so every existing loader keeps working unchanged.
 * `RemoteTestRunner.readTestNames()` recognises `:`-separated lines and routes through the new dispatch when present; class-only lines keep their previous semantics so older clients are unaffected.
 * `JUnit5TestLoader` builds one `LauncherDiscoveryRequest` with multiple `selectMethod` selectors, letting JUnit Platform run the entire selection natively within a single test JVM.
 * `JUnit4TestLoader` composes a `Request.classWithoutSuiteMethod(c).filterWith(...)` using a new `MultiMethodFilter` to express the same selection on JUnit 4.
 * Both `JUnitLaunchConfigurationDelegate` variants (in `org.eclipse.jdt.junit.core` and `org.eclipse.jdt.ui.unittest.junit`) relax `createTestNamesFile()` to also write `IMethod` elements as `Class:method` lines, allowing the new behaviour to be reached via the existing 
`-testNameFile` path without any new protocol parameters.
 
 ## What it does
 
 When a user selects multiple test methods in Eclipse's Test Explorer (or via clients such as `vscode-java-test`), each method is currently launched in its own JVM, because the existing protocol can carry at most one `(class, method)` pair per launch. For tests with expensive
 per-class setup (e.g. Spring `@SpringBootTest`, Testcontainers), this means the fixture is rebuilt N times.
 
 This change extends the existing `-testNameFile` protocol with a fully backwards-compatible `Class:method` line format and lets every test loader opt-in to multi-method discovery:
 
 - `ITestLoader` gains a default `loadTests(LinkedHashMap<Class<?>, List<String>>, ...)` that pair-wise delegates to the legacy overload, so loaders that don't override it keep behaving exactly as before.
 - `RemoteTestRunner.readTestNames()` recognises `:`-separated lines and routes through the new dispatch only when at least one such line is present. Class-only lines keep their previous semantics, so older clients (and old `-testNameFile` files) are unaffected.
 - `JUnit5TestLoader` builds a single `LauncherDiscoveryRequest` with multiple `selectMethod` selectors, letting JUnit Platform discover the whole selection in one JVM.
 - `JUnit4TestLoader` composes per class `Request.classWithoutSuiteMethod(c).filterWith(new MultiMethodFilter(...))` to express the same selection on JUnit 4. The filter strips both the parameter list `(...)` and the `[idx]` suffix added by `Parameterized` before matching.
 - Both `JUnitLaunchConfigurationDelegate` variants (`org.eclipse.jdt.junit.core` and `org.eclipse.jdt.ui.unittest.junit`) relax `createTestNamesFile` to also accept `IMethod` elements and emit the new `Class:method` lines.
 
 JUnit 3 needs no changes — it inherits the default `ITestLoader` impl which still loops per `(class, method)`.
 
 ## How to test
 
 1. Open an Eclipse workspace with a JUnit 5 test class that has multiple `@Test` methods and a non-trivial `@BeforeAll` (e.g. starting an embedded server, or a Spring `@SpringBootTest`).
 2. In the JUnit view / Project Explorer, multi-select two or more methods inside the same class and **Run As → JUnit Test**.
 3. Observe that:
    - There is a single test JVM (one entry in the Console view, one `RemoteTestRunner` process).
    - `@BeforeAll` runs once and `@AfterAll` runs once for the whole selection.
    - All selected methods report Pass/Fail individually in the JUnit view.
 4. Repeat with methods spread across two different classes — both should still execute inside one JVM, with each class's own `@BeforeAll` / `@AfterAll` firing once.
 5. Repeat with a JUnit 4 class (including a `@RunWith(Parameterized.class)` test) to verify the JUnit 4 path and the `[idx]` / `(params)` stripping.
 6. Sanity check backwards compatibility: a launch from an older client that still writes only class FQNs to `-testNameFile` continues to run as before (whole class executed).
 
 ## Author checklist
 
### DIFF
--- a/org.eclipse.jdt.junit.core/src/org/eclipse/jdt/junit/launcher/JUnitLaunchConfigurationDelegate.java
+++ b/org.eclipse.jdt.junit.core/src/org/eclipse/jdt/junit/launcher/JUnitLaunchConfigurationDelegate.java
@@ -696,6 +696,20 @@ public class JUnitLaunchConfigurationDelegate extends AbstractJavaLaunchConfigur
 						String testName= type.getFullyQualifiedName();
 						bw.write(testName);
 						bw.newLine();
+					} else if (testElement instanceof IMethod) {
+						// Extended -testNameFile format: "fully.qualified.ClassName:methodName".
+						// Allows running an arbitrary set of methods (potentially across
+						// multiple classes) within a single launch. The remote runner falls
+						// back to legacy class-only behavior when no ':' is present, so older
+						// runtime jars remain compatible.
+						IMethod method= (IMethod) testElement;
+						IType declaringType= method.getDeclaringType();
+						if (declaringType == null) {
+							abort(JUnitMessages.JUnitLaunchConfigurationDelegate_error_wrong_input, null, IJavaLaunchConfigurationConstants.ERR_UNSPECIFIED_MAIN_TYPE);
+						} else {
+							bw.write(declaringType.getFullyQualifiedName() + ':' + method.getElementName());
+							bw.newLine();
+						}
 					} else {
 						abort(JUnitMessages.JUnitLaunchConfigurationDelegate_error_wrong_input, null, IJavaLaunchConfigurationConstants.ERR_UNSPECIFIED_MAIN_TYPE);
 					}
--- a/org.eclipse.jdt.junit.runtime/META-INF/MANIFEST.MF
+++ b/org.eclipse.jdt.junit.runtime/META-INF/MANIFEST.MF
@@ -3,7 +3,7 @@ Automatic-Module-Name: org.eclipse.jdt.junit.runtime
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.jdt.junit.runtime;singleton:=true
-Bundle-Version: 3.8.0.qualifier
+Bundle-Version: 3.8.100.qualifier
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
 Export-Package: 
--- a/org.eclipse.jdt.junit.runtime/pom.xml
+++ b/org.eclipse.jdt.junit.runtime/pom.xml
@@ -18,7 +18,7 @@
   </parent>
   <groupId>org.eclipse.jdt</groupId>
   <artifactId>org.eclipse.jdt.junit.runtime</artifactId>
-  <version>3.8.0-SNAPSHOT</version>
+  <version>3.8.100-SNAPSHOT</version>
   <packaging>eclipse-plugin</packaging>
   <build>
 	<plugins>
--- a/org.eclipse.jdt.junit.runtime/src/org/eclipse/jdt/internal/junit/runner/ITestLoader.java
+++ b/org.eclipse.jdt.junit.runtime/src/org/eclipse/jdt/internal/junit/runner/ITestLoader.java
@@ -16,6 +16,11 @@
 
 package org.eclipse.jdt.internal.junit.runner;
 
+import java.util.ArrayList;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+
 public interface ITestLoader {
 	/**
 	 * @param testClasses classes to be run
@@ -28,5 +33,58 @@ public interface ITestLoader {
 	 * @return the loaded test references
 	 */
 	ITestReference[] loadTests(Class<?>[] testClasses, String testName, String[] failureNames, String[] packages, String[][] includeExcludeTags, String uniqueId, RemoteTestRunner listener);
+
+	/**
+	 * Loads tests for an arbitrary collection of (class, method) pairs in a single launch.
+	 * <p>
+	 * The default implementation delegates to {@link #loadTests(Class[], String, String[], String[], String[][], String, RemoteTestRunner)}
+	 * for each pair, preserving the legacy behavior of one filtered request per method.
+	 * Loaders that can express multi-method discovery natively (e.g. JUnit Platform via multiple
+	 * {@code selectMethod} selectors, JUnit 4 via a multi-method filter) are encouraged to override
+	 * this method so that the entire selection runs inside a single launch.
+	 * </p>
+	 *
+	 * @param classToMethods ordered map from test class to the set of methods to run within that class
+	 * @param failureNames may want to run these first, since they failed
+	 * @param packages packages containing tests to run
+	 * @param includeExcludeTags tags to be included and excluded in the test run
+	 * @param uniqueId unique ID of the test to run
+	 * @param listener to be notified if tests could not be loaded
+	 * @return the loaded test references
+	 */
+	default ITestReference[] loadTests(LinkedHashMap<Class<?>, List<String>> classToMethods, String[] failureNames, String[] packages, String[][] includeExcludeTags, String uniqueId, RemoteTestRunner listener) {
+		List<ITestReference> refs= new ArrayList<>();
+		for (Map.Entry<Class<?>, List<String>> entry : classToMethods.entrySet()) {
+			Class<?>[] singleClass= { entry.getKey() };
+			List<String> methods= entry.getValue();
+			if (methods == null || methods.isEmpty()) {
+				// An empty / null method list represents an unfiltered class run, which
+				// the legacy single-method overload models by passing testName == null.
+				// This lets callers (e.g. RemoteTestRunner) mix class-only and
+				// Class:method entries from the same -testNameFile without losing the
+				// class-only selections.
+				ITestReference[] wholeClass= loadTests(singleClass, null, failureNames, packages, includeExcludeTags, uniqueId, listener);
+				if (wholeClass != null) {
+					for (ITestReference r : wholeClass) {
+						if (r != null) {
+							refs.add(r);
+						}
+					}
+				}
+				continue;
+			}
+			for (String methodName : methods) {
+				ITestReference[] perMethod= loadTests(singleClass, methodName, failureNames, packages, includeExcludeTags, uniqueId, listener);
+				if (perMethod != null) {
+					for (ITestReference r : perMethod) {
+						if (r != null) {
+							refs.add(r);
+						}
+					}
+				}
+			}
+		}
+		return refs.toArray(new ITestReference[0]);
+	}
 }
 
--- a/org.eclipse.jdt.junit.runtime/src/org/eclipse/jdt/internal/junit/runner/RemoteTestRunner.java
+++ b/org.eclipse.jdt.junit.runtime/src/org/eclipse/jdt/internal/junit/runner/RemoteTestRunner.java
@@ -26,6 +26,11 @@ import java.io.StringWriter;
 import java.io.UnsupportedEncodingException;
 import java.net.Socket;
 import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
+import java.util.List;
 import java.util.Vector;
 
 import org.eclipse.jdt.internal.junit.runner.junit3.JUnit3TestLoader;
@@ -65,6 +70,15 @@ public class RemoteTestRunner implements MessageSender, IVisitsTestTrees {
 	 * The name of the test (argument -test)
 	 */
 	private String fTestName;
+	/**
+	 * Optional mapping from test class name to the list of method names to execute
+	 * within that class. Populated when {@code -testNameFile} contains lines of the
+	 * form {@code ClassName:methodName}. When non-empty, multi-method dispatch is
+	 * used so that the entire selection runs in a single launch. When empty, the
+	 * legacy single-method ({@link #fTestName}) path is used and behavior is
+	 * unchanged.
+	 */
+	private LinkedHashMap<String, List<String>> fMethodsByClass= new LinkedHashMap<>();
 	/**
 	 * The names of the packages containing tests to run
 	 */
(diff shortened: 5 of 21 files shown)


org.eclipse.jdt.junit5.runtime/src/org/eclipse/jdt/internal/junit5/runner/JUnit5TestLoader.java
```diff
@@ -59,6 +62,44 @@ public ITestReference[] loadTests(Class<?>[] testClasses, String testName, Strin
 		return refs;
 	}
 
+	@Override
+	public ITestReference[] loadTests(LinkedHashMap<Class<?>, List<String>> classToMethods, String[] failureNames, String[] packages, String[][] includeExcludeTags, String uniqueId, RemoteTestRunner listener) {
```
trancexpress, line 66: Please make like-wise changes in `JUnit6TestLoader`, which is mostly a copy of `JUnit5TestLoader`.

wenytang-ms: added

trancexpress:
This will need tests. Unfortunately no time to check where they can be added.

What is the entry point for this change, in `ILaunchConfiguration`? Please add tests for that entry point, maybe somewhere around `AdvancedJUnitLaunchConfigurationDelegateTest`.

wenytang-ms:
@trancexpress  Feature Test with junit 5
image
image

iloveeclipse:
Please double check, there are two new compiler warnings

trancexpress:
@wenytang-ms I'm not sure what the screenshot should state? The entire JUnit launch is already executed in the same JVM, at least when I run the test case from Eclipse.

What is the entry point for `vscode-java-test`? And what is Eclipse's Test Explorer?

trancexpress:
These tests seem to run JUnit tests and check the resulting executed tests:

- `org.eclipse.jdt.junit.tests.TestRunListenerTest4`
- `org.eclipse.jdt.junit.tests.TestRunListenerTest5`
- `org.eclipse.jdt.junit.tests.TestRunListenerTest6` 

Adding a test to each to check if the new format works (multiple methods in different classes) should be enough testing here, I think.

I'm still not sure what the entry point is though, i.e. how JDT UI is configured to do this type of launching.

wenytang-ms:
@trancexpress sorry, the screenshot was misleading — let me clarify.
 
 **The problem this PR fixes is not visible from Eclipse SDK's UI.** When you Run As → JUnit Test from Eclipse on a single method / class / source folder, the existing protocol already runs everything in one JVM, so you correctly see no difference. The `JUnitLaunchShortcut` 
`contextualLaunch` enablement does not allow selecting multiple methods at once, so the multi-method case simply cannot be triggered from the SDK UI today.
 
 **The affected entry point is downstream consumers**, primarily [`vscode-java-test`](https://github.com/microsoft/vscode-java-test). It subclasses `JUnitLaunchConfigurationDelegate` and overrides `evaluateTests(...)` to return an `IMethod[]` for whatever the user has ticked 
in VS Code's Test Explorer (which does support multi-method selection). With the current protocol, `fTestElements.length > 1` containing methods causes `createTestNamesFile` to write class-only lines and `RemoteTestRunner` then dispatches one JVM per method — for 
`@SpringBootTest` style suites this means N full Spring contexts instead of one. The screenshot was a vscode-java-test reproduction showing 3 PIDs collapsing into 1 after this PR.
 
 So the change has zero effect on existing JDT UI launches and only kicks in when an `ILaunchConfiguration` produces an `IMethod[]` with length > 1, which today only happens via downstream delegates like vscode-java-test.
 
 I'll add the multi-method cross-class cases to `TestRunListenerTest4/5/6` as you suggested — agreed that exercising the full RemoteTestRunner round-trip is much stronger than asserting on the `-testNameFile` contents alone.

wenytang-ms:
updated: the e2e test case added.

wenytang-ms:
@trancexpress please help review, thanks!

org.eclipse.jdt.junit.runtime/src/org/eclipse/jdt/internal/junit/runner/ITestLoader.java
```diff
@@ -28,5 +33,59 @@ public interface ITestLoader {
...
+	 * </p>
+	 *
+	 * @param classToMethods ordered map from test class to the set of methods to run within that class
+	 * @param failureNames may want to run these first, since they failed
+	 * @param packages packages containing tests to run
+	 * @param includeExcludeTags tags to be included and excluded in the test run
+	 * @param uniqueId unique ID of the test to run
+	 * @param listener to be notified if tests could not be loaded
+	 * @return the loaded test references
+	 * @since 3.10
```
trancexpress: This doesn't match the bundle version.

@iloveeclipse what was the guideline for bundle versions after adding new API? Do we bump the minor segment here? Or the micro is enough?

org.eclipse.jdt.junit.core/src/org/eclipse/jdt/junit/launcher/JUnitLaunchConfigurationDelegate.java
```diff
@@ -696,6 +696,20 @@ private String createTestNamesFile(IJavaElement[] testElements) throws CoreExcep
 						String testName= type.getFullyQualifiedName();
 						bw.write(testName);
 						bw.newLine();
+					} else if (testElement instanceof IMethod) {
+						// Extended -testNameFile format: "fully.qualified.ClassName:methodName".
+						// Allows running an arbitrary set of methods (potentially across
+						// multiple classes) within a single test JVM. The remote runner falls
```
trancexpress: I think the single JVM is somewhat confusing here, since JUnit launches run in a single JVM. Maybe a single launch is less confusing here.

The single JVM is a problem coming from not having the new API method and having to do multiple launches, i.e. its a workaround.

Can you adjust to "single launch", here and in the other mentions of a JVM?

org.eclipse.jdt.junit.runtime/src/org/eclipse/jdt/internal/junit/runner/ITestLoader.java
```diff
@@ -28,5 +33,59 @@ public interface ITestLoader {
...
+	 */
+	default ITestReference[] loadTests(LinkedHashMap<Class<?>, List<String>> classToMethods, String[] failureNames, String[] packages, String[][] includeExcludeTags, String uniqueId, RemoteTestRunner listener) {
+		List<ITestReference> refs= new ArrayList<>();
+		for (Map.Entry<Class<?>, List<String>> entry : classToMethods.entrySet()) {
+			Class<?>[] singleClass= new Class<?>[] { entry.getKey() };
```
trancexpress: Can you change this to: `Class<?>[] singleClass= { entry.getKey() };`

org.eclipse.jdt.junit4.runtime/META-INF/MANIFEST.MF
```diff
@@ -4,7 +4,7 @@ Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-Vendor: %providerName
 Bundle-SymbolicName: org.eclipse.jdt.junit4.runtime;singleton:=true
-Bundle-Version: 1.4.0.qualifier
```
trancexpress, line 7: Please keep the version bump only in the bundle with the new `org.eclipse.jdt.internal.junit.runner.ITestLoader` API, i.e. only in `org.eclipse.jdt.junit.runtime`. The other bumps can be done by the bot.

Or you require the bumps for `vscode-java`?

org.eclipse.jdt.junit6.runtime/src/org/eclipse/jdt/internal/junit6/runner/JUnit6TestLoader.java
```diff
@@ -59,6 +62,53 @@ public ITestReference[] loadTests(Class<?>[] testClasses, String testName, Strin
...
+		if (classToMethods == null || classToMethods.isEmpty()) {
+			return new ITestReference[0];
+		}
+		// Build a single LauncherDiscoveryRequest covering every selected method across
+		// every selected class. JUnit Platform supports multiple selectors natively, so
+		// the resulting test plan is executed inside one JVM, sharing per-class
+		// @BeforeAll / @AfterAll lifecycle and any cached fixture (e.g. a Spring
+		// ApplicationContext). A null/empty method list represents an unfiltered class
+		// run, which lets a single -testNameFile mix legacy class lines with
+		// Class:method lines without losing the class-only selections.
```
trancexpress, line 77: Please sync the comments between the JUnit 5 and 6 versions. This one has more infos, so ideally the JUnit 5 version also has the extra infos.

org.eclipse.jdt.ui.tests/ui/org/eclipse/jdt/junit/tests/AbstractTestRunListenerTest.java
```diff
@@ -236,6 +312,43 @@ public static void assertEqualLog(final String[] expectedSequence, String[] logM
...
+	 * {@code sessionStarted}/{@code sessionFinished} pair plus a {@code testCaseStarted}
+	 * and {@code testCaseFinished} event for each of the supplied {@code class#method}
+	 * identifiers (in any order). Used by the multi-method launch tests because the
+	 * dispatch order across classes is implementation-defined.
+	 */
+	protected static void assertMultiMethodRun(String[] log, String... expectedClassHashMethod) {
+		java.util.regex.Pattern p= java.util.regex.Pattern.compile("testCaseMethod: (.+?)\\s*[\\r\\n]+\\s*class: (\\S+)");
+		int sessionStartedCount= 0;
+		int sessionFinishedCount= 0;
+		java.util.Set<String> started= new java.util.TreeSet<>();
```
trancexpress: Can we get rid of the fully-qualified types in this method?

trancexpress:
Looks good, just a few minor things to address. First and foremost the bundle version bump where new API is added.

We have infrastructure trouble with I-builds, which probably means we won't get a new I-build with this change anytime soon. Since the change looks fine, I doubt there is high risk.

org.eclipse.jdt.junit.runtime/src/org/eclipse/jdt/internal/junit/runner/ITestLoader.java
iloveeclipse: It was always minor segment, **if not yet updated for the release**.

trancexpress: OK, 4.39 has `3.8`. So please bump the bundle to `3.9` and update the `@since` here to match.

trancexpress:
Alright, looks good. Please squash your 2 commits (if the bot makes one, don't squash it).

org.eclipse.jdt.junit.runtime/src/org/eclipse/jdt/internal/junit/runner/ITestLoader.java
wenytang-ms: org.eclipse.jdt.junit.runtime minor bump：3.8.100 → 3.9.0；@since 3.10 → @since 3.9

org.eclipse.jdt.junit.core/src/org/eclipse/jdt/junit/launcher/JUnitLaunchConfigurationDelegate.java
wenytang-ms: you are right,  "single JVM" → "single launch"

org.eclipse.jdt.junit.runtime/src/org/eclipse/jdt/internal/junit/runner/RemoteTestRunner.java
```diff
@@ -65,6 +70,17 @@ public RerunRequest(int testId, String className, String testName) {
...
 	private String fTestName;
+	/**
+	 * Optional mapping from test class name to the list of method names to execute
+	 * within that class. Populated when {@code -testNameFile} contains lines of the
+	 * form {@code ClassName:methodName}. When non-empty, multi-method dispatch is
+	 * used so that the entire selection runs in a single launch. When empty, the
+	 * legacy single-method ({@link #fTestName}) path is used and behavior is
+	 * unchanged.
+	 *
+	 * @since 3.9
```
trancexpress: Not needed.

org.eclipse.jdt.junit4.runtime/META-INF/MANIFEST.MF
wenytang-ms: No, we do not need the version bumps in the junit4/5/6.runtime bundles for vscode-java. The capability gate on the vscode-java-test side checks the org.eclipse.jdt.junit.runtime bundle version only (the one with the new ITestLoader API), and the JUnit{4,5,6} runtime jars are
 shipped on a flat classpath, not via OSGi resolution. I have reverted those bumps and kept the version bump only in org.eclipse.jdt.junit.runtime (3.8.100 → 3.9.0). The bot can bump the others.

org.eclipse.jdt.junit.runtime/META-INF/MANIFEST.MF
```diff
@@ -3,7 +3,7 @@ Automatic-Module-Name: org.eclipse.jdt.junit.runtime
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.jdt.junit.runtime;singleton:=true
-Bundle-Version: 3.8.0.qualifier
+Bundle-Version: 3.9.0.qualifier
```
iloveeclipse: Looking at changed code, none of the classes are API. Why do we bump version here to 3.9.0 and not to 3.8.100?

org.eclipse.jdt.junit.runtime/src/org/eclipse/jdt/internal/junit/runner/ITestLoader.java
```diff
@@ -28,5 +33,59 @@ public interface ITestLoader {
...
+	 * </p>
+	 *
+	 * @param classToMethods ordered map from test class to the set of methods to run within that class
+	 * @param failureNames may want to run these first, since they failed
+	 * @param packages packages containing tests to run
+	 * @param includeExcludeTags tags to be included and excluded in the test run
+	 * @param uniqueId unique ID of the test to run
+	 * @param listener to be notified if tests could not be loaded
+	 * @return the loaded test references
+	 * @since 3.9
```
iloveeclipse: This is not API, no `since` needed

org.eclipse.jdt.junit.runtime/src/org/eclipse/jdt/internal/junit/runner/RemoteTestRunner.java
iloveeclipse: This is not API, no since needed

org.eclipse.jdt.junit.runtime/pom.xml
```diff
@@ -18,7 +18,7 @@
   </parent>
   <groupId>org.eclipse.jdt</groupId>
   <artifactId>org.eclipse.jdt.junit.runtime</artifactId>
-  <version>3.8.0-SNAPSHOT</version>
+  <version>3.9.0-SNAPSHOT</version>
```
iloveeclipse: Must be `3.8.100`

org.eclipse.jdt.junit.runtime/META-INF/MANIFEST.MF
trancexpress: Ah, I missed that `ITestLoader` is internal. In that case we don't need any `@since` tags and the version bumps can be handled by the bot.

Must be why the API analysis is complaining...

org.eclipse.jdt.junit.runtime/pom.xml
wenytang-ms: Done — pom version is now 3.8.100-SNAPSHOT, matching the manifest.

org.eclipse.jdt.junit.runtime/src/org/eclipse/jdt/internal/junit/runner/RemoteTestRunner.java
wenytang-ms: Done — @since tag removed. This should also clear the two API-analysis warnings flagged earlier.

org.eclipse.jdt.junit.runtime/src/org/eclipse/jdt/internal/junit/runner/ITestLoader.java
wenytang-ms: Done — @since tag removed.

org.eclipse.jdt.junit.runtime/META-INF/MANIFEST.MF
wenytang-ms: Done. Reverted the version to 3.8.100.qualifier since org.eclipse.jdt.internal.junit.runner is an internal package, so this is not new API and only a micro bump is required.

trancexpress (approved):

trancexpress:
Anything against merging this @iloveeclipse ?

iloveeclipse:
> Anything against merging this

I haven't reviewed this in detail, so if you are OK, im'm fine.

PR #2977, opened 2026-05-08
Remove obsolete ICU references from surrogate-handling comments

vogella:
The XXX comments in JavaWordFinder and NewTestCaseWizardPageOne referenced com.ibm.icu.lang.UCharacter and java.lang.Character behavior on Java 1.4. ICU is no longer a dependency of this repo, and the surrogate branches in those if/else blocks contained no actual logic beyond the comment, so I folded the surrogate check into the outer condition and dropped the stale comments. Behavior is unchanged.


jjohnstn (approved):
Looks good.

PR #2980, opened 2026-05-12
Add check for Other Projects working set in JavaWorkingSetUpdater

jjohnstn:
- in JavaWorkingSetUpdater.elementChange() make sure that the Other Projects working set is processed either as part of the local list or by searching for it

## What it does

See https://github.com/eclipse-platform/eclipse.platform.ui/issues/2250#issuecomment-4407538174

## How to test

See issue above.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/workingsets/JavaWorkingSetUpdater.java
+++ b/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/workingsets/JavaWorkingSetUpdater.java
@@ -25,8 +25,10 @@ import org.eclipse.core.resources.IResource;
 import org.eclipse.core.resources.IResourceDelta;
 import org.eclipse.core.resources.ResourcesPlugin;
 
+import org.eclipse.ui.ILocalWorkingSetManager;
 import org.eclipse.ui.IWorkingSet;
 import org.eclipse.ui.IWorkingSetUpdater;
+import org.eclipse.ui.PlatformUI;
 
 import org.eclipse.jdt.core.ElementChangedEvent;
 import org.eclipse.jdt.core.IElementChangedListener;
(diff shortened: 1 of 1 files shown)


nettozahler:
Hi @jjohnstn . Meanwhile I have tested the fix. But it does not work for me. Neither the "new project" case nor the the "delete project". So for me the wrong behavior is unchanged. I also don't see how the fix should work since the "others working set" is a dummy only, the code line
`IWorkingSet other= manager.getWorkingSet(WorkingSetMessages.WorkingSetModel_others_name);`
always yields `null` (at least for me). Did it work for you with the settings I have described?

jjohnstn:
Hi @nettozahler The scenario did work for me, but perhaps I didn't set up things properly.  I did see projects being deleted from the Other Projects set and these were also deleted from the other views.  I will take another look.

jjohnstn:
@nettozahler You are correct that the new code is not doing anything.  I will remove it for RC1.  I can't get the failures you are seeing.  I tried the following from a clean workspace created in a child Eclipse with latest JDT Core and JDT UI checked out and using latest I-builds for a Target platform:

- open the Java Browsing perspective
- minimize the Projects window and use Window->Show View to open the Project Explorer
- change the Project Explorer to show Working Sets and Hierarchical
- create a new Java Project from the Project Explorers view link and in the dialog hit the New button to create a Java working set and then specify it for the new Java project

At this point, I see the new project but the new working set does not appear in the Project Explorer view.  If I then do File -> New -> Java Project and add a new java project with no working set specified, it creates the project and I see it in the Project Explorer (again, no working set is displayed).  If I open the Package Explorer, I can see the working sets (MyWorkingSet, and Other Projects and both are populated).  Even I don't open it, I can delete either project in the Project Explorer and they are deleted from this view and the Projects view which here is minimized, but also updates if not minimized.

Can you indicate which steps I should change/add to get to just before the error scenarios you have mentioned in the issue?

nettozahler:
@jjohnstn Please try like this: view settings Project Presentation = "Hierarchical" and Top Level Elements = "Working Sets". Have at least one self-defined Working Set (WS) present containing a project. Assert there is no "Other projects" WS yet. Then start your test. During new project creation do NOT create or assign a WS so the newly created project should appear in "Other projects" WS.

PR #2985, opened 2026-05-14
Fix SubMonitor misuse in UndoablePackageDeleteChange

vogella:
The `perform()` method mixed `pm.worked()` with `SubMonitor.convert()` on the same monitor, and called `SubMonitor.convert(pm, 1)` inside a loop. Both patterns corrupt progress reporting: `convert()` resets the monitor's remaining work, so repeated calls inside a loop or after `worked()` discard already-reported progress.

The fix replaces `beginTask`/`worked` and the in-loop `SubMonitor.convert()` calls with a single `SubMonitor.convert()` up front, then `split()` for each unit of work. The loops are also simplified to enhanced for-each since the index is no longer needed.

## Author checklist


jjohnstn (approved):
Looks good.

PR #2987, opened 2026-05-15
Use ILog API for logging in org.eclipse.jdt.junit.core

vogella:
Replace the JUnitCorePlugin.log() wrapper methods with direct use of the Eclipse ILog API. Files with multiple logging call sites get a private static final ILog field via ILog.of(MyClass.class); single-call sites use ILog.of(MyClass.class) inline. This removes the indirection through the plugin activator and stamps each log entry with the bundle that actually owns the code, rather than always attributing it to the UI plugin.

### DIFF
--- a/org.eclipse.jdt.junit.core/src/org/eclipse/jdt/internal/junit/util/CoreTestSearchEngine.java
+++ b/org.eclipse.jdt.junit.core/src/org/eclipse/jdt/internal/junit/util/CoreTestSearchEngine.java
@@ -24,9 +24,9 @@ import org.osgi.framework.Version;
 import org.eclipse.jdt.junit.JUnitCore;
 
 import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.ILog;
 import org.eclipse.core.runtime.IPath;
 import org.eclipse.core.runtime.IProgressMonitor;
-import org.eclipse.core.runtime.Status;
 
 import org.eclipse.jdt.core.Flags;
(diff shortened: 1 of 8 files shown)


org.eclipse.jdt.junit.core/src/org/eclipse/jdt/internal/junit/util/CoreTestSearchEngine.java
```diff
@@ -230,7 +230,7 @@ private static boolean hasJUnitJupiterTestAnnotation(IJavaProject project, int j
 								}
 							}
 						} catch (Throwable e) {
-							JUnitCorePlugin.log(Status.warning("Failed to determine JUnit version", e)); //$NON-NLS-1$
+							ILog.of(CoreTestSearchEngine.class).warn("Failed to determine JUnit version", e); //$NON-NLS-1$
```
akurtakov, line 233: Isn't ILog.get() better here?

vogella: It is slightly slower as it uses stackwalker to find the correct bundle. In exception cases, it should not matter but I usually pick the faster one, I can change if your prefer the shorter one.

akurtakov: Explanation makes perfect sense! No need to change anything.

PR #2988, opened 2026-05-15
Migrate NewTypeWizardPage.createType() from SubProgressMonitor to SubMonitor

vogella:
Replace the deprecated `SubProgressMonitor` pattern in `createType()` with the modern `SubMonitor` API. The method previously called `monitor.beginTask()` once and then passed the same raw monitor repeatedly to `Progress.subMonitor()`, which internally creates `SubProgressMonitor` instances. The new code calls `SubMonitor.convert()` once at the entry point and uses `split()` for each unit of work, including the previously bare `monitor.worked(1)` calls. The unused `Progress` import is also removed.


jjohnstn (approved):
Looks good.

PR #2989, opened 2026-05-15
Migrate AddCustomConstructorOperation.run() from SubProgressMonitor to SubMonitor

vogella:
Replace the deprecated `SubProgressMonitor` pattern in `run()` with the modern `SubMonitor` API. The method previously called `beginTask()` and `setTaskName()` separately on the raw monitor and then passed that same monitor to `Progress.subMonitor()`, which internally calls `SubMonitor.convert()` and triggers a second `beginTask()` on the same monitor. The fix replaces both calls with a single `SubMonitor.convert()` at the entry point (passing the task name directly) and uses `split()` for each unit of work. The unused `Progress` import is removed.


jjohnstn (approved):
Looks good.

Issue #2934, opened 2026-04-10
Add proposal to convert a loop to a forEach method


nlisker:
Given
```java
List<String> list = List.of("a");
for (String adj : list) {
    System.out.println(adj);
}
```
Highlighting the loop and pressing ctrl+1 gives the options:
<img width="263" height="156" alt="Image" src="https://github.com/user-attachments/assets/37610b64-fd37-4fc8-906d-75f982193abe" />

I propose to add an option to convert to
```java
list.forEach(s -> System.out.println(s));
```

Note that the `forEach` argument can be an expression, a block, and sometimes a method reference. I don't know if all valid options should be presented, just a block (always valid), or attempt to use an expression and if that's not possible then a block, or something else. In any case, another transformation on the result (ctrl+1) can convert between them.

carstenartur:
We have that already. See https://github.com/carstenartur/sandbox/blob/main/sandbox_functional_converter/README.md

nlisker:
Nice, but it's an external plugin, my suggestion is for the JDT.

carstenartur:
The idea is finding eclipse jdt commiters willing to include it and then include it.

jjohnstn:
@inuyasha82 Interested in looking at porting this code over to JDT?

inuyasha82:
Ok, i'll have a look.

inuyasha82:
@carstenartur i'll start to look at how to integrate code from your project (related to this specific quick assist)  into JDT, if this is ok for you. 

Meanwhile if you can point me to the files that are involved, it will be a great help, thanks. 

Also can you point me to the instruction on how to debug your project, since I think initially could be easier for me to debug your code to find the relevant parts (i'm facing some compilation issues, i'll open an issue on your project)? (probably just configuration stuff)
@jjohnstn you can assign this issue to me.

carstenartur:
> @carstenartur i'll start to look at how to integrate code from your project (related to this specific quick assist)  into JDT, if this is ok for you. 
> 
> Meanwhile if you can point me to the files that are involved, it will be a great help, thanks. 
> 
> Also can you point me to the instruction on how to debug your project, since I think initially could be easier for me to debug your code to find the relevant parts (i'm facing some compilation issues, i'll open an issue on your project)? (probably just configuration stuff)
> @jjohnstn you can assign this issue to me. 

[...]

PR #3002, opened 2026-05-22
2934 enhanced for loop

inuyasha82:
## What it does
It implements #2934 

It propose to convert an enhanced for loop into a .forEach loop.

## How to test
Enter a code similar to the one below:

```java
package test;

import java.util.ArrayList;
import java.util.List;

public class TestForLoop {

			public void testForLoop() {
				List<String> list = List.of("a", "b", "c");
				for (String adj : list) {
				    System.out.println(adj);
				}
			}
		}
}
```

It should display a new option "Convert enhanced `for` loop to `forEach"

## Author checklist


jjohnstn:
Works well.  Please fix up the description as you have only provided the after example.  It would be nice if this could be offered for an iterative or index for loop as well so the user doesn't have to go through 2 assists.  This can be done as a separate issue.  As well, the current convert to enhanced for clean-up should be modified to either add a new check-box to use Collections.forEach if possible or a new cleanup offered or the current cleanup should default to use forEach when a collection is used and the number of statements is lower than some threshhold.  Again, another issue.

inuyasha82:
> Works well. Please fix up the description as you have only provided the after example. It would be nice if this could be offered for an iterative or index for loop as well so the user doesn't have to go through 2 assists. This can be done as a separate issue. As well, the current convert to enhanced for clean-up should be modified to either add a new check-box to use Collections.forEach if possible or a new cleanup offered or the current cleanup should default to use forEach when a collection is used and the number of statements is lower than some threshhold. Again, another issue.

Description updated. 

Yeah I think we can create separate issues for them.

jjohnstn (changes requested):
One issue.  You can't change an enhanced for loop on an array into a forEach statement.  The forEach method only exists on Collections.  You need an additional check and a test that has something like:

```
public void foo(String[] array) {
   for (String x : array) {
       System.out.println(x);
   }
}
```

Once fixed, this is ready to merge.

inuyasha82:
> One issue. You can't change an enhanced for loop on an array into a forEach statement. The forEach method only exists on Collections. You need an additional check and a test that has something like:
> 
> ```
> public void foo(String[] array) {
>    for (String x : array) {
>        System.out.println(x);
>    }
> }
> ```
> 
> Once fixed, this is ready to merge.

shold be fixed now!

jjohnstn (approved):
Looks good.  Don't forget to add a N&N some time in the future.

nlisker:
> The forEach method only exists on Collections.

Technically, it exists on [Iterable](https://docs.oracle.com/en/java/javase/25/docs/api/java.base/java/lang/Iterable.html#forEach(java.util.function.Consumer)), which `Collection` extends.

PR #3005, opened 2026-05-27
Fix extract temp refactoring for a record accessor

### LINKED ISSUE #3004: Local temp refactoring doesn't honor "refactor all" for record accessor
For the following example:

```
record User(String token) {}

class Demo {
  User user = new User("x");

  void test() {
    String a = use(user.token());
    String b = use(user.token());
  }

  String use(String s) {
    return s;
  }
}
```

Selecting the first `user.token()` call and choosing to Source -> Refactoring -> Extract to Local Variable with the option to refactor all uses results in only the first call being replaced while the second call in the initialization of `String b` is not.

This will fix: https://github.com/eclipse-jdtls/eclipse.jdt.ls/issues/3790

jjohnstn:
- fix SideEffectChecker to not return side effects as true for a record accessor
- add new ExtractTempTests16 and add to AllRefactoringTests
- fixes #3004

## What it does

See issue.

## How to test

See issue or new tests.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.manipulation/refactoring/org/eclipse/jdt/internal/corext/refactoring/util/SideEffectChecker.java
+++ b/org.eclipse.jdt.core.manipulation/refactoring/org/eclipse/jdt/internal/corext/refactoring/util/SideEffectChecker.java
@@ -123,7 +123,7 @@ public class SideEffectChecker extends ASTVisitor {
 			fSideEffect= true;
 			return null;
 		}
-		if (!(iTypeBinding.getJavaElement() instanceof IType))
+		if (!(iTypeBinding.getJavaElement() instanceof IType) || methodBinding.isSyntheticRecordMethod())
 			return null;
 		IType it= (IType) (iTypeBinding.getJavaElement());
 		try {
--- a/org.eclipse.jdt.ui.tests.refactoring/resources/ExtractTemp/canExtract16/A_test160_in.java
+++ b/org.eclipse.jdt.ui.tests.refactoring/resources/ExtractTemp/canExtract16/A_test160_in.java
@@ -1,0 +1,17 @@
+// 10, 24 -> 10, 36   AllowLoadtime == false
+package p;
+record User(String token) {
+}
+
+class A {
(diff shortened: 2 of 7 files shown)


javanoo6:
@jjohnstn the fix is not fixing the chains of method calls e.g.
```java
		String a = use(user.token());
		String b = use(user.token().value());
```
the `user.token()` will be extracted only here `String a = use(user.token());` 
but it is supposed to be 

```java
        String token= user.token();
		String a = use(token);
		String b = use(token.value());
```

and now it is

```java
        String token= user.token();
		String a = use(token);
		String b = use(token.token().value());
```

datho7561:
> the fix is not fixing the chains of method calls e.g.

I tried this:

```java
package dev.datho7561.berrycollector;

record User(Token token) {
}

record Token(String value) {
}

class Demo {
	User user = new User(new Token("x"));

	void test() {
		String a = use(user.token());
		String b = use(user.token().value());
		System.out.println(a + b);
	}

	String use(String s) {
		return s;
	}
	
	String use(Token t) {
		return t.value();
	}
}
```

Select the first `user.token()`, activate quickfixes, and accept "Extract to local variable (replace all occurances)", which gives me:

```java
package dev.datho7561.berrycollector;

record User(Token token) {
}

record Token(String value) {
}

class Demo {
	User user = new User(new Token("x"));

	void test() {
		Token token = user.token();
		String a = use(token);
		String b = use(token.value());
		System.out.println(a + b);
	}

	String use(String s) {
		return s;
	}
	
	String use(Token t) {
		return t.value();
	}
}
```

So I think it's working?

datho7561 (approved):
Works good from my testing!

javanoo6:
> Works good from my testing!

i've made build of jtdls for mine nvim from `jjohnstn:recordrefactor` branch;
i'll investigate further whats broken and will hope its not the jtdls anymore, because the bug persists on mine's work project; 
UPD: the fix works, i did rebuild jdtls correctly this time;

@datho7561 @jjohnstn thank you again for the fix and review;

PR #3009, opened 2026-05-29
Use Hash Lookup in OutlineTreeViewer

r-mennig:
When using the Quick Outline for large class hierarchies, UI slowdowns can occur when showing all inherited members of a class.

Sampling shows that a significant amount of time is spent finding the tree items shown in the OutlineTreeViewer.

Sample: [Uploading snapshot-quickOutline-noUseHashLookup-1780056752557.zip…]()

image

This lookup can be speed up by enabling org.eclipse.jface.viewers.StructuredViewer.setUseHashlookup(boolean) on the viewer:

image

## What it does
This PR improves the OutlineTreeViewer's performance by enabling the `org.eclipse.jface.viewers.StructuredViewer.setUseHashlookup(true)` optimization for faster lookups of tree items. The optimization is safe to use here since all `JavaElement`s override `equals()` and `hashCode()`.

This removes the time spent in `StructuredViewer.findItems()`, reducing the viewer's refresh time in this example from 1.4 sec to just 0.5 sec.

image

## How to test
- Unzip and import the test project and navigate to the class C0 [testProject.zip](https://github.com/user-attachments/files/28393539/testProject.zip)
- Select the class name of the C0 class and open the Quick Outline View (Ctrl-O)
- Press Ctrl-O again to show inherited members and type "C99"  into the search field -> The UI starts "stuttering", taking significant time to react to the user input 
QuickOutlineStuttering

_Note:_ The reproduction uses a large generated class with many inner classes to produce enough work for the UI to start stuttering. The problem is however not bound to just inner classes, we can reproduce this problem consistently with very large "organic" class hierarchies in our code base.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/text/AbstractInformationControl.java
+++ b/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/text/AbstractInformationControl.java
@@ -491,7 +491,10 @@ public abstract class AbstractInformationControl extends PopupDialog implements 
 	 * matches the current filter pattern.
 	 */
 	protected void selectFirstMatch() {
-		Object selectedElement= fTreeViewer.testFindItem(fInitiallySelectedType);
+		Object selectedElement = null;
+		if (fInitiallySelectedType != null) {
+			selectedElement= fTreeViewer.testFindItem(fInitiallySelectedType);
+		}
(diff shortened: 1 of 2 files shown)


jjohnstn (approved):
Looks good.  Will have to wait until the repository opens up for 4.41 M1 as we are currently frozen for the 4.40 release.

PR #3011, opened 2026-05-29
Fix inline method to not inline a super call to another type

### LINKED ISSUE #3007: Inline Method refactoring produces uncompilable code when inlined method contains super call
Inline Method refactoring produces uncompilable code when the inlined method contains a `super` call.

Before refactoring, the program compiles and runs successfully. After applying Inline Method to `B.f()` at the call site `b.f()` inside `C.test(B)`, Eclipse replaces the call with `super.g()`. However, the inlining target is class `C`, where `super.g()` is not valid. The refactored code no longer compiles.

Code before refactoring:

```java
package org.example;

public class Main {
        B b = new B();
        new C().test(b);
    }
}

class A {
    int g() {
        return 1;
    }
}

class B extends A {
    int f() {
        return super.g();
    }
}

class C {
    void test(B b) {
        int x = b.f(); // inline method 'f()'
        System.out.println(x);
    }
}
```

Code after refactoring:

```java
package org.example;

public class Main {
        B b = new B();
        new C().test(b);
    }
}

class A {
    int g() {
        return 1;
    }
}

class B extends A {
    int f() {
        return super.g();
    }
}

class C {
    void test(B b) {
        int x = super.g();
        System.out.println(x);
    }
}
```

jjohnstn:
- modify CallInliner.checkMethodDeclaration() to check if the source contains a super call and the type declaration of the source and target are not the same
- add new test to InlineMethodTests
- fixes #3007

## What it does

See issue.

## How to test

See issue or new test.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.manipulation/META-INF/MANIFEST.MF
+++ b/org.eclipse.jdt.core.manipulation/META-INF/MANIFEST.MF
@@ -3,7 +3,7 @@ Automatic-Module-Name: org.eclipse.jdt.core.manipulation
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.jdt.core.manipulation; singleton:=true
-Bundle-Version: 1.24.100.qualifier
+Bundle-Version: 1.24.200.qualifier
 Bundle-Vendor: %providerName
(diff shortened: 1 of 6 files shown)


iloveeclipse:
@jjohnstn , please don’t merge before master is officially open

PR #3016, opened 2026-06-01
Add additional logic to CallHierarchyCore.getCompilationUnitNode()

### LINKED ISSUE #3015: Outgoing call hierarchy for methods in classes in a jar with attached source doesn't work since 4.40
Setup to test:
1. Create an Eclipse Java project
2. Add a jar onto the modulepath as a dependency (I did https://github.com/stleary/JSON-java) along with its sourcepath
3. Open up a class from the dependency and navigate to a method that has method invocations in its body (I created an instance of `JSONStringer` and called `stringer.value(null)`, then Ctrl+Clicked on `value`)
5. Right click on the method name in the method declaration
6. Open Call Hierarchy
7. Switch to "Show Callee Hierarchy"

**Expected:** there is a list of all method calls in the method body
**Actual:** there are zero method calls listed

At some point during 4.40 ibuilds this was working as expected, but after updating to the latest artifacts, it no longer works.

jjohnstn:
- add check for class file to avoid checking for a java file name which won't find the .java extension and will return false
- fixes #3015

## What it does

See issue.

## How to test

Run jdt.ls test mentioned in issue.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/callhierarchy/CallHierarchyCore.java
+++ b/org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/callhierarchy/CallHierarchyCore.java
@@ -28,6 +28,7 @@ import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.core.runtime.NullProgressMonitor;
 import org.eclipse.core.runtime.OperationCanceledException;
 
+import org.eclipse.jdt.core.IClassFile;
 import org.eclipse.jdt.core.IJavaElement;
 import org.eclipse.jdt.core.IMember;
 import org.eclipse.jdt.core.IMethod;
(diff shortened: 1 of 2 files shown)


arcivanov:
Since this regression was introduced by my PR #2881, I've prepared a regression test for this fix. The test creates a class folder with attached source, verifies the member is an `IClassFile` with a non-null buffer, and asserts that the outgoing callee hierarchy returns the expected callees via the Java AST path.

Commit: https://github.com/arcivanov/eclipse.jdt.ui/commit/cc39ae870f

Feel free to cherry-pick or squash it into this PR. I couldn't push directly to your fork.

datho7561 (changes requested):
This fixes the problem for me.

I think it'd be helpful to add the regression test that Arcadiy suggested.

jjohnstn:
> This fixes the problem for me.
> 
> I think it'd be helpful to add the regression test that Arcadiy suggested.

Will do.

Issue #2675, opened 2025-11-29
Create a new "Add Test Source Folder" item


laeubi:
Currently one can already create a source folder by right click on the project > new > source folder.

But "converting" this into a folder for test sources folder is quite cumbersome:

1. One has to go to project > build path > configure build path
2. Then one has to enable the "Allow output folders for source folders" (??? maybe a typo and should be "Allow individual output folders for source folders")
3. then one has to navigate to the new source folder and define a new output folder
4. then one finally can toggle the "contains test sources"

This can be much more streamlined by have a new item "Add Test Source Folder" that create a new source folder and then:

- enable the option from step 2
- use the default output folder name and append `_test` and set this as the output folder from step 3
- sets the test attribute from step 4

That way it would become much more easier to setup such source folders containing test.

jjohnstn:
@inuyasha82 Would you be interested in looking at this?

inuyasha82:
Sure @jjohnstn please assign it to me.

inuyasha82:
PR #3022 has been merged. It will be available on the next release. 
@jjohnstn can this issue be closed?

jjohnstn:
Closed as completed by #3022

laeubi:
Many thanks for implementing this feature!

PR #3022, opened 2026-06-04
2675 test src folder

inuyasha82:
## What it does
It implements feature request in issue #2675 

## How to test

On a java perspective:

* right click on a project item from project explorer:
* Select new
* A new item: `Test Source Folder` should be visible
* Fill the folder name, and press finish.

Now go to Project Properties -> Java Build Path -> Source  and then select the newly created folder, it should show the customized folder: "test_bin" as `Output Folder`

## Author checklist


jjohnstn (changes requested):
Looks good, but I noticed that for the test folder, the "contains test sources" is set to no.  According to the issue, they want this set too.

inuyasha82:
> Looks good, but I noticed that for the test folder, the "contains test sources" is set to no. According to the issue, they want this set too.

Done!

jjohnstn (changes requested):
Thanks for making the last change.

One last comment.  Opening the test folder source page, the bottom of the page has a mention of creating a default bin directory (in my case test2/bin).  This bin directory was already created as part of creating the default src directory.  Should this checkbox be disabled or not shown??  I am wondering if it will cause issues if someone messes with it after the fact.

inuyasha82:
> Thanks for making the last change.
> 
> One last comment. Opening the test folder source page, the bottom of the page has a mention of creating a default bin directory (in my case test2/bin). This bin directory was already created as part of creating the default src directory. Should this checkbox be disabled or not shown?? I am wondering if it will cause issues if someone messes with it after the fact.

In theory this change is creating a bin_test folder as output folder. It shouldn't create an extra bin folder. 
This behavior was request in the original issue. 

* use the default output folder name and append _test and set this as the output folder from step 3

Screenshot From 2026-06-30 11-48-52

So in theory what it should create is the "bin_test" folder and don't touch the original one. 

Can you confirm that @jjohnstn

jjohnstn:
It is not an issue.

jjohnstn (approved):
Approved.

Issue #3029, opened 2026-06-10
Package Explorer View filter "Libraries from External" not filtering folder plug-ins in Eclipse 4.40


Phillipus:
In Eclipse 4.40 in the Package Explorer I have the filter "Libraries from external" checked. 

<img width="241" height="377" alt="Image" src="https://github.com/user-attachments/assets/87c398c0-4229-4695-8feb-0ed5c659c94d" /><br>

<br>

All plug-ins are hidden in the "Plug-in Dependencies" node except for one - "com.sun.jna": 

<img width="1476" height="197" alt="Image" src="https://github.com/user-attachments/assets/3a5cb6b1-2c23-4bfd-95cd-531d6096d45d" /><br>

The difference with this plug-in is that it's a folder not a jar file.

This was working in Eclipse 4.39 where all plug-ins were hidden.

Phillipus:
My theory is that before now folder packages were not included in the "Plug-in Dependencies" node but are now and they are shown because of returning true in [`org.eclipse.jdt.internal.ui.filters.LibraryFilter`](https://github.com/eclipse-jdt/eclipse.jdt.ui/blob/master/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/filters/LibraryFilter.java) when it should return false if `IPackageFragmentRoot` is not an archive:

```java
public class LibraryFilter extends ViewerFilter {

	@Override
	public boolean select(Viewer viewer, Object parentElement, Object element) {
		if (element instanceof IPackageFragmentRoot) {
			IPackageFragmentRoot root= (IPackageFragmentRoot)element;
			if (root.isArchive()) {
				// don't filter out JARs contained in the project itself
				IResource resource= root.getResource();
				if (resource != null) {
					IProject jarProject= resource.getProject();
					IProject container= root.getJavaProject().getProject();
					return container.equals(jarProject);
				}
				return false;
			}
			return false; // Edited - Should return false here
		} else if (element instanceof ClassPathContainer.RequiredProjectWrapper) {
			return false;
		}
		return true;
	}
}
```

Testing the above fixes the issue.

jjohnstn:
This should be fixed now.  Closing.

PR #3032, opened 2026-06-12
Fix Package Explorer View filter not filtering non-archive plug-ins

Phillipus:
## What it does
When "Libraries from External" is filtering in Package Explorer it should also filter plug-ins that are not in archive format. This is not happening. This fixes the condition to return false for the filter if the package is a folder so that it is not shown in Package Explorer.

## How to test
See https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/3029 for steps to reproduce

## Author checklist

### DIFF
--- a/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/filters/LibraryFilter.java
+++ b/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/filters/LibraryFilter.java
@@ -32,18 +32,15 @@ public class LibraryFilter extends ViewerFilter {
 
 	@Override
 	public boolean select(Viewer viewer, Object parentElement, Object element) {
-		if (element instanceof IPackageFragmentRoot) {
-			IPackageFragmentRoot root= (IPackageFragmentRoot)element;
-			if (root.isArchive()) {
-				// don't filter out JARs contained in the project itself
-				IResource resource= root.getResource();
-				if (resource != null) {
-					IProject jarProject= resource.getProject();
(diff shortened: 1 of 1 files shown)


org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/filters/LibraryFilter.java
```diff
@@ -44,6 +44,7 @@ public boolean select(Viewer viewer, Object parentElement, Object element) {
 				}
 				return false;
 			}
+			return false;
```
jjohnstn, line 43: I did a test where I had a class folder in the plugin project (I just copied over org.eclipse.text.tests from my bundle pool folder).  I then added a classpath class  folder entry in the class path.  Long story short, this logic removed the Referenced Libraries since it removed the folder without checking if it was in the project.  I made a local change to remove the check for if (root.isArchive()) above and just do the resource project check logic within.  With that change, it again shows the referenced library in the Reference Libraries.  If you want, I can push the change or you can do it if you agree with it.

Phillipus: I see what you mean. I removed the check for `if (root.isArchive())` and force pushed the change.

jjohnstn: Thanks.  Looks good.

jjohnstn (approved):
Looks good.

PR #3052, opened 2026-06-30
Only fold inner types when preference is set #3031

### LINKED ISSUE #3031: Fold "Inner types" option also affects folding of top level types
Since the 3.20 release, top-level classes are now folded, which I don't think is a good default. If I open a Java source file, it's more likely to be a single type than many, and all I see now is a single line with a class declaration. The whole point of opening a file is to see its contents. Adding an extra click. There is a default keyboard binding to toggle the folding, but it binds to a numpad key, which is not commonly available on laptop keyboards. All in all, the file opening experience has severely degraded with the new default folding.

Please make top-level class file folding an option and (optionally) revert the default to not fold.

Original change that introduced the new behavior: https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2860

fedejeanne:
## What it does
Avoids folding outer types if the preference **Initially fold these elements > Inner Types** is checked.

Fixes https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/3031

## How to test
Use this class:

```java
package plugin;

class B {
	static class B1 {

	}
}

public class A {
	static class A1 {
	}

	static class A2 {
			Runnable r = new Runnable() {
				// Should be folded
				@Override
				public void run() {

				}
			};
		}

	}
}
``` 

Activate/Deactivate the preference, it should look like this

image

---

image

## Author checklist

### DIFF
--- a/org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/folding/FoldingTest.java
+++ b/org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/folding/FoldingTest.java
@@ -784,4 +784,47 @@ public class FoldingTest {
 		FoldingTestUtils.assertContainsRegionUsingStartAndEndLine(regions, str, 9, 10); // if
 		FoldingTestUtils.assertContainsRegionUsingStartAndEndLine(regions, str, 11, 13); // else
 	}
+
+	/**
+	 * See <a href="https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/3031">GitHub Issue #3031</a>
+	 */
+	@Test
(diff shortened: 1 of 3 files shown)


iloveeclipse:
LGTM, but would it be possible to add test to the `FoldingTestSuite`?

r-mennig (approved):

fedejeanne:
> LGTM, but would it be possible to add test to the `FoldingTestSuite`?

I added `org.eclipse.jdt.text.tests.folding.FoldingTest.testFoldInnerTypesByDefault()`

iloveeclipse (approved):

iloveeclipse:
Thanks!

PR #3058, opened 2026-07-02
Skip organize imports on save while the type indexer is busy

vogella:
When Organize Imports is enabled as a save action, resolving missing imports searches the type index with WAIT_UNTIL_READY_TO_SEARCH. Saving right after startup, a branch switch or a large build therefore blocks until indexing finishes, which regularly exceeds the 2 second threshold and greets the user with the "Slow Save Actions" warning dialog.

This adds OrganizeImportsOperation#setSkipWhenIndexerBusy, which makes the type search use CANCEL_IF_NOT_READY_TO_SEARCH and produce no edit while the indexer is busy. The save actions enable it, so saving stays fast and a missing import is simply added on the next save once the index is ready. The Clean Up wizard and the manual Organize Imports action are unchanged and still wait for the index.


vogella:
This happens all the time at my current client: 

image

PR #3062, opened 2026-07-04
Fix add import operation to change all references in file if possible

### LINKED ISSUE #3050: Strg-Shift-M should create static import for all occurences in current file
Imagine I have a file, containing 10 occurences of Math.min(a,b). 

Now I press Strg-Shift-M on the first occurence and behold! The code is more readable now! A single call to Math.min(a,b) is now min(a,b). 

But now I have to find all other calls to the function and have to static-import them also, althought there is already a min import and obviously on name clashes can exist anymore in my file (except for inner classes, that also defined their min() function). 

Now, since that is out of the way, what reason could there be to not automatically find all other occurences of my import and import them also? 

I beg you, someone, to implement this feature, it would make that feature so much better. IF you want to go the extra mile, you could even evaluate where in the file other occurences would now have changed to another min function in the rare scenario I mentioned in the beginning, but else there is absolutely no reason not to implement this. 

Thanks in advance,
Daniel Migowski

jjohnstn:
- modify AddImportsOperation.evaluateEdits() method to try and find all uses of imported item and remove qualifier instead of just changing a single item
- modify AddImportTest
- fixes #3050

## What it does

See issue.

## How to test

See issue or modified tests.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.ui.tests/ui/org/eclipse/jdt/ui/tests/core/AddImportTest.java
+++ b/org.eclipse.jdt.ui.tests/ui/org/eclipse/jdt/ui/tests/core/AddImportTest.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2025 IBM Corporation and others.
+ * Copyright (c) 2000, 2026 IBM Corporation and others.
  *
  * This program and the accompanying materials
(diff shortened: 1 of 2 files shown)


dmigowski:
I didn't know Strg-1 also provides this, Google wasn't helpful here. Thanks alot.

PR #3063, opened 2026-07-06
Fix Missing Markdown Template contexts and MD Type comment generation

SougandhS:
This commit includes missing Markdown code templates and fixes unwanted "///" generating on types with empty params
Required for https://github.com/eclipse-jdtls/eclipse.jdt.ls/pull/3837

## Author checklist

### DIFF
--- a/org.eclipse.jdt.ui/templates/default-codetemplates.xml
+++ b/org.eclipse.jdt.ui/templates/default-codetemplates.xml
@@ -76,6 +76,8 @@
 <template name="markdowndelegatecomment" id="org.eclipse.jdt.ui.text.codetemplates.markdowndelegatecomment" description="%CodeTemplates.markdowndelegatecomment" context="markdowndelegatecomment_context" enabled="true">/// ${tags}
 /// ${see_to_target}</template>
  
+ <template name="markdownoverridecomment" id="org.eclipse.jdt.ui.text.codetemplates.markdownoverridecomment" description="%CodeTemplates.markdownoverridecomment" context="markdownoverridecomment_context" enabled="true">/// ${see_to_overridden}</template>
+ 
  <template name="newtype" id="org.eclipse.jdt.ui.text.codetemplates.newtype" description="%CodeTemplates.newfile" context="newtype_context" enabled="true">${filecomment}
 ${package_declaration}
 
--- a/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/preferences/CodeTemplateBlock.java
+++ b/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/preferences/CodeTemplateBlock.java
@@ -230,6 +230,8 @@ public class CodeTemplateBlock extends OptionsConfigurationBlock {
 						return 14;
 					case CodeTemplateContextType.MARKDOWNMETHODCOMMENT_ID:
 						return 15;
+					case CodeTemplateContextType.MARKDOWNOVERRIDECOMMENT_ID:
+						return 16;
 					case CodeTemplateContextType.MARKDOWNDELEGATECOMMENT_ID:
 						return 17;
 					case CodeTemplateContextType.MARKDOWNGETTERCOMMENT_ID:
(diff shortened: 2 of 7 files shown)


SougandhS:
Hi @jjohnstn, 
Could you please check this PR when you get some time ?

jjohnstn (changes requested):
I don't understand why you are creating markdown comment template ids for body comments and catch block comments.  These aren't normally specified via Javadoc so don't need markdown support.  Currently, they are under the templates code section and don't appear in comments and markdown sections.

I think adding an override markdown makes sense because there is a way to specify the normal Javadoc version via the preferences.  You will need to add this capability to the templates page under markdown (it can be equal to the current template under the comments section).

For the change in type comment behavior, you need to add a test case.

SougandhS:
> don't understand why you are creating markdown comment template ids for body comments and catch block comments. These aren't normally specified via Javadoc so don't need markdown support. Currently, they are under the templates code section and don't appear in comments and markdown sections.

Removed those 👍 

>I think adding an override markdown makes sense because there is a way to specify the normal Javadoc version via the preferences. You will need to add this capability to the templates page under markdown (it can be equal to the current template under the comments section).

Handled override in other places 👍 

>For the change in type comment behavior, you need to add a test case.

Added a test 👍

org.eclipse.jdt.ui/templates/default-codetemplates.xml
```diff
@@ -76,6 +76,9 @@
 <template name="markdowndelegatecomment" id="org.eclipse.jdt.ui.text.codetemplates.markdowndelegatecomment" description="%CodeTemplates.markdowndelegatecomment" context="markdowndelegatecomment_context" enabled="true">/// ${tags}
 /// ${see_to_target}</template>
  
+ <template name="markdownoverridecomment" id="org.eclipse.jdt.ui.text.codetemplates.markdownoverridecomment" description="%CodeTemplates.markdownoverridecomment" context="markdownoverridecomment_context" enabled="true">/// ${tags}
```
jjohnstn: Remove the $tags.  For a sample hashcode() addition, it adds an empty return comment to be filled in by the end-user, but the truth is the original via $see_to_overridden fills in all the info needed.

org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/preferences/CodeTemplateBlock.java
```diff
@@ -238,6 +238,8 @@ public int category(Object element) {
 						return 19;
 					case CodeTemplateContextType.MARKDOWNMODULECOMMENT_ID:
 						return 20;
+					case CodeTemplateContextType.MARKDOWNOVERRIDECOMMENT_ID:
+						return 21;
```
jjohnstn: Make this 16 so markdown override appears in the same spot in the category as it does for the Comments section.

jjohnstn:
Hi @SougandhS I see you pushed after my latest comments.  I don't see either issue addressed.  Is it possible you didn't include a latest commit in the push?

SougandhS:
@jjohnstn Can you check again ? - just rebased now
I have removed the catch body, method body comments and handled override markdown commnt in other places like you suggested, and included a test for markdown type comment issue..

image

Is there anything missing or do I need to handle anything in one more place ?

jjohnstn:
> @jjohnstn Can you check again ? - just rebased now I have removed the catch body, method body comments and handled override markdown commnt in other places like you suggested, and included a test for markdown type comment issue..

> Is there anything missing or do I need to handle anything in one more place ?

@SougandhS Yes.  I saw you did all of that and did a second review with 2 comments.  See above in this PR or comments in the code section.

SougandhS:
> @SougandhS Yes. I saw you did all of that and did a second review with 2 comments. See above in this PR or comments in the code section.

I think there may have been a technical glitch. I'm not able to see the two comments you're referring to in the code review, and I also didn't receive any email notifications for them. If possible, could you please re-add those comments so I can address them?

jjohnstn (changes requested):
Looks much better.  A few nit comments and this can be merged.

jjohnstn:
> > @SougandhS Yes. I saw you did all of that and did a second review with 2 comments. See above in this PR or comments in the code section.
> 
> I think there may have been a technical glitch. I'm not able to see the two comments you're referring to in the code review, and I also didn't receive any email notifications for them. If possible, could you please re-add those comments so I can address them?

Weird.  I saw them but it said they were pending.  I guess the submit review button didn't register.  I just resubmitted.

org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/preferences/CodeTemplateBlock.java
SougandhS: done

org.eclipse.jdt.ui/templates/default-codetemplates.xml
SougandhS: removed 👍

jjohnstn (approved):
Looks good.  Thanks.

SougandhS:
Thanks @jjohnstn

PR #3072, opened 2026-07-09
Add new option for newline on annotation record parameters

inuyasha82:
## What it does
It implement changed discussed in #4152  UI part
To be merged only after https://github.com/eclipse-jdt/eclipse.jdt.core/pull/5184 will be merged into jdt.core.

image

To be merged after PR  https://github.com/eclipse-jdt/eclipse.jdt.core/pull/5184

## How to test
If the new setting Newline -> After Annotations -> On record parameter is enabled it should format the following code snippet: 

```java
public record TestRecord(
        @Deprecated @Native @SuppressWarnings(value = { "" }) String name, @SuppressWarnings(value = { "" }) String age
	) {}
```

to: 
```java
public record TestRecord(
		@Deprecated
		@Native
		@SuppressWarnings(value = {
				"" })
		String name,
		@SuppressWarnings(value = { "" })
		String age) {
}
```

## Author checklist


jjohnstn:
@inuyasha82 Quick question.  Does this new option only apply to a record declaration or does it apply also to the new instance call?  If not, then change the text for the setting to be "record components" rather than "record parameters".

inuyasha82:
From what I understood, is not possible to pass annotation inside the constructor call 
```java
TestRecord tr = new TestRecord(@Annotation "example");
```
and same with methods. Is higlighted as an error. 

So I think that it can't be done, hence it applies only to the record declaration. (but correct me if i'm wrong), then i'm going to change the wording.

jjohnstn:
Thanks @inuyasha82

PR #3077, opened 2026-07-11
Fix inline of method with instantiation of local class

### LINKED ISSUE #3070: Inline Method incorrectly qualifies local class instantiation and produces uncompilable code
Before refactoring, the program compiles and runs successfully. After applying Inline Method to `I02.f()` at the call site `obj.f()` inside the static method `main`, Eclipse moves the local class `H` into `main`. However, it also changes the local class instantiation from:

```java
new H().g()
```

to:

```java
obj.new H().g()
```

This is invalid because `H` is a local class declared inside a method, not a member inner class of `I02`. A local class should be instantiated with `new H()`, not with `obj.new H()`.

Code before refactoring:

```java
package org.example;

public class I02 {
        I02 obj = new I02();

        // Refactoring operation: Inline Method
        // Target invocation: obj.f()
        int result = obj.f();

        System.out.println(result);
    }

    int f() {
        class H {
            int g() {
                return 7;
            }
        }

        return new H().g();
    }
}
```

Code after refactoring:

```java
package org.example;

public class I02 {
        I02 obj = new I02();

        class H {
            int g() {
                return 7;
            }
        }

        // Refactoring operation: Inline Method
        // Target invocation: obj.f()
        int result = obj.new H().g();

        System.out.println(result);
    }

    int f() {
        class H {
            int g() {
                return 7;
            }
        }

        return new H().g();
    }
}
```

Expected behavior:

The refactoring should preserve compilability. A valid result would instantiate the local class directly:

```java
class H {
    int g() {
        return 7;
    }
}

int result = new H().g();
```

jjohnstn:
- modify SourceAnalyzer.visit(ClassInstanceCreation) to only mark a member class as a receiver
- add new tests to InlineMethodTests
- fixes #3070

## What it does

See issue.

## How to test

See issue or new tests.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/refactoring/code/SourceAnalyzer.java
+++ b/org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/refactoring/code/SourceAnalyzer.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2024 IBM Corporation and others.
+ * Copyright (c) 2000, 2026 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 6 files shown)


jjohnstn:
Compiler issue failure appears to be a new deprecation that says 2026-09.  It is not in any code that was changed as part of this PR.

PR #3090, opened 2026-07-22
Add Wizard support for creating Compact Source Files

SougandhS:
Introduces a new "Compact" entry in the New Java wizard to create implicit class source files (JEP 463, Java 25+). The wizard generates a file with a top-level void main() method, similar to what other IDEs have.
image

Screenshot 2026-07-22 at 5 15 31 PM

Screenshot 2026-07-22 at 5 16 10 PM

## Author checklist

### DIFF
--- a/org.eclipse.jdt.ui.tests/ui/org/eclipse/jdt/ui/tests/wizardapi/NewTypeWizardTest25.java
+++ b/org.eclipse.jdt.ui.tests/ui/org/eclipse/jdt/ui/tests/wizardapi/NewTypeWizardTest25.java
@@ -1,0 +1,136 @@
+/*******************************************************************************
+ * Copyright (c) 2026 IBM Corporation and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
+ *******************************************************************************/
+
+package org.eclipse.jdt.ui.tests.wizardapi;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+
+import java.util.Hashtable;
+
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+
+import org.eclipse.jdt.testplugin.JavaProjectHelper;
+import org.eclipse.jdt.testplugin.StringAsserts;
+import org.eclipse.jdt.testplugin.TestOptions;
+
+import org.eclipse.jface.preference.IPreferenceStore;
+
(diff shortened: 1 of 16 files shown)


SougandhS:
Hi @jjohnstn, could you please check this PR when you get some time ?

jjohnstn (changes requested):
Some general usage issues.  First of all, the new Compact is not offered from the New menu in the Java perspective.  It is offered in the Plug-in Development perspective.  Secondly, it creates an invalid class if offered for a project with a module-info.java file that contains a named module.  In such a case, the compiler complains if you try to add it to the default package and also complains if you add it to an existing package.  It should be disabled for the cases where there is a named module in the module-info.java file and should also be disabled when New is run from a package.

SougandhS:
> Some general usage issues. First of all, the new Compact is not offered from the New menu in the Java perspective. It is offered in the Plug-in Development perspective. Secondly, it creates an invalid class if offered for a project with a module-info.java file that contains a named module. In such a case, the compiler complains if you try to add it to the default package and also complains if you add it to an existing package. It should be disabled for the cases where there is a named module in the module-info.java file and should also be disabled when New is run from a package.

Will look into this 👍

SougandhS:
>In such a case, the compiler complains if you try to add it to the default package and also complains if you add it to an existing package. It should be disabled for the cases where there is a named module in the module-info.java file 

Added warning for module projects 👍 
Screenshot 2026-07-25 at 4 22 05 PM

_________________
>disabled when New is run from a package.

For this I had excluded package input field - so even if it is created from a package it will only be created in default package..

https://github.com/user-attachments/assets/db75aa86-5f06-4929-bd5b-c07bd28e2e28

_________________
>First of all, the new Compact is not offered from the New menu in the Java perspective. It is offered in the Plug-in Development perspective.

I think there's some bug in perspective caching area, when I tried this on a new workspace it is showing everywhere. The identifier of compact wizard is already added in perspective layouts..
I noticed for one workspace it was not showing in package explorer and File->new but available in project explorer, then I tried reset perspective and it fixed.
Can u try by resetting ur current workspace's perspective or try in a new workspace ?


https://github.com/user-attachments/assets/daebacf7-ed21-4da2-95ce-d8bbf085bc83

jjohnstn:
Hi @SougandhS Resetting the perspective solved the problem for the Package Explorer.  I checked what New does for records on projects that don't support records and it is similar to your logic where the New record is still offered but shows the error when the wizard gets filled in more.

jjohnstn (changes requested):
Everything now works in the perspectives and the check for module or JVM works good, but there is a bit of a glitch when a modifier is selected.  I am seeing:

```
public ///
void main() {

}
```

Note how there is a markdown comment added between public and void.  This may result in a separate issue being opened, but you need to check.

SougandhS:
>Note how there is a markdown comment added between public and void. This may result in a separate issue being opened, but you need to check.

Fixed this now 👍

jjohnstn (approved):
Approved.

SougandhS:
Thanks @jjohnstn, will include in N&N 👍

PR #3095, opened 2026-07-29
Remove duplicate link with editor from menu bar in package explorer

elsazac:
## What it does

This change removes the Link with Editor option from the view menu in package explorer.
The same action is already available in the toolbar, so having it in both places is unnecessary. This also makes the view consistent with the other toolbar actions, which do not have duplicate entries in the view menu.

Before :
image

After :
image

## Author checklist


SougandhS:
Package explorer already has this option in its toolbar area, so users can still find and use it quickly
image

but for other implementations eg. Java Browsing, Java Outline view, Type hierarchy etc.. Link with editor is only available as a context option
image

How about including this as a toolbar option for all, and remove this from context, so users can quickly access this like in Project explorer and Package explorer?

WDYT @iloveeclipse @jjohnstn @noopur2507 ?

SougandhS (approved):
LGTM 👍

PR #3107, opened 2026-08-07
Asynchronously open internal browser for javadoc hover

### LINKED ISSUE #3105: Clicking javadoc web link freezes Eclipse when using internal browser
If Eclipse is configured to use an internal web browser (via the Web Browser -> Use internal web browser) preference, clicking on a web link in a javadoc popup reliably causes Eclipse to permanently freeze.

### Reproduction

This problem can be reproduced on the latest I-Build and from within a runtime workspace on the latest master commit (dd2f4198d1a25a5cd7b41e5a6f2dd55b19b17691).

1. Set browser preference to use internal web browser 

Image

2. Inside any Java project, navigate to `java.util.List.of()` (e.g. via "Open Type" search)

3. Hover the method and click on the hyperlink
Image

### Current behavior
The user interface freezes permanently

### Expected behavior
An internal browser tab opens, showing the linked documentation

### Further information 

This issue seems closely related to https://github.com/eclipse-mylyn/org.eclipse.mylyn/issues/1208 and https://github.com/eclipse-platform/eclipse.platform.ui/issues/3104. In both cases, the issue tracks back to opening the internal browser synchronously from inside an OS callback (as pointed out [here](https://github.com/eclipse-mylyn/org.eclipse.mylyn/issues/1208#issuecomment-5204938102)). 

This also seems to be the problem in this case:

Image

Image

r-mennig:
Currently, when opening an external link in a javadoc hover in the internal browser under Windows, the user interface permanently freezes.

This is caused by a deadlock within the Edge/WebView2 browser runtime which occurs if a new browser instance is being created from within a (navigation) callback of another browser instance.

Adapting `handleExternalLink(URL, Display)` in the link handler used by the `JavadocHover` to open the new browser instance asynchronously fixes the problem as the browser creation is now moved out of the (navigation) callback.

Fixes https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/3105

### DIFF
--- a/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/actions/OpenBrowserUtil.java
+++ b/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/actions/OpenBrowserUtil.java
@@ -29,7 +29,7 @@ import org.eclipse.jdt.internal.ui.JavaPlugin;
 public class OpenBrowserUtil {
 
 	/**
-	 * Opens the given url in the browser as choosen in the preferences.
+	 * Opens the given url in the browser as chosen in the preferences.
 	 *
 	 * @param url the URL
 	 * @param display the display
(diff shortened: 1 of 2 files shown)


HeikoKlare (approved):
I have tested the proposal with the scenario we have identified in https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/3105 as well as other navigation scenarios in the JavadocHover (navigating to other Javadoc pages, having an external browser preferred via prefereces). I have tested with Windows and macOS.

The change works fine for me on both systems and in all scenarios, and it particularly resolves the UI freeze issue on Windows.
It is also conceptually sound to me, as opening a browser view does not need to happen synchronous from a URL navigation callback inside a view but may reasonably be scheduled for asynchronous execution.

iloveeclipse (approved):

PR #3115, opened 2026-08-13
Add ambiguous import checking for compact source

### LINKED ISSUE #3111: Organize Imports save action deletes required imports in Source File programs
When the _Organize Imports_ clean up is enabled as save action (`Java` -> `Editor` -> `Save Actions` -> `Organize imports`) and is applied to a JEP 330/458 Source-Code Program, it deletes required imports:
```
import java.security.cert.Certificate;

Certificate certificate;

	Certificate certificate;
}
```
In that case the `java.security.cert.Certificate` import is also deleted if there is only a Certificate field or a Certificatevariable, not only with both.

For the corresponding ordinary class the import persists as required:
```
import java.security.cert.Certificate;

public class MyClass {
	Certificate certificate;

		Certificate certificate;
	}
}
```

jjohnstn:
- add a new OrganizeImportsOperation.typeNameAmbiguousForImplicitModule method
- if an import matches one of the implicit package names, verify it is not ambiguous for the implicit module using new method
- add new test to ImportOrganizeTest25
- fixes #3111

## What it does

Add ambiguity checking for imports in compact source.

## How to test

See issue or new test.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.manipulation/common/org/eclipse/jdt/core/manipulation/OrganizeImportsOperation.java
+++ b/org.eclipse.jdt.core.manipulation/common/org/eclipse/jdt/core/manipulation/OrganizeImportsOperation.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2025 IBM Corporation and others.
+ * Copyright (c) 2000, 2026 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -56,6 +56,7 @@ import org.eclipse.jdt.core.dom.IBinding;
 import org.eclipse.jdt.core.dom.IModuleBinding;
 import org.eclipse.jdt.core.dom.IPackageBinding;
 import org.eclipse.jdt.core.dom.ITypeBinding;
+import org.eclipse.jdt.core.dom.ImplicitTypeDeclaration;
 import org.eclipse.jdt.core.dom.ImportDeclaration;
 import org.eclipse.jdt.core.dom.MethodInvocation;
 import org.eclipse.jdt.core.dom.Modifier;
@@ -323,7 +324,8 @@ public class OrganizeImportsOperation implements IWorkspaceRunnable {
 			while (parent instanceof Type) {
 				parent= parent.getParent();
 			}
-			if (parent instanceof AbstractTypeDeclaration && parent.getParent() instanceof CompilationUnit) {
+			if ((parent instanceof AbstractTypeDeclaration && parent.getParent() instanceof CompilationUnit)
+					|| (parent instanceof ImplicitTypeDeclaration && parent.getParent() instanceof CompilationUnit)){
 				return true;
 			}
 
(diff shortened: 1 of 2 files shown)


org.eclipse.jdt.core.manipulation/common/org/eclipse/jdt/core/manipulation/OrganizeImportsOperation.java
```diff
@@ -396,6 +398,25 @@ public void add(SimpleName ref) {
 			fUnresolvedTypes.put(typeName, new UnresolvedTypeData(ref));
 		}
 
+		private boolean typeNameAmbiguousForImplicitModule(String typeName) {
+			IJavaProject project= fCurrPackage.getJavaProject();
+			IType foundType= null;
+			for (String packageName : fImplicitImports) {
```
HannesWell, line 404: I don't know the exact size of this collection, but if it's larger, could this loop become a performance problem?
In general, doesn't `fImplicitImports` only contain packages from the `java.base` module (in case of a Source file program)? Or are there other means to define implicit-imports?
If it's only `java.base`, I assume the list of potentially ambiguous simple names is quite static and could be a Set cached per VMInstall, couldn't it?

jjohnstn: I didn't notice any lag when I performed your test.  Perhaps if someone has thousands of imports for a compact source where all of them are in java.base (hightly unlikely).  Yes, it could be a static list - perhaps using AI to generate?

jjohnstn: I used AI to generate the list of duplicate java.base classes but the static list solution would have to handle changes to java.base from release to release.  For example, running AI on Java 21, there are 85 duplicate classes, but in Java 25, this drops to 65 classes (they removed the bundled ASM library).  This potentially means a list per JVM from Java 25 on.

In my opinion, a compact source file is used by new users or to do a simple quick test.  I don't imagine these classes are going to have many imports, if any.  I think we should leave it dynamic and be open to creating static lists if there are real users running into performance issues.   A similar issue exists for non-compact source files and organize imports but no one seems to have run into a situation yet where the check for ambiguous classes is causing any noticeable lag.

HannesWell: > I used AI to generate the list of duplicate java.base classes but the static list solution would have to handle changes to java.base from release to release. For example, running AI on Java 21, there are 85 duplicate classes, but in Java 25, this drops to 65 classes (they removed the bundled ASM library).

I assumed that it would be possible to compute the list at runtime once, e.g. when "loading" a VM-install by iterating all classes in the `java.base` module and collecting all with same simple name. No AI would be involved and the list should always be up to date.

But if it's probably not a real performance problem, then I'm totally fine with it too. Just wanted to made it sure, this has been cleared.

Thanks for your work on this.

Btw. if you are interested for some users:
I'm using source file programs also to create Java scripts for RelEng task:
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/tree/master/scripts/releng

Currently there are only three, but I intend to add more.

PR #3116, opened 2026-08-14
skip folding implicitly declared classes

danthe1st:
## What it does

[Compact Source Files](https://openjdk.org/jeps/512) contain an implicit class which not present in the source code However, this such a class may contain a folding region which can be collapsed.

As folding classes not present in the source code doesn't make sense, this PR skips them in the folding logic.

## How to test
Create the following simple main class in a project using JDK 25+ (the line break at the end is important):
```java
import module java.base;

int i;
int j;

void main() {

}

```

There should be no folding region next to `int i`; that contains the entire class (the screenshot shows the folding region that shouldn't be there):
image

I hope the `ProjectTestSetup` will make sure JDK25+ is used in the test.

CC @HannesWell because you reported this in the devcall (by the way, the idea about skipping top level types on `Collapse All` may not be as easy as I thought because that logic is part of `ProjectionViewer`/Eclipse Platform while Java folding is JDT-UI so this might need adding new methods to the API)

## Author checklist

### DIFF
--- a/org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/folding/FoldingTest.java
+++ b/org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/folding/FoldingTest.java
@@ -15,6 +15,7 @@ package org.eclipse.jdt.text.tests.folding;
 
 import static org.junit.Assume.assumeFalse;
 import static org.junit.Assume.assumeTrue;
+import static org.junit.jupiter.api.Assertions.assertEquals;
 
 import java.util.List;
 
@@ -30,7 +31,6 @@ import org.junit.runners.Parameterized.Parameters;
 import org.eclipse.jdt.testplugin.JavaProjectHelper;
 
 import org.eclipse.jface.preference.IPreferenceStore;
-
 
 import org.eclipse.jdt.core.IJavaProject;
 import org.eclipse.jdt.core.IPackageFragment;
@@ -825,6 +825,25 @@ public class FoldingTest {
 		FoldingTestUtils.assertContainsCollapsedRegionUsingStartAndEndLine(regions, str, 12, 17); // Runnable
 		JavaPlugin.getDefault().getPreferenceStore().setToDefault(PreferenceConstants.EDITOR_FOLDING_INNERTYPES);
 
+	}
+
+	@Test
+	public void testSimpleMainDoesNotContainRegionSurroundingImplicitClass() throws Exception {
+		String str = """
+				import module java.base;
+
+				int i;
+				int j;
+
+				void main() {
+					// main
+				}
+
+				""";
+		List<FoldingTestUtils.ProjectionRegion> regions= FoldingTestUtils.getProjectionRangesOfPackage(packageFragment, str);
+		FoldingTestUtils.assertContainsRegionUsingStartAndEndLine(regions, str, 5, 7); // main
+		FoldingTestUtils.assertDoesNotContainRegionUsingStartLine(regions, str, 2); // int i
+		assertEquals(1, regions.size()); // no other unexpected regions anywhere else
 	}
 }
 
--- a/org.eclipse.jdt.ui/ui/org/eclipse/jdt/ui/text/folding/DefaultJavaFoldingStructureProvider.java
+++ b/org.eclipse.jdt.ui/ui/org/eclipse/jdt/ui/text/folding/DefaultJavaFoldingStructureProvider.java
@@ -1258,6 +1258,13 @@ public class DefaultJavaFoldingStructureProvider implements IJavaFoldingStructur
 			case IJavaElement.TYPE:
 				// only inner types may be collapsed
 				collapse= ctx.collapseInnerTypes() && isInnerType((IType) element);
+				try {
+					if (((IType) element).isImplicitlyDeclared()) {
+						return;
+					}
+				} catch (JavaModelException e) {
+					return;
+				}
 				break;
 			case IJavaElement.METHOD:
 			case IJavaElement.FIELD:


org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/folding/FoldingTest.java
```diff
@@ -826,5 +826,24 @@ public void run() {
...
+		String str = """
+				import module java.base;
+
+				int i;
+				int j;
+
+				void main() {
+					// main
+				}
+
```
fedejeanne: No need for this extra line.

org.eclipse.jdt.ui/ui/org/eclipse/jdt/ui/text/folding/DefaultJavaFoldingStructureProvider.java
```diff
@@ -1258,6 +1258,13 @@ protected void computeFoldingStructure(IJavaElement element, FoldingStructureCom
 			case IJavaElement.TYPE:
 				// only inner types may be collapsed
 				collapse= ctx.collapseInnerTypes() && isInnerType((IType) element);
+				try {
+					if (((IType) element).isImplicitlyDeclared()) {
+						return;
+					}
+				} catch (JavaModelException e) {
+					return;
+				}
```
fedejeanne, line 1273: I'd put it in a separate method and add some comment to improve readability:

### 
```suggestion
				if (isImplicitlyDeclared((IType) element)) {
					// Implicitly declared classes where introduced in Java 25
					// (see https://openjdk.org/jeps/512) and they don't need
					// to be folded.
					return;
				}
```

The extracted method would also log a `WARNING` in case something happens:

```java
private boolean isImplicitlyDeclared(IType element) {
		try {
			return element.isImplicitlyDeclared();
		} catch (JavaModelException e) {
			JavaPlugin.log(new Status(IStatus.WARNING, JavaPlugin.getPluginId(), IJavaStatusConstants.INTERNAL_ERROR, "An error occured while processing " + element.getFullyQualifiedName() + ".", e)); //$NON-NLS-1$ //$NON-NLS-2$
		}
		return false;
	}

```

fedejeanne: Building on this: since we now `return` before `computeProjectionRanges(...)` runs for the implicit `IType`, this doesn't just skip the "code" region — it skips *all* regions for that element, including any header/javadoc comment folding region that used to be anchored to it via `ctx.setFirstType(...)` (see `computeHeaderComment`, lines ~1397-1450). A compact source file with a header comment (license header, javadoc, etc.) before the code will silently lose that comment's folding region.

Test to reproduce (add to `FoldingTest.java`, e.g. right after the new `testSimpleMainDoesNotContainRegionSurroundingImplicitClass`):

```java
@Test
public void testHeaderCommentBeforeImplicitClassIsStillFolded() throws Exception {
	String str = """
			/**
			 * Header comment that should still be folded
			 */
			import module java.base;

			void main() {
				// main
			}

			""";
	List<FoldingTestUtils.ProjectionRegion> regions= FoldingTestUtils.getProjectionRangesOfPackage(packageFragment, str);
	FoldingTestUtils.assertContainsRegionUsingStartAndEndLine(regions, str, 0, 2); // header comment
	FoldingTestUtils.assertContainsRegionUsingStartAndEndLine(regions, str, 5, 7); // main
}
```

This fails on the current diff (the header-comment region is missing) and should pass once the fix only skips the "code" region for the implicit type instead of returning before any region computation happens.

Re: swallowing the exception — your extracted-method proposal is correct in spirit (logging instead of silently returning), just one addition: `element.getFullyQualifiedName()` can itself throw/be misleading for an implicit type (its "name" is normally empty/synthetic), so I'd use `element.getElementName()` or just log the exception's message/status without formatting a fully-qualified name into it, to avoid a confusing/empty log message for this specific case:

```java
private boolean isImplicitlyDeclared(IType element) {
	try {
		return element.isImplicitlyDeclared();
	} catch (JavaModelException e) {
		JavaPlugin.log(new Status(IStatus.WARNING, JavaPlugin.getPluginId(), IJavaStatusConstants.INTERNAL_ERROR, "An error occurred while processing " + element.getElementName() + ".", e)); //$NON-NLS-1$ //$NON-NLS-2$
	}
	return false;
}
```

Otherwise the extraction + logging approach looks good to me.

*(Created by copilot)*

fedejeanne: @danthe1st ignore my agent, it looks good:

image

```java
/**
 * Copyright <br/>
 * fede
 */

/**
 * Some other comment
 */
int i;
int j;

void main() {
	System.out.println();

}

void fede() {
	System.out.println();

}
```

fedejeanne: I would still separate your changes into a new method though (my 1st comment)

danthe1st: Fixed in e3949ea47554c6b6ac58eb546a53c3de2a463857.
The method suggested by you returns `false` in case of a `JavaModelException` (pretending to be a "real" type) and to be honest, I'd rather not show a folding region for error cases. Because of that, I inverted the method to check for implicitly declared types. An alternative name for that method would be `isVisibleInSourceCode` but that would not accurately represent that.

org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/folding/FoldingTest.java
danthe1st: Fixed in e3949ea47554c6b6ac58eb546a53c3de2a463857.

org.eclipse.jdt.ui/ui/org/eclipse/jdt/ui/text/folding/DefaultJavaFoldingStructureProvider.java
fedejeanne: Good point about not showing a folding region for error cases, I agree. This makes me now realize that the current code (`master`) actually doesn't really recover from a `JavaModelException` while processing an `IJavaElement` deep down in the calls to `computeFoldingStructure(...)` (in red), it does it in the caller: `processSourceReference(...)` (in green).

image

Which makes me wonder: should we even catch the `JavaModelException` in `protected void computeFoldingStructure(IJavaElement element, FoldingStructureComputationContext ctx)` in the first place or should be better `throw` it back at the caller just like `private void computeFoldingStructure(IJavaElement[] elements, FoldingStructureComputationContext ctx) throws JavaModelException` does?

I vote for throwing it. It would mean that a faulty `IJavaElement` would ruin the folding structure for lots of elements (probably even the whole file) but it would at least preserve the same error-handling logic we have today.

A side-note: you used `element.getElementName()` in the `catch` clause (probably taken from Copilot's comment), but notice that I used `element.getFullyQualifiedName()` in _my_ proposal. I did that because anonymous classes have an _empty_ (`""`) name but something like `MyClass$1` as a FQN.

My concrete proposal: 
- Don't catch/handle the `JavaModelException`, simply throw it
- Let's see how to better handle faulty elements in another PR.

WDYT?

fedejeanne: oh, and a last nit-pick: I prefer not to invert the meaning of methods in the code e.g. when you call the new method `isExplicitlyDeclared` and let it return `!type.isImplicitlyDeclared()`. For someone who is used to the API talking about _implicit_, switching to _explicit_ and adding a `!` is an unnecessary mental overload.

... but that's just my very own personal preference and I wouldn't oppose it if you do it differently :-)

danthe1st: > oh, and a last nit-pick: I prefer not to invert the meaning of methods in the code e.g. when you call the new method isExplicitlyDeclared and let it return !type.isImplicitlyDeclared(). For someone who is used to the API talking about implicit, switching to explicit and adding a ! is an unnecessary mental overload.

I generally agree. The only reason I did that (as alluded to in my previous comment) is the error case. I don't want to make the semantical statement of "the type is implicitly declared" if the Java element errors (even though I am not sure how that could happen).

I would like to do it in a non-inverted way but what should I name it? `isImplicitlyDeclaredOrErrorneous` `isNonActualTypeInSourceCode` (has the negation as well)?

danthe1st: Your proposal wouldn't work because `computeFoldingStructure(IJavaElement element, FoldingStructureComputationContext ctx)` is `protected` and API so we can't just add exceptions to it.

fedejeanne: Oh, good point: being an API method makes adding a checked exception break compatibility in the caller role:

- https://github.com/fedejeanne/eclipse.platform/blob/b6dcc49429153039e3ab8de695ead495d3e17638/docs/Evolving-Java-based-APIs.md?plain=1#L129

- https://github.com/fedejeanne/eclipse.platform/blob/b6dcc49429153039e3ab8de695ead495d3e17638/docs/Evolving-Java-based-APIs-2.md?plain=1#L119

In that case I suggest you go back to your previous implementation and handle the exception inside `protected void computeFoldingStructure(IJavaElement element, FoldingStructureComputationContext ctx)`:

```java
try {
	if (((IType) element).isImplicitlyDeclared()) {
		return;
	}
} catch (JavaModelException e) {
	JavaPlugin.log(new Status(IStatus.WARNING, JavaPlugin.getPluginId(), IJavaStatusConstants.INTERNAL_ERROR, "An error occurred while processing " + type.getFullyQualifiedName() + ".", e)); //$NON-NLS-1$ //$NON-NLS-2$
	return;
}
```

The reason is that returning `true`/`false` in the extracted method, regardless of how is it called (with/without negation in its name) feels odd. After all, when the method is faulty and an exception is thrown, it is neither implicitly nor explicitly declared: it's broken.

Notice that I logged the exception in the `catch` clause and that I used `type.getFullyQualifiedName()` because of what I mentioned [before](https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/3116#discussion_r3921786157):

> A side-note: you used `element.getElementName()` in the catch clause (probably taken from Copilot's comment), but notice that I used `element.getFullyQualifiedName()` in my proposal. I did that because anonymous classes have an empty ("") name but something like `MyClass$1` as a FQN.

danthe1st: Done/updated.

Note: For the `IStatus`, one could also use `JavaModelException#getStatus` but that would result in an `ERROR` status.

danthe1st:
@jjohnstn Would you (or someone else) like to review this? (the push I just made is a rebase)

jjohnstn:
> @jjohnstn Would you (or someone else) like to review this? (the push I just made is a rebase)

Sure.

jjohnstn (approved):
Looks good and works as advertised.

Issue #2416, opened 2025-08-05
Eclipse only offers to remove argument or create new record constructor on unknown argument to Java record constructor


steinarb:
Platform: 

- debian 12.11 "bookworm", amd64
- OpenJDK "21.0.4" 2024-07-16 LTS
- eclpse Java IDE Version: 2025-06 (4.36.0) Build id: 20250605-1316

When eclipse sees an unexpected argument to a Java record constructor. e.g.
```
    var image = new Image("Image");
```
then eclipse offers to ways to correct the issue:

1. Remove the argument (i.e. making it be `new Image()`)
2. Add an extra constructor to the record type with the argument

What I usually want when introducing an extra argument during test driven development, is to add the argument to the record type definition (or primary constructor or whatever it is called).

A fix option to change the record type property list should be added.

jjohnstn:
@steinarb One issue with adding it to the record declaration is that all existing calls that were valid will then be invalid.  Eclipse should never offer quick fixes that introduce errors so this would only be possible if there are no previous instances.

steinarb:
Also in my usage records are rarely changed after the initial creation so the only instances at this point in time are only in the unit test used to TDD the [record and its manually created builder](https://steinar.bang.priv.no/2024/05/10/build-java-records-with-builders/) (I create builders manually since [JEP 486 withers](https://steinar.bang.priv.no/2024/05/09/java-record-withers-what-and-when/) didn't get into Java 25), and later, the only place the record instance is created is by its builder.

But others may have different usages.

jjohnstn:
Hi @steinarb Looking at current refactorings, there is an introduce parameter object refactoring whereby one can add a new parameter to a method.  It starts by offering to create a new parameter class but you can override and specify existing classes such as String, Integer, etc.. and this changes the various calls to add a default constructor of the class specified.  I think a quick-fix could piggyback on that logic and make the refactoring call directly.

jjohnstn:
This fell through the cracks.  @inuyasha82 Want to tackle this?

inuyasha82:
Yeah, sure assign it to me.

inuyasha82:
@jjohnstn so when calling the refactoring,  it should refactor all the classes in the project (if the record is used anywhere else),  or only the  currenet class? 

I'm not sure if a quick fix usually change other files, or the scope should be limited to the current class only.

inuyasha82:
[...]

PR #3121, opened 2026-08-18
Refactor record signature

inuyasha82:
## What it does
Still in draft it implements changes for #2416 

## Author checklist


jjohnstn (changes requested):
Works well in testing.  One issue.  If I, for example, add a parameter that is List<String>, the code needs to add an import for List, otherwise, the changed code is marked with an error and all users as well since the record isn't valid any more.

inuyasha82:
> Works well in testing. One issue. If I, for example, add a parameter that is List, the code needs to add an import for List, otherwise, the changed code is marked with an error and all users as well since the record isn't valid any more.

Ok I think I've added this behavior, and di some refactoring inside the TypeContextChecker, to avoid too much code Duplication, I've removed also the isError function, that was implemented to return always true, and removed the if statement that was using it. 

Anyway the import is added only on the RecordDeclaration, not on the various instatiation, and this is the same behavior of Change Method Signature (i tried it and if i add for example a List, it updates correctly the method declaration, but where the method is used the import is not added, so this is still up to the user to add them. 

Since this is a behavior missing on both processors, i would suggest to implement it on a separate PR (if is possible) because this one is already huge. And for adding this logic an investigation is needed first.

jjohnstn:
Hi @inuyasha82 The import would only be necessary if the default value was a constructor (e.g. new ArrayList<>()).  It is very likely that a default value will usually be null for a class.

org.eclipse.jdt.ui/ui/org/eclipse/jdt/ui/actions/ModifyRecordParametersAction.java
```diff
@@ -0,0 +1,193 @@
+package org.eclipse.jdt.ui.actions;
```
jjohnstn, line 14: Missing  copyright.

org.eclipse.jdt.ui/ui refactoring/org/eclipse/jdt/internal/ui/refactoring/RefactoringMessages.java
```diff
@@ -88,6 +88,10 @@ public final class RefactoringMessages extends NLS {
 
 	public static String ChangeSignatureRefactoring_modify_Parameters;
 
+	public static String ChangeSignatureRefactoring_modify_Record_Parameters;
+
+	public static String ChangeSignatureRefactoring_records_modify_Parameters;
```
jjohnstn: Suspecting the 2nd message is unused.

org.eclipse.jdt.ui/plugin.properties
```diff
@@ -898,6 +900,9 @@ ActionDefinition.renameElement.description= Rename the selected element
 ActionDefinition.modifyMethodParameters.name= Change Method Signature
 ActionDefinition.modifyMethodParameters.description= Change method signature includes parameter names and parameter order
 
+ActionDefinition.modifyRecordParameters.name= Change Record Signature
+ActionDefinition.modifyRecrodParameters.description= Change record signature includes parameter names and parameter order
```
jjohnstn: Typo.  Also I don't see parameter reordering allowed.  Not sure why this wasn't added since the logic to do this was already in place for method signatures.

org.eclipse.jdt.ui/plugin.xml
```diff
@@ -4920,7 +4938,7 @@
       <key
             platform="carbon"
             sequence="COMMAND+ALT+C"
-            commandId="org.eclipse.jdt.ui.edit.text.java.modify.method.parameters"
+            commandId="org.eclipse.jdt.ui.edit.text.java.modify.record.parameters"
```
jjohnstn, line 4941: You can't just usurp the method parameters key sequence unless the new code handles both??

org.eclipse.jdt.ui/plugin.xml
```diff
@@ -4912,6 +4925,11 @@
             commandId="org.eclipse.jdt.ui.edit.text.java.modify.method.parameters"
             schemeId="org.eclipse.ui.defaultAcceleratorConfiguration"/>
             
+      <key
+            sequence="M2+M3+C"
```
jjohnstn: This key sequence seems to be already in use below.

org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/refactoring/TypeContextChecker.java
```diff
@@ -258,126 +424,42 @@ private int appendMethodDeclaration(StringBuilder cuString, String[] types, int
...
-			}
-			return result;
+		@Override
+		IType getDeclaringType() {
+			return fMethod.getDeclaringType();
 		}
 
 	}
 
 	private static class MethodTypesSyntaxChecker {
```
jjohnstn: Maybe rename this ParameterTypesSyntaxChecker

org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/refactoring/structure/ChangeRecordSignatureProcessor.java
```diff
@@ -0,0 +1,728 @@
+package org.eclipse.jdt.internal.corext.refactoring.structure;
```
jjohnstn, line 14: Missing copyright.

org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/refactoring/structure/ChangeRecordSignatureProcessor.java
```diff
@@ -0,0 +1,728 @@
...
+		private ASTNode fNullNode;
+		protected NullOccurrenceUpdate(ASTNode node, CompilationUnitRewrite cuRewrite, RefactoringStatus result) {
+			super(cuRewrite, null, result);
+			fNullNode= node;
+		}
+		@Override
+		public void updateNode() throws JavaModelException {
+			int start= fNullNode.getStartPosition();
+			int length= fNullNode.getLength();
+			String msg= "Cannot update found node: nodeType=" + fNullNode.getNodeType() + "; "  //$NON-NLS-1$//$NON-NLS-2$
```
jjohnstn, line 130: Shouldn't this be a convertible message?

jjohnstn (changes requested):
Works well.  Not sure why you didn't include moving of parameters which is done in method parameters wizard which could be copied/reused.  The import is now done so that is good.

inuyasha82:
> Works well. Not sure why you didn't include moving of parameters which is done in method parameters wizard which could be copied/reused. The import is now done so that is good.

Didn't add the fields ordering, because given the size of the PR, was not sure if it was adding even more lines of code to it, and also tried to keep the minimalisti approach to add only the requested feature, to keep focused on it. Parameter re-ordering can be added later.

org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/refactoring/structure/ChangeRecordSignatureProcessor.java
inuyasha82: Maybe, i just copied how  the changeSignatureProcessor does, and it is using the Non convertible strings too.

org.eclipse.jdt.ui/plugin.xml
inuyasha82: Selected M2+M3+E that apparently is  available.

org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/refactoring/structure/AbstractSignatureProcessor.java
```diff
@@ -0,0 +1,148 @@
+/*******************************************************************************
+ * Copyright (c) 2000, 2024 IBM Corporation and others.
```
jjohnstn: Copyright date should be 2026 for a new file.  You can note that this was copied from xxxxx in the Contributors list.

org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/refactoring/structure/ChangeRecordSignatureProcessor.java
```diff
@@ -0,0 +1,607 @@
...
```
jjohnstn: Msgs should be nls unless they are just log messages.  I note that you add an error below which means the user sees this message in the refactoring dialog so it needs to be translatable.

org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/refactoring/structure/ChangeRecordSignatureProcessor.java
```diff
@@ -0,0 +1,607 @@
...
+		@Override
+		protected void changeParamgumentType(ParameterInfo info) {
+			SingleVariableDeclaration oldParam= getParameter(info.getOldIndex());
+			SingleVariableDeclaration oldSVDParam= oldParam;
+			replaceTypeNode(oldSVDParam.getType(), ParameterInfo.stripEllipsis(info.getNewTypeName()), info.getNewTypeBinding());
+		}
+
+	}
+
+	abstract class OccurrenceUpdate <N extends ASTNode> extends AbstractOccurrenceUpdate<N>{
```
jjohnstn: no space between class and type arguments

Any reason for this abstract class extending the other abstract class?  If the other class is never used directly except here, you might as well create one abstract class.  If it is used with other type arguments, then ignore this comment.

org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/refactoring/structure/ChangeRecordSignatureProcessor.java
```diff
@@ -0,0 +1,607 @@
...
+					if (!ASTNodes.isExistingNode(node)) //XXX:should better be addressed in ListRewriteEvent.replaceEntry(ASTNode, ASTNode)
+						listRewrite.replace(newOldMap.get(node), newNode, fDescription);
+					else
+						listRewrite.replace(node, newNode, fDescription);
+				}
+				//remove remaining existing nodes:
+				while (nodesIter.hasNext()) {
+					ASTNode node= nodesIter.next();
+					if (!ASTNodes.isExistingNode(node))
+						listRewrite.remove(newOldMap.get(node), fDescription);
```
jjohnstn, line 272: For removed nodes, you need to use an ImportRemover.  I tried an example, where I remove a List<> parameter and it failed to remove the import of java.util.List.

org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/refactoring/structure/ChangeRecordSignatureProcessor.java
```diff
@@ -0,0 +1,607 @@
...
+		} finally {
+			pm.done();
+		}
+	}
+
+	@Override
+	public RefactoringStatus checkFinalConditions(IProgressMonitor pm, CheckConditionsContext context) throws CoreException, OperationCanceledException {
+		pm.beginTask(RefactoringCoreMessages.ChangeSignatureRefactoring_checking_preconditions, 8);
+		RefactoringStatus result= new RefactoringStatus();
+		fBaseCuRewrite.clearASTAndImportRewrites();
```
jjohnstn, line 387: Is this needed?  Will there be any changes or import rewrites at this point??

org.eclipse.jdt.ui.tests.refactoring/test cases/org/eclipse/jdt/ui/tests/refactoring/ChangeRecordSignatureTests.java
```diff
@@ -0,0 +1,223 @@
...
+	}
+
+	@Test
+	public void testReorderParameters() throws Exception {
+		// reorder A(int a, String b) -> A(String b, int a), instantiation args should also be reordered
+		helperReorder(new int[]{1, 0});
+	}
+
+	@Test
+	public void testAddParamWithImport() throws Exception {
```
jjohnstn, line 247: Add another test for List parameter being removed and verify that import of List is also removed.

org.eclipse.jdt.ui/ui refactoring/org/eclipse/jdt/internal/ui/refactoring/ChangeParametersControl.java
```diff
@@ -80,18 +80,22 @@ private Mode(String name) {
...
 		public String toString() {
 			return fName;
 		}
 		public boolean canChangeTypes() {
-			return this == CHANGE_METHOD_SIGNATURE;
+			return this == CHANGE_METHOD_SIGNATURE || this == CHANGE_RECORD_SIGNATURE;
 		}
 		public boolean canAddParameters() {
-			return this == Mode.CHANGE_METHOD_SIGNATURE;
+			return this == Mode.CHANGE_METHOD_SIGNATURE || this == CHANGE_RECORD_SIGNATURE;
```
jjohnstn, line 92: Why are some references prefixed with Mode. and others not?

org.eclipse.jdt.ui/ui refactoring/org/eclipse/jdt/internal/ui/refactoring/ChangeRecordSignatureWizard.java
```diff
@@ -0,0 +1,123 @@
+package org.eclipse.jdt.internal.ui.refactoring;
```
jjohnstn, line 14: Missing copyright header.

org.eclipse.jdt.ui/plugin.xml
```diff
@@ -4912,15 +4925,20 @@
             commandId="org.eclipse.jdt.ui.edit.text.java.modify.method.parameters"
             schemeId="org.eclipse.ui.defaultAcceleratorConfiguration"/>
             
+      <key
+            sequence="M2+M3+E"
```
jjohnstn: Not all commands need to have a key shortcut.  I'm not sure anyone will find E intuitive for changing a record signature.  I think if you can do without it, it would be better to leave the key combo for something else to use.

jjohnstn (changes requested):
Works well.  Just one scenario where deleting a parameter forgets to remove the import.

org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/refactoring/structure/ChangeRecordSignatureProcessor.java
inuyasha82: Ok, Anyway I just copied it from the ChangeSignatureProcessor, do you want me to change it there as well?

org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/refactoring/structure/ChangeRecordSignatureProcessor.java
inuyasha82: The Reason is mainly that OccurrenceUpdate here and in ChangeSignatureProcessor are different, and the AbstractOccurrenceUpdate is about the common part of the two occurrence updates. Then OccurrenceUpdate is used  as superclass by the other *Update classes in ChangeSignatureProcessor and ChangeRecordSignatureProcessor.

org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/refactoring/structure/ChangeRecordSignatureProcessor.java
inuyasha82: yeah, import remover was the missing part, next commit will add it (although this code here shouldn 'need to be changed.

org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/refactoring/structure/AbstractSignatureProcessor.java
inuyasha82: done

org.eclipse.jdt.ui/plugin.xml
inuyasha82: Ok so for this one do I just need to remove the <key sequence...> tags?

org.eclipse.jdt.ui/ui refactoring/org/eclipse/jdt/internal/ui/refactoring/ChangeParametersControl.java
inuyasha82: About the why, not sure! The original code was mixing them. I think there is no exact reason, since looks like the checks they are doing use the same logic, I just added my || part to what was already there. Let me know if you want me just to make my changes coherent with the already existing code, or simply make everything equal by adding or removing Mode.

org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/refactoring/structure/ChangeRecordSignatureProcessor.java
inuyasha82: Again not sure, i created checkFinalCondition following more or less the pattern of the one in ChangeSignatureProcessor. From what I understood from the code so far the change is not made yet, it will be done later in this function, but not yet done. So I think it could be needed, but again, I used ChangeSignatureProcessor as Base, and this was done the exact same way there, so i kept it.

org.eclipse.jdt.ui.tests.refactoring/test cases/org/eclipse/jdt/ui/tests/refactoring/ChangeRecordSignatureTests.java
inuyasha82: done

org.eclipse.jdt.ui/plugin.xml
jjohnstn: Just remove the key tags.  I don't think this refactoring will be common enough to justify a key short-cut.

org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/refactoring/structure/ChangeRecordSignatureProcessor.java
jjohnstn: You can do it another time.

org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/refactoring/structure/ChangeRecordSignatureProcessor.java
jjohnstn: Ok

org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/refactoring/structure/ChangeRecordSignatureProcessor.java
jjohnstn: Actually, if it is the same message and you are changing ChangeSignatureProcessor, you might as well do it there as well.

org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/refactoring/structure/ChangeRecordSignatureProcessor.java
jjohnstn: Ok

org.eclipse.jdt.ui/ui refactoring/org/eclipse/jdt/internal/ui/refactoring/ChangeParametersControl.java
jjohnstn: They are the same so no issue.

org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/refactoring/structure/ChangeRecordSignatureProcessor.java
jjohnstn: This is a duplicate of another comment below.

jjohnstn:
Import remove works.  Just fix the msg that is used for a error status and remove the keys settings and this is ready to merge.

org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/refactoring/structure/ChangeRecordSignatureProcessor.java
inuyasha82: Done. Yeah it was the same message.

jjohnstn (approved):
Looks good.  Thanks.

jjohnstn:
@inuyasha82 Please add a N&N entry.

inuyasha82:
Thanks, sure will do @jjohnstn

PR #3126, opened 2026-08-23
Skip top level types in "Collapse All" editor actions

danthe1st:
## What it does

Currently, the `Collapse All` editor operation results in all elements including top level types being collapsed. So for most classes, the `Collapse All` operation results in a mostly empty class.

In some older Eclipse IDE releases (before extended folding was introduced), top level types were excluded from folding so this editor action only affected nested elements (and things like imports).

This PR disables the `Collapse All` operation on top level types.

## How to test

Create a class like this:
```java
class MyClass {
    void someMethod() {
        // does some work
    }
}
```

The "Collapse All" action should collapse `someMethod` but not `MyClass`.

[Screencast_20260823_115609.webm](https://github.com/user-attachments/assets/0d91bf4b-457a-4319-a4fa-fb769e4c2b5a)

~~This PR depends on https://github.com/eclipse-platform/eclipse.platform.ui/pull/4272.~~ This has been merged.

~~I know that we currently have a code freeze and I'm submitting it with the intention of ideally getting this into the release afterwards (4.42).~~

CC @HannesWell since you suggested this and @fedejeanne since you were involved in the folding changes.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/folding/FoldingTest.java
+++ b/org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/folding/FoldingTest.java
@@ -28,10 +28,13 @@ import org.junit.runners.Parameterized.Parameter;
 import org.junit.runners.Parameterized.Parameters;
 
 import org.eclipse.jdt.testplugin.JavaProjectHelper;
+import org.eclipse.jdt.text.tests.folding.FoldingTestUtils.ProjectionRegion;
 
 import org.eclipse.jface.preference.IPreferenceStore;
 
+import org.eclipse.jface.text.source.projection.ProjectionAnnotationModel;
 
+import org.eclipse.jdt.core.ICompilationUnit;
 import org.eclipse.jdt.core.IJavaProject;
 import org.eclipse.jdt.core.IPackageFragment;
 import org.eclipse.jdt.core.IPackageFragmentRoot;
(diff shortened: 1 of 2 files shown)


fedejeanne:
While testing this (specifically, [this comment](https://github.com/eclipse-platform/eclipse.platform.ui/pull/4272#discussion_r3894874910) from Copilot) I noticed that there is some residual line in the line below the opening bracker (`{`) when I manually fold the class and place the cursor at the end of the line:

image

**To reproduce**
- Checkout this PR and also https://github.com/eclipse-platform/eclipse.platform.ui/pull/4272
- Copy the class in the description (add an empty line at the end, without it I see no `-`-symbol that lets me fold the class)
- Run the command `Collapse All`
- Press the `-`-symbol to fold `MyClass`
- Put the text cursor at the end of the line `class MyClass {`

danthe1st:
@fedejeanne 

> While testing this (specifically, https://github.com/eclipse-platform/eclipse.platform.ui/pull/4272#discussion_r3894874910 from Copilot) I noticed that there is some residual line in the line below the opening bracker ({) when I manually fold the class and place the cursor at the end of the line:

It seems like this generally happens when first collapsing the inner and then the outer region in that class (and it also seems to happen only if there is no class/other code below. Given that I was able to reproduce it even in 2026-03, I think it is better to address this outside of this PR.

danthe1st:
~~I am marking this as a draft due to likely being obsoleted by #3139.~~

danthe1st:
With #3139 being closed due to the performance issue, I am marking this as ready for review again.
When https://github.com/eclipse-platform/eclipse.platform.ui/pull/4272 is merged, I will rebase this PR after an I-build.

PR #3135, opened 2026-09-01
Add highlight feature for resources in package explorer

elsazac:
This PR adds a new highlight feature to the Eclipse Package Explorer. It allows users to mark individual files with bold text with COLOR_LIST_FOREGROUND and an optional foreground colour, making important files easier to identify quickly.

In large Eclipse workspaces with many projects and files, it can be difficult to find the files you need. For example:

- Files with similar names — for eg, files such as IBreakpointListener.java, IBreakpointManager.java, IBreakpointManagerListener.java, and IBreakpointsListener.java can look very similar and be difficult to distinguish. Highlighting makes the file you need stand out.
- Files currently being worked on — Highlighting helps users keep track of files that are important for the current task.
- Important configuration files — Build scripts, deployment files, properties files, and other important configuration files can be highlighted so they are easier to notice.

To remove the highlight, right-click the highlighted resource and select `Remove Highlight`. The highlight is then removed. The highlight color can optionally be customized through the Colors and Fonts preference page. 

## What it does

To highlight a resource:

- Right-click a file in the Package Explorer.
- Select `Highlight Resource` from the context menu.
- The resource is displayed using bold text and the default foreground colour COLOR_LIST_FOREGROUND.

https://github.com/user-attachments/assets/11da691e-b9f5-41db-ad84-9cf6f2adb130

## Author checklist

### DIFF
--- a/org.eclipse.jdt.ui/ui/org/eclipse/jdt/ui/PreferenceConstants.java
+++ b/org.eclipse.jdt.ui/ui/org/eclipse/jdt/ui/PreferenceConstants.java
@@ -4056,6 +4056,16 @@ public class PreferenceConstants {
 	public static final String EDITOR_JAVA_CODEMINING_FILTER_IMPLIED_PARAMETER_NAMES = "java.codemining.filter.implied.parameterNames"; //$NON-NLS-1$
 
 	/**
+	 * A named preference that holds the color used to highlight resources
+	 * in the Package Explorer.
+	 * <p>
+	 * Value is of type <code>String</code>.When this preference has a non-default value, the specified foreground color is applied in addition to the bold styling.
+	 * </p>
+	 * @since 3.39
+	 */
+	public static final String PACKAGE_EXPLORER_HIGHLIGHT_COLOR= "packageExplorerHighlightColor"; //$NON-NLS-1$
+
+	/**
 	 * A named preference that stores the maximum number of chain completions
 	 * to be proposed at one time.
 	 * <p>
--- a/org.eclipse.jdt.ui/plugin.properties
+++ b/org.eclipse.jdt.ui/plugin.properties
@@ -174,6 +174,8 @@ coloredLabels.writeaccess_highlight.description=The background color used to hig
 
 coloredLabels.inherited.label=Inherited members
 coloredLabels.inherited.description=The color used to render to inherited members. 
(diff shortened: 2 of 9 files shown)


elsazac:
cc/ @SougandhS

iloveeclipse:
My concern here is that the data for the marked files is saved independently from the resource tree. If the tree changes (file renamed / moved / deleted), the preference data will stay forever, growing over time. 
Wouldn't it be enough to save the data into the field or session property for the current session only?

elsazac:
> Wouldn't it be enough to save the data into the field or session property for the current session only?

Yes, thanks for the review. As you suggested, I have removed the part where we store this in the preference store, so that it won’t persist across sessions.

org.eclipse.jdt.ui/ui/org/eclipse/jdt/ui/PreferenceConstants.java
```diff
@@ -4055,6 +4055,16 @@ private PreferenceConstants() {
 	 */
 	public static final String EDITOR_JAVA_CODEMINING_FILTER_IMPLIED_PARAMETER_NAMES = "java.codemining.filter.implied.parameterNames"; //$NON-NLS-1$
 
+	/**
+	 * A named preference that holds the color used to highlight resources
+	 * in the Package Explorer.
+	 * <p>
+	 * Value is of type <code>String</code>.When this preference has a non-default value, the specified foreground color is applied in addition to the bold styling.
+	 * </p>
+	 * @since 3.39
```
iloveeclipse, line 4064: Since is wrong, and API check fails because of missing increment

iloveeclipse:
I wonder if this concept can be generalized to the platform. 

So **any** explorer (or from editor) can mark some resource as "important" and one could see "important" resources in some view. If so, it would make sense to create a marker (or bookmark) on the "important" resource and have generic label decoration for "important" resources, with a dedicated marker group, so it could be found in "Markers" (or "Bookmarks") view if needed?

Can you please start discussion in platform UI about this PR and whether possible generic solution could be considered as useful for the platform in general?

elsazac:
> Can you please start discussion in platform UI about this PR and whether possible generic solution could be considered as useful for the platform in general?

Sure will start a discussion in platform.

elsazac:
I have started one and attaching the discussion here:
https://github.com/eclipse-platform/eclipse.platform.ui/discussions/4336

Issue #2040, opened 2025-02-25
Remember JUnit results upon restart


mickaelistria:
I often need to restart my IDE because of an update or a quirk. When I do so, I loose my JUnit history, while I'm actually spending most of my time comparing those reports. As a result, whenever I restart, I need to remember which particular test I was running and loose ability to compare with previous runs.
The JUnit reports would IMO be worth being persisted across restarts.

carstenartur:
Hi @mickaelistria,

you originally reported #2394, and I have now implemented it in #3137. Would
you be interested in testing the PR in a real workspace?

The implementation restores completed JUnit runs after a clean Eclipse restart.
Only the small history metadata is read at startup; the complete test tree is
loaded when a restored entry is selected.

A particularly useful test would be to run a few successful, failing, and
stopped test runs, restart Eclipse with the same workspace, and check the
history ordering, icons, counters, stopped state, test trees, and failure
traces.

The automated persistence tests and all current CI checks pass. Feedback on the
behavior or any missing edge cases would be greatly appreciated. Thank you!

jjohnstn:
Closing as implemented by: https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/3137

PR #3137, opened 2026-09-02
Persist JUnit history across restart

### LINKED ISSUE #2394: Persist/Restore JUnit history across restart
Currently, history is lost on restarting, which can require some time for a rebuild/re-configure/re-run when someone has to restart but is still interested in reading test results.
The history should be persisted and restored.

carstenartur:
Persist and restore the bounded JUnit test-run history for a workspace across
clean Eclipse restarts.

Fixes #2394  
Related to #2040

## Problem

The JUnit Results view already keeps multiple test-run sessions during an
Eclipse session. However, this history is discarded when Eclipse shuts down.

This can be inconvenient when Eclipse must be restarted while a developer still
needs to inspect a failure trace, compare recent runs, or rerun tests using the
same launch configuration.

## User-visible behavior

This change persists the existing bounded JUnit test-run history in the
workspace state area and restores it when the same workspace is opened again.

After a restart:

- the retained test runs are available through the existing JUnit history
  dropdown and page switcher;
- the most recent retained run is displayed initially;
- users can switch between all restored runs up to the configured history
  limit;
- run name, project, original start time, stopped/completed state, aggregate
  result, counters, and include/exclude tags are preserved;
- the complete test tree and failure details are loaded only when a restored run
  is accessed;
- “Rerun Tests” and “Run Errors First” can use the original saved launch
  configuration when it is still available.

The original launch process is not reconstructed. For rerunning, the persisted
history stores only:

- the memento of the persistent source launch configuration; and
- the original launch mode.

Temporary launch configurations created by JUnit rerun actions, including
“Run Errors First”, retain a reference to their persistent source
configuration. This allows subsequent reruns to work during the current Eclipse
session and after a restart.

If the original saved launch configuration has been deleted, the historical
test result remains available, but actions requiring that configuration are
disabled.

## Configuration

The existing “Maximum count of remembered test runs” setting continues to
define the maximum number of sessions kept by the JUnit history. The same limit
is applied to the persisted history.

No additional preference is introduced by this change.

## Implementation

The persisted history consists of:

1. a small, versioned metadata index used during Eclipse startup; and
2. one XML file per retained test run, using the existing JUnit test-run
   serialization format.

Only the metadata index is read during startup. Complete test trees, stack
traces, and failure details remain on disk until the corresponding history entry
is accessed.

The implementation:

- stores history only in the JUnit core state area of the workspace and does not
  modify project files;
- uses UUID-based file identities that do not depend on locale, time zone, or
  timestamp uniqueness;
- writes test-run files and the history index through temporary files and
  atomically replaces their published versions where supported;
- keeps the previously published history generation intact until a replacement
  index has been written successfully;
- does not overwrite an unsupported, unreadable, or only partially understood
  history generation;
- excludes restored sessions that became invalid after a failed lazy load,
  without preventing other valid sessions from being persisted;
- removes obsolete, unpublished, and abandoned temporary files only when it is
  safe to do so;
- preserves files referenced by the currently published index until a
  replacement index has been published.

## Scope and limitations

The feature applies to clean Eclipse shutdowns and restarts.

It does not attempt to recover a test run that existed only in memory when the
Eclipse process crashed or was terminated forcibly.

The history is local to the workspace. It is not synchronized between
workspaces or machines.

Rerunning a restored session requires that its persistent source launch
configuration can still be resolved. Imported test reports or restored entries
without a resolvable launch configuration remain available for inspection but
cannot be rerun.

## Automated tests

Automated coverage is provided by `TestRunSessionHistoryTests`.

The tests cover:

- metadata-only restoration and lazy test-tree loading;
- releasing and repeatedly loading restored test trees;
- preservation of imported start times across locale and time-zone changes;
- stopped runs, including runs containing failures and runs where all tests had
  already started;
- completed empty runs and their undefined result;
- equal timestamps without file collisions;
- configured history limits;
- restoration of the launch configuration and launch mode required for reruns;
- preservation of the persistent source configuration for temporary rerun
  working copies created by “Run Errors First”;
- disabling rerun without losing the historical result when the original saved
  launch configuration no longer exists;
- invalid restored sessions without blocking persistence of other runs;
- stale, missing, truncated, and unpublished test-run files;
- unsupported, non-regular, and partially invalid history indexes;
- preservation of the last published history generation when a replacement
  cannot safely be written;
- cleanup of abandoned temporary files.

Validation results:

- `TestRunSessionHistoryTests`: 23 tests, 0 failures, 0 errors, 0 skipped;
- complete Maven/Tycho reactor validation: successful;
- pull-request checks: successful;
- CodeQL: successful;
- service-version, merge-commit, freeze-period, and Eclipse ECA checks:
  successful.

## Suggested manual verification

1. Run several saved JUnit launch configurations, including successful and
   failing tests.
2. Optionally stop one test run before it completes.
3. Switch between the runs using the JUnit history dropdown.
4. Close Eclipse cleanly and reopen the same workspace.
5. Verify that the retained runs reappear with the expected names, ordering,
   result icons, counters, and stopped state.
6. Select restored runs and verify that their complete test trees and failure
   traces are loaded.
7. Switch away from a restored run and select it again.
8. Run a configuration containing failures and select “Run Errors First”.
9. Verify that “Rerun Tests” works for the resulting run.
10. Restart Eclipse and verify that “Rerun Tests” resolves and launches the
    original saved configuration.
11. Delete the original saved launch configuration and verify that the
    historical result remains available while rerun actions are disabled.
12. Remove individual history entries and clear the history.
13. Change the maximum number of remembered test runs and verify that the
    current and restored history remain bounded.

## Author checklist

- [x] The change follows the coding conventions
- [x] I have signed the Eclipse Contributor Agreement (ECA)

### DIFF
--- a/org.eclipse.jdt.junit.core/META-INF/MANIFEST.MF
+++ b/org.eclipse.jdt.junit.core/META-INF/MANIFEST.MF
@@ -3,7 +3,7 @@ Automatic-Module-Name: org.eclipse.jdt.junit.core
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.jdt.junit.core;singleton:=true
-Bundle-Version: 3.15.0.qualifier
+Bundle-Version: 3.15.100.qualifier
 Bundle-Activator: org.eclipse.jdt.internal.junit.JUnitCorePlugin
 Bundle-ActivationPolicy: lazy
 Bundle-Vendor: %providerName
--- a/org.eclipse.jdt.junit.core/src/org/eclipse/jdt/internal/junit/launcher/JUnitLaunchConfigurationConstants.java
+++ b/org.eclipse.jdt.junit.core/src/org/eclipse/jdt/internal/junit/launcher/JUnitLaunchConfigurationConstants.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2018 IBM Corporation and others.
+ * Copyright (c) 2000, 2026 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -19,6 +19,7 @@ import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.resources.ResourcesPlugin;
 
 import org.eclipse.debug.core.ILaunchConfiguration;
+import org.eclipse.debug.core.ILaunchConfigurationWorkingCopy;
 
 import org.eclipse.jdt.core.IJavaProject;
 import org.eclipse.jdt.core.JavaCore;
@@ -70,6 +71,8 @@ public class JUnitLaunchConfigurationConstants {
 
 	public static final String ATTR_FAILURES_NAMES= JUnitCorePlugin.PLUGIN_ID+".FAILURENAMES"; //$NON-NLS-1$
 
+	public static final String ATTR_RERUN_LAUNCH_CONFIG_MEMENTO= JUnitCorePlugin.PLUGIN_ID + ".RERUN_LAUNCH_CONFIG_MEMENTO"; //$NON-NLS-1$
+
 	public static final String ATTR_TEST_RUNNER_KIND= JUnitCorePlugin.PLUGIN_ID+".TEST_KIND"; //$NON-NLS-1$
 
 	public static final String ATTR_TEST_HAS_INCLUDE_TAGS= JUnitCorePlugin.PLUGIN_ID + ".HAS_INCLUDE_TAGS"; //$NON-NLS-1$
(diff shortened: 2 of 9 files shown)


jjohnstn (commented):
Playing around with it, it restores the dialog, showing any exceptions for tests upon clicking on them.  I do note that the buttons to rerun the tests or to rerun error tests are disabled in the dialog upon restart.  Right clicking on any test allows rerun or redebug, but it would be nice to have those two buttons for the dialog re-enabled as well.

carstenartur:
> Playing around with it, it restores the dialog, showing any exceptions for tests upon clicking on them.  I do note that the buttons to rerun the tests or to rerun error tests are disabled in the dialog upon restart.  Right clicking on any test allows rerun or redebug, but it would be nice to have those two buttons for the dialog re-enabled as well.

Thank you for testing this and for pointing out the missing rerun support.

I have now persisted the launch configuration memento and the original launch
mode with completed test runs. After a restart, restored sessions resolve the
existing launch configuration, so the “Rerun Test” and “Rerun Failed Tests
First” toolbar actions are enabled again. The per-test Run and Debug actions now
use the same restored configuration.

The previous launch process itself is not reconstructed. If the referenced
launch configuration has since been deleted, the test results remain available,
but the rerun actions are disabled.

I also added regression tests for restoring the launch configuration and mode,
and for the case where the configuration has been deleted. The focused history
suite now passes with 22 tests, and the complete Maven/Tycho reactor build is
successful.

jjohnstn:
One additional observation with the latest changes which fixes the average situation.  If you do a run that has failures, then hit the Run errors first button, the rerun occurs and the order of the tests change as expected.  However, the Run errors first button is then disabled while the Rerun tests button is still enabled.  Even though the Rerun tests button is enabled, it doesn't work.  If you restart, then both buttons end up disabled on restart.

carstenartur:
> One additional observation with the latest changes which fixes the average situation.  If you do a run that has failures, then hit the Run errors first button, the rerun occurs and the order of the tests change as expected.  However, the Run errors first button is then disabled while the Rerun tests button is still enabled.  Even though the Rerun tests button is enabled, it doesn't work.  If you restart, then both buttons end up disabled on restart.

Thank you — this exposed a separate case involving the temporary launch
configuration created by “Run errors first”.

`ILaunchConfiguration.copy(String)` creates a detached working copy. Such a
configuration is valid for launching during the current Eclipse session, but
`exists()` is false and it has no durable original-configuration reference.
Consequently, “Rerun Tests” could remain visually enabled while the actual
rerun returned without launching anything. The temporary configuration also
had no persistent memento that could be restored after a restart.

The update now:

- treats launch working copies as usable for rerunning during the current
  session;
- records the memento of the persistent source configuration in temporary
  JUnit rerun configurations;
- propagates that memento through subsequent rerun configurations;
- persists the source-configuration memento for restored test-run history; and
- refreshes the “Rerun Tests” action when a session starts, ends, or stops.

A regression test now covers persistence and restoration from an actual
temporary working copy. The history suite contains 23 tests.

When the errors-first run no longer contains failures, “Run Errors First”
remains correctly disabled. “Rerun Tests” is now backed by the preserved
original configuration both before and after restarting Eclipse.

jjohnstn:
Thanks @carstenartur Works great.  I'll do a review of the code now.

jjohnstn (approved):
Approved.

jjohnstn:
Thanks @carstenartur Would you be willing to write up a N&N entry?

mickaelistria:
I'm not in need of this feature any more at this moment so I couldn't test, but thanks for the patch anyway!

PR #3138, opened 2026-09-02
Move SortMembersOperation to core.manipulation

MeherSru:
This moves SortMembersOperation from org.eclipse.jdt.ui into org.eclipse.jdt.core.manipulation, so headless clients like the Eclipse JDT Language Server can actually use it. Right now it's stuck in the UI plug-in, so anything that isn't a full Eclipse IDE can't reach it.

The only thing blocking the move was one line depending on a UI-only class (JavaPlugin.getDefault().getMemberOrderPreferenceCache()). I swapped it for the equivalent that already exists in core.manipulation (JavaManipulationPlugin.getDefault().getMembersOrderPreferenceCacheCommon()), so the file no longer needs anything from org.eclipse.jdt.ui.

Related to redhat-developer/vscode-java#3990, which asks for alphabetical member sorting in jdtls,  jdtls will pick this up in a follow-up change.

Manually tested in a runtime Eclipse instance: created a class with out-of-order fields and methods, ran Source > Sort Members with "Sort all members" selected, and confirmed everything sorted alphabetically as expected.

### DIFF
--- a/org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/codemanipulation/SortMembersOperation.java
+++ b/org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/codemanipulation/SortMembersOperation.java
@@ -13,10 +13,9 @@
  *******************************************************************************/
 package org.eclipse.jdt.internal.corext.codemanipulation;
 
+import java.text.Collator;
 import java.util.Comparator;
 import java.util.List;
-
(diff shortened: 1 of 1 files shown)


jjohnstn (approved):
Approved.

iloveeclipse:
No manifest changes for 4.42?

jjohnstn:
> No manifest changes for 4.42?

I had already rebased some patches that didn't make it into 4.41 so they invoked bumps.

PR #3139, opened 2026-09-02
Skip top level types in collapse all editor actions

fedejeanne:
## What it does
This is a JDT-only alternative to eclipse-jdt/eclipse.jdt.ui#3126 (which requires a new API in platform).

## How to test
Same as with #3126 , use this class and `Ctrl + 3 > Collapse All`

```java
class MyClass {
    void someMethod() {
        // does some work
    }
}
// notice the extra line here. This is necessary, otherwise MyClass will never have a region in the first place. See https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/3117
```

`someMethod()` should be folded but `MyClass` shouldn't

## Author checklist

### DIFF
--- a/org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/folding/FoldingTest.java
+++ b/org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/folding/FoldingTest.java
@@ -31,7 +31,6 @@ import org.eclipse.jdt.testplugin.JavaProjectHelper;
 
 import org.eclipse.jface.preference.IPreferenceStore;
 
-
 import org.eclipse.jdt.core.IJavaProject;
 import org.eclipse.jdt.core.IPackageFragment;
 import org.eclipse.jdt.core.IPackageFragmentRoot;
@@ -824,7 +823,26 @@ public class FoldingTest {
 		FoldingTestUtils.assertContainsCollapsedRegionUsingStartAndEndLine(regions, str, 10, 20); // class A2
 		FoldingTestUtils.assertContainsCollapsedRegionUsingStartAndEndLine(regions, str, 12, 17); // Runnable
 		JavaPlugin.getDefault().getPreferenceStore().setToDefault(PreferenceConstants.EDITOR_FOLDING_INNERTYPES);
+	}
 
+	/**
+	 * "Collapse All" should not collapse top-level types, so that classes are not left completely
+	 * empty. See <a href="https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/3126">GitHub Issue
+	 * #3126</a>
+	 */
+	@Test
+	public void testCollapseAllSkipsTopLevelType() throws Exception {
+		String str= """
+				class MyClass {
+					void someMethod() {
+						// does some work
+					}
+				}
+				""";
+		FoldingTestUtils.collapseAll(packageFragment, str);
+		List<FoldingTestUtils.ProjectionRegion> regions= FoldingTestUtils.getProjectionRangesOfPackage(packageFragment, str);
+		FoldingTestUtils.assertContainsExpandedRegionUsingStartAndEndLine(regions, str, 0, 4); // class MyClass
+		FoldingTestUtils.assertContainsCollapsedRegionUsingStartAndEndLine(regions, str, 1, 3); // someMethod
 	}
 }
 
--- a/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/javaeditor/JavaSourceViewer.java
+++ b/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/javaeditor/JavaSourceViewer.java
@@ -16,6 +16,7 @@ package org.eclipse.jdt.internal.ui.javaeditor;
 import java.text.Bidi;
 import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.Map;
 
 import org.eclipse.swt.SWT;
@@ -45,9 +46,12 @@ import org.eclipse.jface.text.formatter.FormattingContextProperties;
 import org.eclipse.jface.text.formatter.IFormattingContext;
 import org.eclipse.jface.text.information.IInformationPresenter;
 import org.eclipse.jface.text.reconciler.IReconciler;
+import org.eclipse.jface.text.source.Annotation;
 import org.eclipse.jface.text.source.IOverviewRuler;
 import org.eclipse.jface.text.source.IVerticalRuler;
 import org.eclipse.jface.text.source.SourceViewerConfiguration;
+import org.eclipse.jface.text.source.projection.ProjectionAnnotation;
+import org.eclipse.jface.text.source.projection.ProjectionAnnotationModel;
 import org.eclipse.jface.text.source.projection.ProjectionViewer;
 
 import org.eclipse.ui.texteditor.AbstractDecoratedTextEditorPreferenceConstants;
@@ -60,6 +64,7 @@ import org.eclipse.jdt.ui.PreferenceConstants;
 import org.eclipse.jdt.ui.text.IJavaColorConstants;
 import org.eclipse.jdt.ui.text.IJavaPartitions;
 import org.eclipse.jdt.ui.text.JavaSourceViewerConfiguration;
+import org.eclipse.jdt.ui.text.folding.DefaultJavaFoldingStructureProvider.JavaProjectionAnnotation;
 
 import org.eclipse.jdt.internal.ui.text.SmartBackspaceManager;
 import org.eclipse.jdt.internal.ui.text.java.CompletionProposalComputerRegistry;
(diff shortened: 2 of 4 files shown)


fedejeanne:
@danthe1st here's an alternative that doesn't require any new API in platform. I prefer to keep it in JDT since the base classes were already there and the change is rather localized. Providing a new API for this sole use-case (JDT) is IMO unnecessary.

org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/javaeditor/JavaSourceViewer.java
```diff
@@ -178,10 +183,39 @@ private boolean doOperationCached(int operation) {
...
+		if (model == null) {
+			return;
+		}
+
+		for (Iterator<Annotation> it= model.getAnnotationIterator(); it.hasNext();) {
+			Annotation annotation= it.next();
+			if (!(annotation instanceof ProjectionAnnotation projectionAnnotation) || projectionAnnotation.isCollapsed())
+				continue;
+
+			if (annotation instanceof JavaProjectionAnnotation javaAnnotation && javaAnnotation.isTopLevelType())
```
danthe1st, line 212: Nitpick:
I think checking against both `JavaProjectionAnnotation` and `ProjectionAnnotation` is unnecessary.

Alternative:
```java
if (!(annotation instanceof JavaProjectionAnnotation javaAnnotation)) {
    continue;
}
if (javaAnnotation.isCollapsed() || javaAnnotation.isTopLevelType()) {
    continue;
}
```
or combine both:
```java
if (!(annotation instanceof JavaProjectionAnnotation javaAnnotation) || javaAnnotation.isCollapsed() || javaAnnotation.isTopLevelType()) {
    continue;
}
```
Also, using curley braces around the `return` but not around the `continue;` seems a bit inconsistent but I don't care about that.

org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/javaeditor/JavaSourceViewer.java
```diff
@@ -178,10 +183,39 @@ private boolean doOperationCached(int operation) {
...
+				continue;
+
+			model.collapse(projectionAnnotation);
```
danthe1st, line 215: AFAIK calling `model.collapse(Annotation)` results in a `fireModelChanged()` call for each annotation.

I think we should check whether this could cause screen flickering and/or a performance regression with large files.

org.eclipse.jdt.ui/ui/org/eclipse/jdt/ui/text/folding/DefaultJavaFoldingStructureProvider.java
```diff
@@ -326,8 +326,10 @@ public boolean collapseCustomRegions() {
 
 	/**
 	 * A {@link ProjectionAnnotation} for java code.
+	 *
+	 * @noextend This class is not intended to be subclassed by clients.
```
danthe1st, line 330: Is the `@noextend` necessary for a `final` class?

danthe1st (commented):
If this doesn't cause screen flickering or performance issues, I would prefer that approach. I'll mark the other PRs as a draft for now and close them if we go forward with this.

org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/folding/FoldingTest.java
```diff
@@ -824,7 +823,26 @@ public void run() {
 		FoldingTestUtils.assertContainsCollapsedRegionUsingStartAndEndLine(regions, str, 10, 20); // class A2
 		FoldingTestUtils.assertContainsCollapsedRegionUsingStartAndEndLine(regions, str, 12, 17); // Runnable
 		JavaPlugin.getDefault().getPreferenceStore().setToDefault(PreferenceConstants.EDITOR_FOLDING_INNERTYPES);
+	}
 
+	/**
+	 * "Collapse All" should not collapse top-level types, so that classes are not left completely
+	 * empty. See <a href="https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/3126">GitHub Issue
```
danthe1st, line 830: (Technically, it's a PR and not an issue but I guess it can be treated as one)

org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/folding/FoldingTest.java
```diff
@@ -824,7 +823,26 @@ public void run() {
...
+	@Test
+	public void testCollapseAllSkipsTopLevelType() throws Exception {
+		String str= """
+				class MyClass {
+					void someMethod() {
+						// does some work
+					}
+				}
+				""";
+		FoldingTestUtils.collapseAll(packageFragment, str);
```
danthe1st, line 842: The `FoldingTestUtils.collapseAll` seems to create the compilation unit and then `FoldingTestUtils.getProjectionRangesOfPackage` creates it again? That doesn't really sound right to me.

org.eclipse.jdt.ui/ui/org/eclipse/jdt/ui/text/folding/DefaultJavaFoldingStructureProvider.java
fedejeanne: Good point, I can remove it if I decide to carry one with this PR.

org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/javaeditor/JavaSourceViewer.java
fedejeanne: Wouldn't that leave out subclasses of `ProjectionAnnotation` e.g. `org.eclipse.ui.internal.genericeditor.folding.IndentFoldingStrategy.FoldingAnnotation`?

Image

If I change it then a `FoldingAnnotation f` such that `f.isCollapsed() == false` would be skipped i.e. not folded.

Or am I missing something?

org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/javaeditor/JavaSourceViewer.java
fedejeanne: Good call, I think my solution is too expensive because of that. 

I sampled this PR, the `master` state from which this PR started and also your PR (#3126 ) and the performance degradation introduced by this PR is bad, really bad.

Here are the samples:
- [Samples.zip](https://github.com/user-attachments/files/31776868/Samples.zip)

In the names you can already see the total times in milliseconds:

Image

i.e. your PR introduces a tiny degradation but mine introduces a huge degradation.

I used this snippet to generate the classes and ran it with `NUM_CLASSES` set to `1_000` and to `10_000` so the generated classes have 3K and 30K foldings respectively:

<details><summary>snippet</summary>

```java
import java.io.IOException;
import java.io.PrintWriter;
import java.nio.file.Paths;

/**
 * Generates a MyLargeClass.java file containing one top-level class
 * (MyLargeClass) and NUM_CLASSES nested classes named A1, A2, A3, ...
 *
 * Run with: java GenerateMyLargeClass Output: MyLargeClass.java (written next
 * to this program)
 */
public class GenerateMyLargeClass {

	private static final int NUM_CLASSES = 10_000;
	private static final String OUTPUT_FILE_NAME = "MyLargeClass_" + NUM_CLASSES + ".java";

	public static void main(String[] args) throws IOException {
		String outputPath = Paths.get(OUTPUT_FILE_NAME).toAbsolutePath().toString();
		try (PrintWriter writer = new PrintWriter(outputPath, "UTF-8")) {
			writer.println("class MyLargeClass_" + NUM_CLASSES + " {");
			writer.println();

			for (int n = 1; n <= NUM_CLASSES; n++) {
				writeNestedClass(writer, n);
				writer.println();
			}

			writer.println("}");
		}

		System.out.println("Generated " + NUM_CLASSES + " nested classes in " + outputPath);
	}

	private static void writeNestedClass(PrintWriter writer, int index) {
		writer.println("\tclass A" + index + " {");
		writer.println("\t\tint i;");
		writer.println();
		writer.println("\t\tvoid someMethod() {");
		writer.println("\t\t\t// Do something");
		writer.println("\t\t\tfor (int i = 0; i < 10; i++) {");
		writer.println("\t\t\t\t// Do something else here too");
		writer.println("\t\t\t}");
		writer.println("\t\t}");
		writer.println("\t}");
	}
}
```

</details>

fedejeanne (commented):
@danthe1st I analyzed the performance as you suggested and I am inclined to drop this PR and carry on with yours because of the performance. I will decide this week.

Thank you for the reviewing, good call on the performance.

org.eclipse.jdt.ui/ui/org/eclipse/jdt/ui/text/folding/DefaultJavaFoldingStructureProvider.java
danthe1st: Possible alternative to making it `public`: Add an interface for getting the Java element or checking if it's a top level type somewhere in an internal package.

org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/javaeditor/JavaSourceViewer.java
danthe1st: Do you think there is a way to only do the `fireModelChanged` once without a modification to Platform UI or temporarily disabling redrawing?

org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/javaeditor/JavaSourceViewer.java
danthe1st: I thought there was an `||` in the second `if`, my bad

org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/javaeditor/JavaSourceViewer.java
fedejeanne: Honestly, I think adding a new API to platform and using your proposal would be cleaner. But thank you for the suggestion!

Issue #1445, opened 2024-06-10
SaveParticipantTest.testFormatChange* failed with additional whitespace


jukzi:
SaveParticipantTest

4 fails in https://download.eclipse.org/eclipse/downloads/drops4/I20240608-0320/testresults/html/org.eclipse.jdt.ui.tests_ep433I-unit-mac64-java17_macosx.cocoa.x86_64_17.html
for example:

```
expected:<... { String s [= (String) ]o; } }> but was:<... { String s [ = (String)]o; } }>

org.junit.ComparisonFailure: expected:<... {
String s [= (String) ]o;
}
}> but was:<... {
String s [ = (String)]o;
}
}>
at org.junit.Assert.assertEquals(Assert.java:117)
at org.junit.Assert.assertEquals(Assert.java:146)
at org.eclipse.jdt.ui.tests.quickfix.SaveParticipantTest.testFormatChanges01(SaveParticipantTest.java:152)
at java.base/java.lang.reflect.Method.invoke(Method.java:568)
```

![image](https://github.com/eclipse-jdt/eclipse.jdt.ui/assets/51790620/6cfec0c5-b76f-429c-84e0-b0e053685d4a)

jukzi:
Locally **randomly** reproducible sometimes on  various test methods. examples:
![image](https://github.com/eclipse-jdt/eclipse.jdt.ui/assets/51790620/8ff87ca2-66e7-4bc8-8823-af3471faf02a)
![image](https://github.com/eclipse-jdt/eclipse.jdt.ui/assets/51790620/554126e3-d6db-489e-90cd-e285759ac58e)

jukzi:
again on mac:
https://download.eclipse.org/eclipse/downloads/drops4/I20240619-1800/testresults/html/org.eclipse.jdt.ui.tests_ep433I-unit-mac64-java17_macosx.cocoa.x86_64_17.html

jukzi:
duplicate: https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/79 (also only mac)

jukzi:
single test fail on linux:
https://download.eclipse.org/eclipse/downloads/drops4/I20240619-1800/testresults/html/org.eclipse.jdt.ui.tests_ep433I-unit-cen64-gtk3-java21_linux.gtk.x86_64_21.html
```
expected:<...* A block comment on[] two lines */ ...> but was:<...* A block comment on[ * ] two lines */ ...>

[...]

PR #3140, opened 2026-09-02
Wait for LastSaveReferenceProvider job before making test changes in SaveParticipantTest

trancexpress:
Tracings during `SaveParticipantTest` indicate that `LastSaveReferenceProvider` jobs can change text buffer contents in parallel to the test case. So we wait on these jobs, to avoid jobs setting unexpected contents in the text buffer.

See: https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1445


trancexpress:
This will need a rerun of the tests when the next I-build is available. The change in platform UI is still not visible.

PR #3142, opened 2026-09-02
Invalidate semantic highlighting once per preference burst

vogella:
A style change of a semantic or syntax highlighting invalidated the text presentation once per highlighted position, and each ranged invalidation merges into the whole style range list of the StyledText. A theme switch changes every colour key of every open Java editor twice, so with a few large files open the UI thread was busy for minutes in SemanticHighlightingPresenter.highlightingStyleChanged.

The presenter now invalidates a single span from the first to the last affected position, and the manager coalesces all style changes of one UI event loop turn into one asynchronous invalidation and reconciler refresh. This mirrors the coalescing of AbstractTextEditor.handlePreferenceStoreChanged on the platform side and reduces a theme switch to one re-style per editor.

Verified by compiling the bundle and running the SemanticHighlightingTest, AutoboxingSemanticHighlightingTest and Java25SemanticHighlightingTest suites headless.


vogella:
@jjohnstn can you review? This speeds up significantly theme switching.

PR #3145, opened 2026-09-03
Issue 3128 reload imported test results

### LINKED ISSUE #3128: Add refresh button to junit view
When opening a junit results XML file there is no easy way that I know of to get the results view to refresh when the underlying junit XML file has changed.  I have to keep re-opening the same file to load the latest results.  This often happens when running the tests from command line with gradle or maven when I want to view the results.  It would be nice if the view had a refresh button/option that would cause it to re-read the results XML file.

carstenartur:
Fixes #3128.

## What it does

Imported JUnit XML results are currently loaded as one-time snapshots. When the
underlying file is updated by a Maven or Gradle run, reopening it creates another
test-run history entry instead of refreshing the existing one.

This change adds a **Reload Test Run** command to the JUnit view for sessions
imported from a local XML file.

- Retains the normalized source file of an imported test-run session.
- Re-reads and parses the source XML when the reload command is invoked.
- Parses into a new session first and replaces the existing session only after
  the import completed successfully.
- Keeps the replacement at the same position in the test-run history instead of
  adding a duplicate entry.
- Preserves the previously displayed session when the source file cannot be
  read or contains incomplete or invalid XML.
- Enables the reload command only when the active session was imported from a
  local file.
- Adds regression tests for successful replacement and failed reloads.

### Scope

This PR implements the manual reload requested by #3128.

An optional `WatchService`-based automatic reload was also discussed in the
issue. It is intentionally left for a separate follow-up so that this change
remains focused and independently reviewable. Manual reload would remain
necessary as a fallback even with automatic file watching.

## How to test

1. Run tests from Maven or Gradle and import the generated JUnit XML file into
   the JUnit view.
2. Regenerate or modify the XML file without reopening it in Eclipse.
3. Select the imported test-run session.
4. Invoke **Reload Test Run** from the JUnit view toolbar.
5. Verify that the displayed results are updated and that no additional history
   entry is created.
6. Replace the XML contents with incomplete or invalid XML and invoke reload
   again.
7. Verify that an error is reported and that the previously displayed valid
   session remains unchanged.

Automated coverage is provided by `ImportedTestRunReloadTest`, including:

- replacement at the same test-run history position;
- retention of the original source file for the replacement session;
- preservation of the previous session after a failed XML parse.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.junit/plugin.xml
+++ b/org.eclipse.jdt.junit/plugin.xml
@@ -413,7 +413,7 @@
             locationURI="toolbar:org.eclipse.jdt.junit.ResultView">
          <command
                commandId="org.eclipse.jdt.junit.reloadImportedTestRun"
-               icon="$nl$/icons/full/elcl16/relaunch.svg"
+               icon="$nl$/icons/full/elcl16/refresh.svg"
                id="org.eclipse.jdt.junit.reloadImportedTestRun.toolbar"
                label="%ReloadTestRunAction.label"
                style="push"
--- a/org.eclipse.jdt.junit/plugin.xml
+++ b/org.eclipse.jdt.junit/plugin.xml
@@ -383,6 +383,43 @@
             categoryId="org.eclipse.debug.ui.category.run"
             id="org.eclipse.jdt.junit.junitShortcut.rerunFailedFirst">
       </command>
+      <command
+            name="%ReloadTestRunAction.label"
+            description="%ReloadTestRunAction.tooltip"
+            categoryId="org.eclipse.debug.ui.category.run"
+            id="org.eclipse.jdt.junit.reloadImportedTestRun">
+      </command>
+   </extension>
+
+   <extension
+         point="org.eclipse.ui.handlers">
+      <handler
+            class="org.eclipse.jdt.internal.junit.ui.ReloadTestRunHandler"
+            commandId="org.eclipse.jdt.junit.reloadImportedTestRun">
+         <activeWhen>
+            <with
+                  variable="activePartId">
+               <equals
+                     value="org.eclipse.jdt.junit.ResultView">
+               </equals>
+            </with>
+         </activeWhen>
+      </handler>
+   </extension>
+
+   <extension
+         point="org.eclipse.ui.menus">
+      <menuContribution
+            locationURI="toolbar:org.eclipse.jdt.junit.ResultView">
+         <command
+               commandId="org.eclipse.jdt.junit.reloadImportedTestRun"
+               icon="$nl$/icons/full/elcl16/relaunch.svg"
+               id="org.eclipse.jdt.junit.reloadImportedTestRun.toolbar"
+               label="%ReloadTestRunAction.label"
+               style="push"
+               tooltip="%ReloadTestRunAction.tooltip">
+         </command>
+      </menuContribution>
    </extension>
 
    <extension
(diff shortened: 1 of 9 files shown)


org.eclipse.jdt.junit/plugin.xml
```diff
@@ -383,6 +383,43 @@
...
+      </handler>
+   </extension>
+
+   <extension
+         point="org.eclipse.ui.menus">
+      <menuContribution
+            locationURI="toolbar:org.eclipse.jdt.junit.ResultView">
+         <command
+               commandId="org.eclipse.jdt.junit.reloadImportedTestRun"
+               icon="$nl$/icons/full/elcl16/relaunch.svg"
```
jjohnstn: Make this the refresh icon.  refresh.svg

jjohnstn (changes requested):
Works well, however, if I import an xml file, then use the history to switch to another test, then switch back to the imported test, the new reload button is disabled.  I think the enablement should be able to track this.  Also, please comment on copilot review comments to note whether they are resolved, no longer valid, obsoleted, or they are wrong.  Feel free to mark them resolved in such cases.  If you are still considering them or haven't implemented them yet, they should remain open.

jjohnstn:
Hmm.  Some further testing and the button is enabled again and I can't get it to disable.  I do notice it is enabled for other tests in the history that were run as JUnit tests.  Should this be the case?

org.eclipse.jdt.junit/plugin.xml
carstenartur: The icon is copied unchanged from org.eclipse.jdt.ui/icons/full/elcl16/refresh.svg and is included in the built JUnit UI bundle.

carstenartur:
> Works well, however, if I import an xml file, then use the history to switch to another test, then switch back to the imported test, the new reload button is disabled.  I think the enablement should be able to track this.  Also, please comment on copilot review comments to note whether they are resolved, no longer valid, obsoleted, or they are wrong.  Feel free to mark them resolved in such cases.  If you are still considering them or haven't implemented them yet, they should remain open.

Thanks for testing this. The history-switch enablement issue is fixed.

Switching between history entries changes the view's input without changing
the active workbench part. The view now requests an enablement re-evaluation
after each session change, so Reload is enabled when returning to a
file-imported session and disabled when selecting a session without an
import source.

I also replaced the relaunch icon with the existing refresh.svg.

The Copilot finding was valid and has been fixed: a missing source-file
mapping now causes reloadTestRunSession(...) to return null rather than
throw an assertion failure. This is covered by
testReloadRemovedSessionReturnsNull(), and that thread is now marked resolved.

carstenartur:
> Hmm.  Some further testing and the button is enabled again and I can't get it to disable.  I do notice it is enabled for other tests in the history that were run as JUnit tests.  Should this be the case?

No, that is not intended. Reload should only be enabled for sessions imported
from a local XML file with a known source file. It must be disabled for tests
launched normally from Eclipse. Reloading imported results is distinct from
rerunning a test.

The stale enablement could leave the button stuck in either state; the fix
now re-evaluates it when the selected history entry changes.

To cover your specific observation, there is
testReloadEnablementAfterJUnitLaunch(). It launches an actual JUnit 4 test,
verifies its successful completion, and repeatedly switches between that
session and an imported XML session. It checks the actual toolbar button:
disabled for the launched test, enabled for the imported session, including
after reloading the XML. The test does not manually refresh the handler
between selections.

The regression fails with the enablement fix removed and passes with it
present. In the dedicated verification run, all five targeted reload tests
passed, as did the full JUnitJUnitTests suite: 207 passed, 5 skipped, no
failures or errors.

jjohnstn:
Further testing confirms that imported files from before were the culprit.  Older files I imported don't work as expected, but with the newest changes, files imported work fine and regular tests don't enable the refresh button.

jjohnstn (approved):
Approved.

jjohnstn:
@carstenartur Would you like to write a N&N for this new feature?

carstenartur:
Thanks for the suggestion, @jjohnstn! I've submitted a New & Noteworthy entry for Eclipse 4.42 in https://github.com/eclipse-platform/www.eclipse.org-eclipse/pull/612.

It describes the `Reload Test Run` command for locally imported XML results, including updating the existing history entry without creating duplicates and preserving the previous results if reloading fails. It also clarifies that reloading refreshes the results rather than rerunning the tests.

Thanks again for reviewing and testing this feature!

Issue #2704, opened 2025-12-19
Fails in TestRunListenerTest5


trancexpress:
Must be a regression from either https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2702 or https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2696.

See: https://download.eclipse.org/eclipse/downloads/drops4/I20251218-1800/testresults/html/org.eclipse.jdt.ui.tests_ep439I-unit-linux-x86_64-java25_linux.gtk.x86_64_25.html

```
org.eclipse.jdt.junit.tests.TestRunListenerTest5	testTerminateLaunch	Failure	Expected JUnit update jobs to be scheduled

java.lang.AssertionError: Expected JUnit update jobs to be scheduled
at org.junit.Assert.fail(Assert.java:89)
at org.junit.Assert.assertTrue(Assert.java:42)
at org.eclipse.jdt.junit.tests.TestRunListenerTest5.testTerminateLaunch(TestRunListenerTest5.java:276)
at java.base/java.lang.reflect.Method.invoke(Method.java:565) 
```

```
	expected:<...TestCase state: [Completed result: Failure testSuiteClass: pack.ATestCase state: Completed result: Failure testCaseMethod: testFail class: pack.ATestCase state: Completed result: Failure exception: org.opentest4j.AssertionFailedError actual: null expected: null] > but was:<...TestCase state: [Running result: Undefined testSuiteClass: pack.ATestCase state: Running result: Undefined testCaseMethod: testFail class: pack.ATestCase state: Running result: Undefined] >

org.junit.ComparisonFailure: expected:<...TestCase
state: [Completed
result: Failure

testSuiteClass: pack.ATestCase
state: Completed
result: Failure

testCaseMethod: testFail
class: pack.ATestCase
state: Completed
result: Failure
exception: org.opentest4j.AssertionFailedError
actual: null
expected: null]

> but was:<...TestCase
state: [Running
result: Undefined

testSuiteClass: pack.ATestCase
state: Running
result: Undefined

testCaseMethod: testFail
class: pack.ATestCase
state: Running
result: Undefined]

>
at org.junit.Assert.assertEquals(Assert.java:117)
at org.junit.Assert.assertEquals(Assert.java:146)
at org.eclipse.jdt.junit.tests.AbstractTestRunListenerTest.assertEqualLog(AbstractTestRunListenerTest.java:233)
[...]

PR #3163, opened 2026-09-09
Bug 569992 - Report accurate JUnit test execution times

carstenartur:
## What it does

This change addresses [Bug 569992](https://bugs.eclipse.org/bugs/show_bug.cgi?id=569992) by measuring individual JUnit test execution time in the remote test VM instead of reconstructing it from protocol-message arrival times in the Eclipse process.

It also adds optional CPU-time diagnostics, preserves those diagnostics across JUnit history swapping and restart, and makes rerun timing consistent with the normal execution path.

### Measure elapsed time in the test VM

Previously, individual test timings were reconstructed in the Eclipse process from `TEST_START` and `TEST_END` using `System.currentTimeMillis()`.

That has two independent problems:

- communication and scheduling delays between the test VM and Eclipse can become part of the reported duration;
- `System.currentTimeMillis()` is a wall-clock source rather than a monotonic elapsed-time source, so clock adjustments can produce incorrect durations.

The remote runner now measures each individual test with `System.nanoTime()` and reports the measurement through an optional `%TSTTIME` protocol message.

`ITestElement#getElapsedTimeInSeconds()` keeps its existing meaning: elapsed wall-clock execution time. With a new runner/client pair, however, it is measured where the test executes rather than inferred from message-arrival times.

For compatibility:

- `%TSTTIME` is an optional protocol extension;
- older runners continue to work;
- when remote timing is absent, the client falls back to a monotonic `System.nanoTime()` measurement;
- malformed or invalid optional timing payloads are ignored without affecting authoritative test result/status messages.

### Preserve meaningful suite timing

For the initial test run, suites derive elapsed time from the span between the earliest reported child start and the latest reported child end.

This preserves elapsed wall-clock semantics when tests execute concurrently.

Summing child durations would overstate elapsed time when executions overlap, while reconstructing the interval from client-side protocol arrival would reintroduce communication latency.

### Add optional CPU-time diagnostics

When supported by the test VM, the remote runner also uses `ThreadMXBean` to measure CPU consumption of the execution thread.

It records:

- elapsed wall time;
- total CPU time;
- user-mode CPU time;
- system CPU time, derived as `CPU - user`;
- non-CPU elapsed time, derived as `wall - CPU`.

If current-thread CPU timing is supported but disabled, the runner makes a best-effort attempt to enable it in the launched test VM.

No additional Java system property or launch option is required from the user.

If CPU timing is unsupported or cannot be enabled, elapsed timing still works and CPU details remain unavailable.

The CPU values are diagnostic rather than a new public `ITestElement` API. They describe the thread on which the test execution callbacks are delivered; arbitrary worker threads spawned by the test are not included.

If start and end callbacks occur on different threads, CPU/user values are not reported.

`non-CPU` is intentionally not labelled as strictly "waiting time": it is the part of elapsed wall time not accounted for by CPU consumption of the measured execution thread.

### Keep timing display preferences independent from recorded data

The JUnit view has two independent presentation options:

- `Show Execution Time`
- `Show Execution Time Details`

The details option is off by default, preserving the existing compact presentation unless explicitly enabled.

These options are view preferences only.

They do not control whether timing data is measured or stored, and changing them does not modify historical test-run data.

This distinction is important for history: a test run retains the timing information that was recorded when that run executed, while the current view settings merely determine how the selected run is displayed.

### Preserve timing details in JUnit history

CPU and user CPU values are serialized as optional attributes in the existing JUnit test-run XML format.

This is the same serialization boundary used by the persisted and lazily loaded JUnit history.

Consequently, switching from a recent run to an older history entry and back no longer loses the timing details of the recent run. The data also survives persisted-history reloads across restart.

Older or externally imported XML without the new optional attributes remains compatible and simply has no CPU details.

The history index format itself is unchanged.

### Measure keep-alive reruns freshly

JUnit keep-alive mode reruns a test inside the existing test VM and uses `RerunExecutionListener`, which deliberately suppresses the normal transient `TEST_START` / `TEST_END` IDs.

The rerun path now reuses the same timing implementation as the initial run.

Fresh timing is associated with the original test ID and delivered before `TEST_RERAN`, so it is already available when the synchronous model/view update for the rerun occurs.

If no valid fresh timing is received, timing values from the previous execution are invalidated instead of being left visible as stale data.

A rerun also uses a new monotonic timing origin. Its individual elapsed and CPU values therefore update the rerun test, but its start offset is deliberately not folded into the original test-run suite interval.

Otherwise, the idle period between the original run and a later rerun would incorrectly inflate the suite duration.

## How to test

Regression coverage includes the timing protocol, model behaviour, history serialization and reruns.

### `TestExecutionTiming`

Covers:

- normal `TEST_START` → `%TSTTIME` → `TEST_END` ordering;
- valid elapsed, CPU and user CPU values;
- the `user <= CPU` invariant;
- unavailable CPU timing;
- fresh timing for keep-alive reruns;
- reuse of the original test ID for rerun timing;
- reruns updating the individual test without extending the original suite interval;
- reruns without fresh valid timing invalidating timing from the previous execution;
- propagation through a real JUnit 5 launch containing both CPU work and `Thread.sleep()`.

### `TestExecutionTimingProtocol`

Covers:

- valid timing payloads;
- unavailable CPU/user values;
- malformed payloads;
- invalid numeric ranges;
- overflow-sensitive timing intervals;
- compatibility with result streams containing no timing message.

### `TestRunSessionHistoryTests`

Verifies that elapsed, CPU and user CPU values survive history serialization and lazy reload, and that the derived system and non-CPU values remain correct afterwards.

### Existing serialization tests

The existing JUnit 3 and JUnit 4 XML golden tests already normalize the dynamic `time` value.

They now also normalize the optional, VM-dependent `cpuTime` and `userTime` attributes when comparing against historical golden XML files.

Persistence of those attributes is tested separately by the dedicated history round-trip tests, so the golden tests can continue to validate the stable XML structure independently of machine-dependent timing values.

## Author checklist


jjohnstn:
@carstenartur Works well.  I might suggest that the new time break-down be optional in addition to the current simple  "Show Execution Time".  I think that you could have "Show Execution Time" and "Show Execution Time Details".  Both could be selected or neither.

carstenartur:
> @carstenartur Works well.  I might suggest that the new time break-down be optional in addition to the current simple  "Show Execution Time".  I think that you could have "Show Execution Time" and "Show Execution Time Details".  Both could be selected or neither.

Thanks, that makes sense. I’ve split the presentation into two independent, persisted options: Show Execution Time and Show Execution Time Details.

The new details option is off by default, so the existing compact elapsed-time presentation remains unchanged unless it is explicitly enabled. Both options can be selected independently, or neither can be selected.

I also added regression coverage for all four display combinations, while keeping CPU timing optional when the VM does not provide it.

jjohnstn:
Hi @carstenartur That's better.  The only thing left is history support.  If I switch between the latest run with timing details to an old run, then switch back, the details are gone and can't be recovered, regardless of the setting.

carstenartur:
> Hi @carstenartur That's better.  The only thing left is history support.  If I switch between the latest run with timing details to an old run, then switch back, the details are gone and can't be recovered, regardless of the setting.

Thanks, this exposed an integration gap with the history work from #3137.

The history implementation swaps sessions out through the existing JUnit XML serializer and lazily imports them again. The new CPU/user timing values were initially only stored in the live `TestElement`, so selecting another history entry caused them to be lost when the current session was swapped out.

I now persist `cpuTime` and `userTime` as optional test-case attributes in that existing XML format and restore them on import. This covers both switching between history entries and persisted history across restart. Older or imported XML remains compatible and simply has no CPU details.

The two view options remain presentation-only: changing `Show Execution Time` or `Show Execution Time Details` does not alter stored run data. The currently selected settings are applied when displaying any historical run.

While following the lifecycle through reruns, I found another related edge case in keep-alive mode. `RerunExecutionListener` deliberately suppresses the transient start/end IDs, so the timing implementation did not initially produce a fresh measurement there.

The rerun path now reuses the same measurement implementation as the initial run. Fresh timing is associated with the original test ID and delivered before `TEST_RERAN`, so it is available to the synchronous model/view update.

If no valid fresh timing is available, the previous execution's timing is explicitly invalidated instead of being left behind as stale data.

The rerun's monotonic start offset is also deliberately not folded into the original suite interval, because it uses a new timing origin and the idle period between the initial run and a later rerun must not inflate the suite duration.

I added regression coverage for the history lazy-reload path, fresh keep-alive rerun timing, preserving the original suite interval, and avoiding stale timing when a rerun has no fresh measurement.

The history change also exposed an existing XML-golden-test assumption: those tests already normalize the dynamic `time` attribute, but did not account for the newly persisted, VM-dependent `cpuTime` and `userTime` attributes. The golden comparison now normalizes those optional values as well, while the dedicated history tests continue to verify that the actual CPU timing values survive serialization and reload.

carstenartur:
@jjohnstn A quick correction from my side, to avoid creating unnecessary noise around this PR.

While investigating the previous Jenkins failure in `TestRunListenerTest5.testTerminateLaunch`, I followed the existing #2704 issue and opened #3174 for the remaining scheduling race in that test. In retrospect, I moved too quickly there.

The important result is that the Jenkins rerun of this PR is now green without that separate change. This confirms that the `testTerminateLaunch` failure was unrelated to the timing/history changes in #3163.

I also noticed that #3154 already contains a more substantial stabilization of the same termination test, so keeping another overlapping PR open for this investigation would create more parallel review work than it solves.

I am therefore going to close #3174. Please disregard my earlier request to look at that separate fix; there is no additional dependency for #3163.

For #3163 itself, the history persistence fix and the serialization-test adjustment are now covered by a successful upstream Jenkins run. Thanks again for pointing out the history problem — that was a real integration gap and the review helped uncover it.

jjohnstn (approved):
Works well.  Imported test runs without cpu timings works fine - the option is ignored and the normal time is shown.  Switching in history works as expected with details still available when switching to a freshly run test.

Issue #3125, opened 2026-08-21
JavaLeakTest tests failed on I20260820-0719


noopur2507:
The following tests failed on I20260820-0719:

macOS, ARM 64‑bit, Java 25
org.eclipse.jdt.ui.tests.leaks.JavaLeakTest 

testJavaEditorCloseAll
testJavaEditorContextMenu
testTextEditorCloseOneOfTwo

Cannot access 'private java.lang.constant.ClassDesc java.lang.classfile.TypeKind.upperBound' in package 'java.lang.classfile'. JVM settings for --add-modules, --add-opens, and --illegal-access are probably missing. see 'javavmargs' in test.xml
...

trancexpress:
There were a lot of fails here:

* https://download.eclipse.org/eclipse/downloads/drops4/I20260908-2300/testresults/html/org.eclipse.jdt.ui.tests_ep442I-unit-macosx-x86_64-java25_macosx.cocoa.x86_64_25.html
* https://download.eclipse.org/eclipse/downloads/drops4/I20260908-2300/testresults/html/org.eclipse.jdt.ui.tests_ep442I-unit-win32-x86_64-java25_win32.win32.x86_64_25.html

E.g.:

```
	Cannot access 'private java.lang.constant.ClassDesc java.lang.classfile.TypeKind.upperBound' in package 'java.lang.classfile'. JVM settings for --add-modules, --add-opens, and --illegal-access are probably missing. see 'javavmargs' in test.xml

java.lang.RuntimeException: Cannot access 'private java.lang.constant.ClassDesc java.lang.classfile.TypeKind.upperBound' in package 'java.lang.classfile'. JVM settings for --add-modules, --add-opens, and --illegal-access are probably missing. see 'javavmargs' in test.xml
at org.eclipse.jdt.ui.leaktest.reftracker.ReferenceTracker.followFieldReference(ReferenceTracker.java:77)
at org.eclipse.jdt.ui.leaktest.reftracker.ReferenceTracker.visit(ReferenceTracker.java:149)
at org.eclipse.jdt.ui.leaktest.reftracker.ReferenceTracker.start(ReferenceTracker.java:173)
at org.eclipse.jdt.ui.leaktest.LeakTestCase.collect(LeakTestCase.java:62)
at org.eclipse.jdt.ui.leaktest.LeakTestCase.collect(LeakTestCase.java:56)
[...]

PR #3167, opened 2026-09-09
Generate necessary --add-opens VM arguments for leak tests

trancexpress:
This change adjusts the `org.eclipse.jdt.ui.tests` `pom.xml` to generate a properties file, which contains `--add-opens` arguments for all JDK modules of the currently used Java version.

The properties file is used in the `pom.xml` and `test.xml`, to pass the arguments to the tests. In addition, the leak tests are re-enabled in `pom.xml`.

See: https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/3125


trancexpress:
The `pom.xml` changes seem to work, at least when I run `mvn clean verify` the tests are starting with a long string of VM arguments.

I have no idea how to test the changes in `test.xml` though:

```
➜  org.eclipse.jdt.ui.tests git:(gh3125) ant -f test.xml 
Buildfile: /data/git/eclipse/eclipse.jdt.ui/org.eclipse.jdt.ui.tests/test.xml

init:

leaksuite:

BUILD FAILED
/data/git/eclipse/eclipse.jdt.ui/org.eclipse.jdt.ui.tests/test.xml:51: The following error occurred while executing this line:
Error reading project file /data/git/eclipse/plugins/org.eclipse.test/library.xml: /data/git/eclipse/plugins/org.eclipse.test/library.xml

Total time: 0 seconds
```

I'm also not sure the phase in which we generate the new properties file is the right one: https://maven.apache.org/guides/introduction/introduction-to-the-lifecycle.html

`process-test-resources` is listed to be after `compile`, but I have no idea if we run that phase during I-builds. I guess I need to run the I-build tests locally, hopefully its just `mvn clean verify` from the root of the aggregator repository as listed here: https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/master/README.md

trancexpress:
I also need to check why the leak tests were commented out in the first place. Maybe it wasn't because of the (frequent) need to update the `--add-opens` VM arguments.

And probably we want some check during the tests (i.e. at runtime) that the VM arguments do include all the `--add-opens` we want. To check if its possible via:

```
RuntimeMXBean runtimeMxBean = ManagementFactory.getRuntimeMXBean();
List<String> arguments = runtimeMxBean.getInputArguments();
```

carstenartur:
Thanks for working on this. I like the general direction a lot: generating the `--add-opens` arguments from the JDK in use seems much more maintainable than continuing to extend the large static list whenever the JDK changes.
In particular, I think it is good that this approach keeps the `ALL-UNNAMED` semantics and does not rely on private JDK methods such as `Module.implAddExportsOrOpens(...)`.
There are a few things I think we should address before merging this, though.
## 1. `ModuleLayer.boot()` does not necessarily describe all system modules
The generator currently uses:
    ModuleLayer.boot().modules()
but the generated arguments also contain:
    --add-modules ALL-SYSTEM
Those two sets are not necessarily the same.
`ModuleLayer.boot()` contains the modules which have already been resolved into the boot layer of the generator process, whereas `ALL-SYSTEM` can cause additional system modules to be resolved in the test VM.
So it seems possible that we generate `--add-opens` for fewer modules than the test VM will eventually contain.
I think `ModuleFinder.ofSystem().findAll()` together with each module's `ModuleDescriptor.packages()` would match the intention more closely: inspect the packages that actually exist in the current JDK, independently of which modules happened to be resolved for the generator itself.
It would probably also be useful to sort modules and packages before generating the output so that the resulting file is deterministic.
## 2. I am not sure the generated properties file can work for the `test.xml` / I-build path
This seems particularly important because the original failure in #3125 happened in an Eclipse I-build.
`test.xml` currently reads:
    ${basedir}/target/vmarguments.properties
That works when Maven has just generated the file in the source checkout.
However, my understanding of the Eclipse automated-test infrastructure is that `test.xml` is later executed from the installed/extracted test bundle. The `target` directory is not part of that bundle, so the generated properties file would not be available there.
There is another related issue: even if we packaged the generated file, it would describe the JDK which built the bundle. The automated tests may subsequently be run using another JDK. In #3125, for example, what matters is the Java 25 installation used to execute the tests.
Ideally the argument generation for the `test.xml` path therefore needs to happen using the actual test JVM shortly before that JVM is launched.
I think this is worth verifying with the releng/I-build path before we consider the issue fixed.
## 3. I would avoid carrying historical package names as exceptions unless they are demonstrably required
The remaining `additionalVmArguments` contain entries such as:
    java.base/jdk.internal.util.jar
    java.base/sun.io
    java.base/sun.security.action
and several internal ASM packages.
Some of these are exactly the kind of entries for which recent Java 25 test runs already print warnings that the package no longer exists in the specified module.
One of the main advantages of generating the list from the current JDK should be that obsolete entries disappear automatically.
So I would prefer something like:
    packages actually present in the test JDK
    + a very small, documented set of genuine exceptions if required
rather than preserving entries merely because they existed in the previous static list.
If an old entry is missing from `ModuleFinder.ofSystem()`, I think we should first determine whether it is still a real package and whether the leak tracker can actually reach it before keeping it as an exception.
## 4. Command-line length may become relevant, especially on Windows
Opening every package in every JDK module produces a very large argument string.
That might be fine on Linux/macOS, but Windows has significantly tighter command-line size limits. Since this problem already occurs in the Windows Eclipse test jobs as well, I think we should make sure the generated argument list cannot make process creation fail there.
A Java argument file might be a good fit here.
Instead of putting tens of thousands of characters directly into `argLine`, the generator could produce a file containing the VM arguments and the launcher would only need something similar to:
    @path/to/leak-test-vmargs
That would also make the generated result easier to inspect when diagnosing future JDK changes.
## 5. Re-enabling `LeakTestSuite` in the Tycho build should probably be investigated separately
The PR currently also changes:
    
back to an active include.
I noticed you already mentioned this in your comment, and I agree that we should first find out why it was disabled.
It may well have been disabled because the module-access configuration had become unreliable, but if there was another reason, this PR could accidentally reintroduce an unrelated problem.
I would either establish the history/reason before enabling it, or keep that change separate from the `--add-opens` generation itself.
## 6. A runtime verification would be useful, but I would test the effective module state rather than only the command line
Checking `RuntimeMXBean.getInputArguments()` can tell us which strings were passed to the VM, but it does not necessarily prove that the desired package is actually open.
For example, an invalid `--add-opens` can still appear in the input arguments while the JVM merely emits a warning.
A stronger regression check could use the public module API, for example conceptually:
    module.isOpen(packageName, ReferenceTracker.class.getModule())
That would verify the property the leak tests actually rely on: that the package is open to the module containing the tracker.
## Overall
So overall I think the architecture of this PR is a substantial improvement over adding another handful of entries to the existing static list.
I would mainly want to make sure that:
1. we enumerate the actual system modules rather than only the current boot layer;
2. generation also works for the real I-build / `test.xml` execution path and uses the test JDK;
3. obsolete historical entries are not retained unnecessarily;
4. the generated argument set remains safe on Windows; and
5. re-enabling `LeakTestSuite` is intentional and not an unrelated behavioural change.
With those points addressed, this looks like it could turn the recurring Java 9/11/17/24/25 maintenance problem into a much more durable solution rather than just fixing the current `java.lang.classfile` failure.

trancexpress:
@carstenartur since you are already asking an AI agent, could you ask the agent to try and implement the suggestions? I'll probably have time only in the evenings so progress will be slow (not that we are in a hurry here).

The agent makes valid points, especially `test.xml` will likely have to generate the properties file on its own - since it will run with another Java version. Using a file for extra VM arguments is also a great suggestion but its not clear to me how to do it, I'll also need to search/ask Copilot how to do that.

carstenartur:
I implemented the suggested follow-up changes.

Since my GitHub integration does not have push access to your fork, I couldn’t update gh3125 directly. I created a small cross-fork PR against your gh3125 branch instead:

https://github.com/trancexpress/eclipse.jdt.ui/pull/1

It contains a single commit based directly on the current head of #3167, so merging it should update this PR without introducing a separate PR against Eclipse master.

The follow-up includes:

* use ModuleFinder.ofSystem() instead of ModuleLayer.boot() so all modules of the actual JDK are considered;
* deterministic module/package ordering;
* remove the historical hard-coded package exceptions;
* generate a Java @argfile rather than putting the complete list on the command line, avoiding the Windows command-line length issue;
* generate the argument file from test.xml using the JVM which is actually running the Eclipse automated tests, so I-builds are not tied to the JDK used during the Maven build;
* add a regression test which verifies effective module openness rather than only checking the command-line arguments;
* keep the Maven LeakTestSuite include in its existing disabled state until we know why it was disabled.

I also checked the current Eclipse releng test path: the test bundle is installed into the test Eclipse instance and its test.xml is invoked from there, so generating the argument file at that point should use the actual test JDK. I checked Tycho 5.0.4’s argLine handling as well; the @argfile remains a single JVM argument.

Locally I exercised the generator, the generated argument file, the positive/negative module-openness check, and a simulated installed-bundle → test.xml → generator → forked JVM path.

Please feel free to merge the small follow-up PR into gh3125 if the approach looks good to you.

trancexpress:
Thank you @carstenartur , this looks great! I'll squash and I'll try to run tests from the aggregator repository.

One question, I've mentioned Copilot as co-author, is that the AI agent you are using?

@iloveeclipse do we mention the AI agent as a co-author of the commit? I don't remember the guidelines for this in JDT/platform.

trancexpress:
@merks one question here, do we keep the version of these 3 lines in the `pom.xml`?

```
        <groupId>org.codehaus.mojo</groupId>
        <artifactId>exec-maven-plugin</artifactId>
        <version>3.6.3</version>
```

Is it OK to omit the version? It seems like something we'd have to update eventually, so easier to omit it?

trancexpress:
From `mvn clean verify` the command line is:

```
[INFO] Command line: /usr/lib/jvm/java-25-openjdk/bin/java -Dosgi.noShutdown=false -Dosgi.os=linux -Dosgi.ws=gtk -Dosgi.arch=x86_64 @/data/git/eclipse/eclipse.jdt.ui/org.eclipse.jdt.ui.tests/target/leak-test-vmargs -XX:CompileCommand=exclude,org.eclipse.jdt.internal.core.dom.rewrite.ASTRewriteAnalyzer::getExtendedRange -Dosgi.clean=true -Declipse.test.screenshot_directory=/data/git/eclipse/eclipse.jdt.ui/org.eclipse.jdt.ui.tests/target/screenshots -ea -jar /home/sandreev/.m2/repository/p2/osgi/bundle/org.eclipse.equinox.launcher/1.8.0.v20260804-1928/org.eclipse.equinox.launcher-1.8.0.v20260804-1928.jar -data /data/git/eclipse/eclipse.jdt.ui/org.eclipse.jdt.ui.tests/target/work/data -install /data/git/eclipse/eclipse.jdt.ui/org.eclipse.jdt.ui.tests/target/work -configuration /data/git/eclipse/eclipse.jdt.ui/org.eclipse.jdt.ui.tests/target/work/configuration -application org.eclipse.tycho.surefire.osgibooter.uitest -testproperties /data/git/eclipse/eclipse.jdt.ui/org.eclipse.jdt.ui.tests/target/surefire.properties -timeout 1200000
```

The file is arguments looking good:

```
➜  org.eclipse.jdt.ui.tests git:(gh3125) head -n 10 /data/git/eclipse/eclipse.jdt.ui/org.eclipse.jdt.ui.tests/target/leak-test-vmargs
--add-modules=ALL-SYSTEM
--add-opens=java.base/com.sun.crypto.provider=ALL-UNNAMED
--add-opens=java.base/com.sun.security.ntlm=ALL-UNNAMED
--add-opens=java.base/java.io=ALL-UNNAMED
--add-opens=java.base/java.lang=ALL-UNNAMED
--add-opens=java.base/java.lang.annotation=ALL-UNNAMED
--add-opens=java.base/java.lang.classfile=ALL-UNNAMED
--add-opens=java.base/java.lang.classfile.attribute=ALL-UNNAMED
--add-opens=java.base/java.lang.classfile.constantpool=ALL-UNNAMED
--add-opens=java.base/java.lang.classfile.instruction=ALL-UNNAMED
```

```
➜  org.eclipse.jdt.ui.tests git:(gh3125) wc -l /data/git/eclipse/eclipse.jdt.ui/org.eclipse.jdt.ui.tests/target/leak-test-vmargs
916 /data/git/eclipse/eclipse.jdt.ui/org.eclipse.jdt.ui.tests/target/leak-test-vmargs
```

iloveeclipse:
> do we mention the AI agent as a co-author of the commit? I don't remember the guidelines for this in JDT/platform.

I've just updated instructions, see https://github.com/eclipse-jdt/.github/blob/main/CONTRIBUTING.md#commit-message-recommendations and https://github.com/eclipse-jdt/.github/blob/main/CONTRIBUTING.md#using-generative-ai-for-contributions.

merks:
> @merks one question here, do we keep the version of these 3 lines in the `pom.xml`?
> 
> ```
>         <groupId>org.codehaus.mojo</groupId>
>         <artifactId>exec-maven-plugin</artifactId>
>         <version>3.6.3</version>
> ```
> 
> Is it OK to omit the version? It seems like something we'd have to update eventually, so easier to omit it?

The version should be managed somewhere. I’m on a phone so I can’t search. Maybe @HannesWell or @akurtakov know the best answer.

trancexpress:
> The version should be managed somewhere. I’m on a phone so I can’t search. Maybe @HannesWell or @akurtakov know the best answer.

From the aggregator repository submodules, only SWT uses this plug-in and it specifies no version (that I see):

```
➜  eclipse.platform.releng.aggregator git:(master) ✗ cat eclipse.platform.swt/pom.xml | grep -A 3 -B 3 exec-maven-plugin
        <plugins>
          <plugin>
            <groupId>org.codehaus.mojo</groupId>
            <artifactId>exec-maven-plugin</artifactId>
            <executions>
              <execution>
                <id>run-javadoc-basher</id>
--
        <plugins>
          <plugin>
            <groupId>org.codehaus.mojo</groupId>
            <artifactId>exec-maven-plugin</artifactId>
            <executions>
              <execution>
                <id>run-jni-generator</id>
--
          <!-- Used by the javadoc-basher and jni-generator profiles to run tools from
               bundles/org.eclipse.swt.tools with the java source-file launcher. -->
          <groupId>org.codehaus.mojo</groupId>
          <artifactId>exec-maven-plugin</artifactId>
          <version>3.6.3</version>
          <configuration>
            <forceJava>true</forceJava>
```

akurtakov:
There is no shared declaration of exec-maven-plugin and SWT has been the first one to use it so it made no sense to declare it in parent pom as there were no other users. Ideally declaration like https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/master/eclipse-platform-parent/pom.xml#L333  should be done.

merks:
I think Dependabot will update versions in the future.

trancexpress:
> There is no shared declaration of exec-maven-plugin and SWT has been the first one to use it so it made no sense to declare it in parent pom as there were no other users. Ideally declaration like https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/master/eclipse-platform-parent/pom.xml#L333 should be done.

OK, please review https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/4039 and let me know what else is missing. E.g. do I need to add version properties in SWT and now in this PR, after the aggregator PR is merged?

akurtakov:
Once the parent is merged skip the version in jdt. I'll take care of swt one.

trancexpress:
I'm having problems building the aggregator repository: https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/4040

So probably I won't be able to run tests overnight.

carstenartur:
> I'm having problems building the aggregator repository: https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/4040
> 
> 
> 
> So probably I won't be able to run tests overnight.

The current state looks good to me, and I don't think the aggregator failure in `eclipse-platform/eclipse.platform.releng.aggregator#4040` should currently block this PR.
The important parts of this change have already been verified locally with Java 25:
- Tycho passes the generated `@argfile` correctly to the test JVM;
- the generated file contains `java.lang.classfile` and the related packages;
- the normal PR checks are green;
- `exec-maven-plugin` version management is now handled centrally by `eclipse-platform/eclipse.platform.releng.aggregator#4039`.
The failure in aggregator #4040 happens earlier during SWT/p2 dependency resolution and appears unrelated to the JDT leak-test changes. Since the requested SWT fragment actually exists in the published repository, I would first try the aggregator build with a refreshed Tycho/p2 cache, e.g. `mvn -U ...`, before changing anything here.
If that still fails, SWT's recent local-build/fragment-resolution changes would be worth investigating separately.
For #3167 itself, I think the remaining useful verification is the actual installed-bundle → `test.xml` → generated argfile → leak-test JVM path. I would prefer testing that path directly rather than making this PR depend on a complete aggregator build succeeding.

trancexpress:
I'd lke to run the ant file directly as well, but I have no clue how to set the environment up - as commented earlier. I was hoping the aggregator build will do that.

carstenartur:
> I'd lke to run the ant file directly as well, but I have no clue how to set the environment up - as commented earlier. I was hoping the aggregator build will do that.

I think there is a reasonably small way to test the new `test.xml` part without first getting a complete aggregator build working.
The important detail is that `test.xml` is not normally run standalone from the source checkout. It expects an Eclipse test installation containing `org.eclipse.test/library.xml` and the installed test bundles, so the earlier failure when calling `ant -f test.xml` directly is expected.
However, the new `generate-leak-test-vmargs` target itself does not depend on that complete test infrastructure. We can test it with a minimal fake `eclipse-home`.
After running `mvn clean verify` in `org.eclipse.jdt.ui.tests`, something along these lines should exercise the interesting path:
    rm -rf /tmp/jdt-leak-harness
    mkdir -p /tmp/jdt-leak-harness/plugins/org.eclipse.jdt.ui.tests_test
    cp -a target/classes/. \
      /tmp/jdt-leak-harness/plugins/org.eclipse.jdt.ui.tests_test/
    ant -f test.xml \
      -Declipse-home=/tmp/jdt-leak-harness \
      -Djvm=/usr/lib/jvm/java-25-openjdk/bin/java \
      generate-leak-test-vmargs
Then we can verify both the generated contents and that Java accepts the argument file:
    grep 'java.lang.classfile' \
      /tmp/jdt-leak-harness/jdt-ui-leak-test-vmargs
    /usr/lib/jvm/java-25-openjdk/bin/java \
      @/tmp/jdt-leak-harness/jdt-ui-leak-test-vmargs \
      -version
It may also be useful to run the same target once with Java 25 and once with another installed JDK. That would directly demonstrate the important property of this change: the VM arguments are generated from the **test JDK**, not from whichever JDK happened to build the bundle.
For a complete end-to-end test we would still want the real Eclipse automated-test harness, but I don't think we need to block this smaller verification on the unrelated aggregator build problem.

trancexpress:
> Then we can verify both the generated contents and that Java accepts the argument file

```
$ ant -f test.xml  -Declipse-home=/tmp/jdt-leak-harness  -Djvm=/usr/lib/jvm/java-25/bin/java  generate-leak-test-vmargs
Buildfile: .../eclipse.jdt.ui/org.eclipse.jdt.ui.tests/test.xml

generate-leak-test-vmargs:
     [echo] Generating leak-test VM arguments with /usr/lib/jvm/java-25/bin/java

BUILD SUCCESSFUL
Total time: 0 seconds
$ /usr/lib/jvm/java-25/bin/java @/tmp/jdt-leak-harness/jdt-ui-leak-test-vmargs -version
WARNING: Using incubator modules: jdk.incubator.vector
openjdk version "25" 2025-09-16 LTS
OpenJDK Runtime Environment Temurin-25+36 (build 25+36-LTS)
OpenJDK 64-Bit Server VM Temurin-25+36 (build 25+36-LTS, mixed mode, sharing)
$ grep 'java.lang.classfile' /tmp/jdt-leak-harness/jdt-ui-leak-test-vmargs
--add-opens=java.base/java.lang.classfile=ALL-UNNAMED
--add-opens=java.base/java.lang.classfile.attribute=ALL-UNNAMED
--add-opens=java.base/java.lang.classfile.constantpool=ALL-UNNAMED
--add-opens=java.base/java.lang.classfile.instruction=ALL-UNNAMED
```
```
$ ant -f test.xml  -Declipse-home=/tmp/jdt-leak-harness  -Djvm=~/Downloads/jdk27/jdk-27/bin/java generate-leak-test-vmargs
Buildfile: .../eclipse.jdt.ui/org.eclipse.jdt.ui.tests/test.xml

generate-leak-test-vmargs:
     [echo] Generating leak-test VM arguments with ~/Downloads/jdk27/jdk-27/bin/java

BUILD SUCCESSFUL
Total time: 0 seconds
$ ~/Downloads/jdk27/jdk-27/bin/java @/tmp/jdt-leak-harness/jdt-ui-leak-test-vmargs -version
WARNING: Using incubator modules: jdk.incubator.vector
openjdk version "27-ea" 2026-09-15
OpenJDK Runtime Environment (build 27-ea+32-2315)
OpenJDK 64-Bit Server VM (build 27-ea+32-2315, mixed mode, sharing)
$ grep 'java.lang.classfile' /tmp/jdt-leak-harness/jdt-ui-leak-test-vmargs
--add-opens=java.base/java.lang.classfile=ALL-UNNAMED
--add-opens=java.base/java.lang.classfile.attribute=ALL-UNNAMED
--add-opens=java.base/java.lang.classfile.constantpool=ALL-UNNAMED
--add-opens=java.base/java.lang.classfile.instruction=ALL-UNNAMED
```

Not much else I can validate here, I still have problems with the aggregator build. I guess we merge this, trigger an I-build and revert if something is going wrong. But lets wait for test results on the PR.

trancexpress:
@carstenartur one more question before we merge, can you provide the info for the AI agent you are using, so I can put it in the commit in this style? https://github.com/eclipse-jdt/.github/blob/main/CONTRIBUTING.md#commit-message-recommendations

```
Assisted-by: GitHub Copilot (GPT-4o-2024-08-06)
```

I.e. the AI "frontend" and its model?

carstenartur:
> @carstenartur one more question before we merge, can you provide the info for the AI agent you are using, so I can put it in the commit in this style? https://github.com/eclipse-jdt/.github/blob/main/CONTRIBUTING.md#commit-message-recommendations
> 
> 
> 
> ```
> 
> Assisted-by: GitHub Copilot (GPT-4o-2024-08-06)
> 
> ```
> 
> 
> 
> I.e. the AI "frontend" and its model?  

Not exactly. I use a combination of GitHub Copilot in different variants/modes and ChatGPT with different models, depending on the task.

In some workflows, ChatGPT also asks GitHub Copilot for an additional review when it considers that useful, so there isn’t really a single AI agent behind a particular change. The tools can effectively review or complement each other’s work as part of the process.

So attributing the contribution specifically to Copilot would not fully reflect how it was produced. I review the resulting changes and remain responsible for what I submit.

Regarding Co-authored-by, I’m happy to follow whatever convention JDT/Platform prefers for AI-assisted contributions.

trancexpress:
> So attributing the contribution specifically to Copilot would not fully reflect how it was produced. I review the resulting changes and remain responsible for what I submit.
> 
> Regarding Co-authored-by, I’m happy to follow whatever convention JDT/Platform prefers for AI-assisted contributions.

So far we have:

```
    Co-authored-by: Carsten Hammer <carsten.hammer@t-online.de>
    Co-authored-by: Simeon Andreev <simeon.danailov.andreev@gmail.com>
    Co-authored-by: Copilot copilot@github.com
```

@iloveeclipse based on the reply from @carstenartur do we try to change this or we leave as is?

So far we have commits like this in JDT UI:

```
commit b74fb23a45b62025974a6fa154132e81b8335c2d (tag: I20260714-0956, tag: I20260713-2300)
Author: wenyt <75360946+wenytang-ms@users.noreply.github.com>
Date:   Tue Jul 14 02:34:30 2026 +0800

    Expose JUnit VM runner configuration to subclasses (#3079)
    
    * Expose JUnit VM runner configuration to subclasses
    
    Allow launch delegate subclasses to reuse the fully resolved VMRunnerConfiguration without reflection or duplicating the launch orchestration.
    
    Copilot-Session: c3cd113b-dad4-4d4d-90c2-60abfd204405
    Signed-off-by: wenytang-ms <wenyutang@microsoft.com>
    
    * Bump JUnit unit test feature version
    
    Update the containing feature after the org.eclipse.jdt.ui.unittest.junit bundle minor version increase.
    
    Copilot-Session: c3cd113b-dad4-4d4d-90c2-60abfd204405
    Signed-off-by: wenytang-ms <wenyutang@microsoft.com>
    
    ---------
    
    Signed-off-by: wenytang-ms <wenyutang@microsoft.com>

commit d378b1a68d9b431601adc3a96834b5da6090097c (tag: I20260418-0433, tag: I20260418-0314, tag: I20260417-1800, tag: I20260416-1800)
Author: Carsten Hammer <carsten.hammer@t-online.de>
Date:   Thu Apr 16 03:58:54 2026 +0200

    "Add disable/enable test functionality" from issue  #2744 (#2907)
    
    * Add disable/enable test functionality from PR #2744
    
    Co-authored-by: carstenartur <3164220+carstenartur@users.noreply.github.com>
    Co-authored-by: copilot-swe-agent[bot] <198982749+Copilot@users.noreply.github.com>
```

akurtakov:
As per https://www.eclipse.org/projects/handbook/#genai Assisted-By should be preferred for noting AI help.

trancexpress:
> As per https://www.eclipse.org/projects/handbook/#genai Assisted-By should be preferred for noting AI help.

So something like "Assisted-by: Copilot, ChatGPT" instead of "Co-authored-by: Copilot copilot@github.com"?

akurtakov:
Separate lines like might be better when you can't go for the full style with model and version, it is at least disclosing AI usage:
```
    Co-authored-by: Carsten Hammer <****>
    Co-authored-by: Simeon Andreev <****>
    Assisted-by: GitHub Copilot
    Assisted-by: OpenAI ChatGPT
```

carstenartur:
That sounds reasonable to me.

Since the JDT guidance was updated only very recently, I would prefer not to rewrite existing PR histories solely to retroactively apply the new recommendation. Rewriting commits also changes SHAs, invalidates some existing review/CI references, and may trigger unnecessary rebuilds.

For new contributions and whenever an existing branch needs to be rebased, squashed, or otherwise rewritten for an independent reason, I am happy to follow the current guidance and use Assisted-by instead of Co-authored-by.

For my workflow, where applicable, the most accurate attribution is generally:

Assisted-by: OpenAI ChatGPT
Assisted-by: GitHub Copilot

I would rather not add model/version information unless I can determine it reliably.

This also seems preferable to repeatedly rewriting already reviewed contributions if the guidance evolves further.

trancexpress:
@iloveeclipse can you trigger an I-build, so that we know if we are done here or if we need to revert?

iloveeclipse:
Triggered https://ci.eclipse.org/releng/job/builds/job/I-build-4.42/19/.

Just in case you would need a rebuild later and I'm not there, you can do it from https://ci.eclipse.org/releng/job/builds/job/I-build-4.42/

iloveeclipse:
FYI, https://download.eclipse.org/eclipse/downloads/drops4/I20260911-1438/ is there, test results will appear in ~3 hours

trancexpress:
The nightly I-build is done, there are no fails in JDT UI: https://download.eclipse.org/eclipse/downloads/drops4/I20260911-2300/reports.html#tests

So lets assume the changes are good.

Thank you for the help @carstenartur !

PR #3169, opened 2026-09-10
Fix move instance method to not move when target type used as lambda

### LINKED ISSUE #3124: Move Method loses lambda target type and produces uncompilable code
## Steps to reproduce

1. Create the following Java file:

```java
public class MIM_E06 {

        Source6 source = new Source6();

        int result = source.calculate(() -> 7);
        System.out.println(result);
    }
}

class Source6 {

    // Move Instance Method target
    // Move to parameter: target
    int calculate(Target6 target) {
        return target.value() + 1;
    }
}

@FunctionalInterface
interface Target6 {

    int value();
}
```

2. Run the program before refactoring.

The output is:

```text
8
```

3. Apply **Move Method** to:

```java
int calculate(Target6 target)
```

4. Select the parameter `target` as the target object and perform the refactoring.

## Code before refactoring

```java
public class MIM_E06 {

        Source6 source = new Source6();

        int result = source.calculate(() -> 7);
        System.out.println(result);
    }
}

class Source6 {

    // Move Instance Method target
    // Move to parameter: target
    int calculate(Target6 target) {
        return target.value() + 1;
    }
}

@FunctionalInterface
interface Target6 {

    int value();
}
```

### Output before refactoring

```text
8
```

## Code after refactoring

Eclipse performs the refactoring and generates:

```java
public class MIM_E06 {

        Source6 source = new Source6();

        int result = () -> 7.calculate();
        System.out.println(result);
    }
}

class Source6 {
}

@FunctionalInterface
interface Target6 {

    int value();

    default int calculate() {
        return value() + 1;
    }
}
```

### Output after refactoring

The refactored program does not compile.

Before refactoring, the lambda expression:

```java
() -> 7
```

appears as an argument to:

```java
calculate(Target6 target)
```

and therefore obtains `Target6` as its functional-interface target type from the method parameter.

After Move Method, Eclipse transforms:

```java
source.calculate(() -> 7)
```

into:

```java
() -> 7.calculate()
```

The lambda expression no longer has the `Target6` target type required for it to be used as the receiver of the moved method.

A semantics-preserving transformation would need to retain the target type, for example:

```java
((Target6) () -> 7).calculate()
```

or otherwise reject the refactoring if the transformation cannot be performed safely.

As generated, the refactored Java code is invalid and does not compile.

jjohnstn:
- modify MoveInstanceMethodProcessor.createInlineMethodInvocation() to recognize when the argument to modify is a lambda expression and in such a case create an error status with message
- add new test to MoveInstanceMethodTests
- fixes #3124

## What it does

See issue.

## How to test

See issue or new test.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/refactoring/RefactoringCoreMessages.java
+++ b/org.eclipse.jdt.core.manipulation/core extension/org/eclipse/jdt/internal/corext/refactoring/RefactoringCoreMessages.java
@@ -1379,6 +1379,8 @@ public final class RefactoringCoreMessages extends NLS {
 
 	public static String MoveInstanceMethodProcessor_no_native_methods;
 
+	public static String MoveInstanceMethodProcessor_no_lambdas;
+
 	public static String MoveInstanceMethodProcessor_no_null_argument;
 
 	public static String MoveInstanceMethodProcessor_no_resolved_target;
@@ -2550,8 +2552,6 @@ public final class RefactoringCoreMessages extends NLS {
 	public static String ConvertToRecordRefactoring_has_initializer;
 
 	public static String ConvertToRecordRefactoring_member_types_not_supported;
-
-
 
 	static {
 		NLS.initializeMessages(BUNDLE_NAME, RefactoringCoreMessages.class);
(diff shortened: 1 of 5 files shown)


jjohnstn:
Test failures are known spurious failures occurring recently.

PR #3177, opened 2026-09-11
[JUnit] Extend JUnitCore API to export any kind of ITestElement

HannesWell:
## What it does
Extend JUnitCore API to export any kind of ITestElement.

This allows to export for example a test-suite as root element of a result JUnit XML file. That aligns better with the de-facto standard for JUnit result files, e.g. established by the Maven Surefire plugin: https://maven.apache.org/surefire/maven-surefire-plugin/xsd/surefire-test-report.xsd

That schema is widely supported by external tools, while a top level 'testrun' can cause problems.

Additionally this adds the number of ignored tests additionally under the attribute name `skipped`, which is also used in the mentioned schema.

## How to test

An adjusted test case is pending, but I can look into it if there is consensus about the proposed new API.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.junit.core/src/org/eclipse/jdt/junit/JUnitCore.java
+++ b/org.eclipse.jdt.junit.core/src/org/eclipse/jdt/junit/JUnitCore.java
@@ -17,13 +17,16 @@ package org.eclipse.jdt.junit;
 
 
 import java.io.File;
+import java.io.IOException;
 import java.io.OutputStream;
 import java.lang.reflect.InvocationTargetException;
+import java.nio.file.Files;
 import java.util.HashSet;
 import java.util.Set;
 
 import javax.xml.transform.TransformerException;
 
+import org.eclipse.jdt.junit.model.ITestElement;
 import org.eclipse.jdt.junit.model.ITestRunSession;
 
 import org.eclipse.core.runtime.CoreException;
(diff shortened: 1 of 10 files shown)


org.eclipse.jdt.junit.core/src/org/eclipse/jdt/junit/JUnitCore.java
```diff
@@ -183,7 +187,7 @@ public static IType[] findTestTypes(IJavaElement container, IProgressMonitor mon
 	 * @since 3.7
 	 */
 	public static void exportTestRunSession(ITestRunSession testRunSession, File file) throws CoreException {
```
HannesWell, line 189: If you think it's suitable, this method plus the `OutputStream` overload could be deprecated (not for removal) in favor of the new, more general methods.

carstenartur:
Thanks for working on this.

Would you mind taking a look at the currently open JUnit-related PRs first, in particular the ones that touch the same model, serialization, and UI code? This PR now overlaps with some of them and will likely create additional merge conflicts if it lands first.

I realize that my own open PRs also create conflicts for subsequent work. The difference is mainly that I cannot merge them myself, so I have no way to reduce that queue other than keeping them rebased and responding to review feedback.

If some of the existing PRs look good to you, it would therefore be very helpful if you could review and, where appropriate, merge them before this change advances further. That would also give this PR a more stable base and reduce the amount of duplicated conflict resolution.

Of course, if you see architectural or correctness issues in any of them, review feedback would be equally helpful.

PR #3196, opened 2026-09-16
Check Java nature instead of null handle when importing projects

vogella:
- fix `getFoldersToIgnore()` in `JDTProjectNatureImportConfigurator` and `ProjectWithJavaResourcesImportConfigurator` to skip projects without the Java nature
- removes the spurious ERROR entries these configurators write to the Error Log during project import

## What it does

`JavaCore.create(IProject)` returns a handle without checking existence or the Java nature and is documented to return `null` only for a `null` argument. The `javaProject == null` guards in both import configurators were therefore dead code, and the following `getOutputLocation()` threw a `JavaModelException` (code 969) for every imported project that does not have the Java nature yet, which was logged as ERROR although the importer recovers. Both classes now test `project.hasNature(JavaCore.NATURE_ID)` instead, which is what the existing comment already described. PDE's `BundleProjectConfigurator.getFoldersToIgnore()` delegates to the JDT one, so this also silences the same stack trace reported from PDE.

## How to test

Import a directory tree containing projects that carry a `.classpath` and `.project` but no Java nature (File > Open Projects from File System). Before the change the Error Log fills with `Java Model Exception: Java Model Status [<project> does not exist]`; afterwards the import is silent and the detected folders are unchanged for real Java projects.

## Author checklist


jjohnstn (approved):
Approved.

vogella:
Awesome @jjohnstn, thanks a lot for the review

PR #3208, opened 2026-09-17
Issue 3207 core ast provider hangs

### LINKED ISSUE #3207: `CoreASTProvider` may hang indefinitely
## Summary

An improvement of the work that started in https://bugs.eclipse.org/bugs/show_bug.cgi?id=366048#c21 but without the problems described in https://bugs.eclipse.org/bugs/show_bug.cgi?id=366048#c31 / https://bugs.eclipse.org/bugs/show_bug.cgi?id=366048#c34

**Expands on:**
- https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1977 

I am fairly sure that one is this exact same problem but I didn't declare it as a _duplicate_ yet since the other issue doesn't mention which other threads were involved. But by looking at the mentioned stack-trace, I see that the involved Thread has the same pattern we all know, namely: the call to the recursion X times and the call to `fWaitLock.wait(30000)` at the top.

---

`CoreASTProvider.getAST(element, WAIT_YES | WAIT_ACTIVE_ONLY, monitor)` can hang indefinitely — observed in production for 76 minutes — when three unrelated call sites all request the AST of the same active editor's element while a reconcile for it is in progress, and that reconcile never completes.

## The specific scenario

Three independent call sites, all targeting the **same active element**, two different wait flags:

| Call site | Thread | Flag |
|---|---|---|
| `OrganizeImportsAction.run()` | **main UI thread** (synchronous) | `WAIT_ACTIVE_ONLY` |
| `OverrideIndicatorLabelDecorator` ("Decoration Calculation" job) | background worker | `WAIT_YES` |
| `SelectionListenerWithASTManager` ("Requesting Java AST from selection" job) | background worker | `WAIT_YES` |

At the time all three called `getAST`, the element was marked as being reconciled (`isReconciling() == true`), but the background reconciler responsible for it had already gone idle — it was not going to call `reconciled()` for this element again. A thread dump of the incident shows all three callers still blocked in `Object.wait()` inside `CoreASTProvider.getAST`, at increasing recursion depth, while every reconciler thread in the process sits idle. The freeze was recorded as still ongoing 4,581 seconds (~76 minutes) after it started — roughly 152 stacked 30-second retries.

```mermaid
gantt
    title CoreASTProvider deadlock — a real 76-minute UI freeze
    dateFormat HH:mm
    axisFormat %H:%M
    section Main UI thread (SWT event loop)
    OrganizeImportsAction -> getAST(WAIT_ACTIVE_ONLY), blocked :crit, main, 11:49, 76m
    section Worker: Decoration Calculation
    OverrideIndicatorLabelDecorator -> getAST(WAIT_YES), blocked :crit, dec, 11:49, 76m
    section Worker: Requesting Java AST from selection
    SelectionListenerWithASTManager -> getAST(WAIT_YES), blocked :crit, sel, 11:49, 76m
    section Java reconciler thread
    Idle in AbstractReconciler.delay() — not reconciling this element anymore :done, idle, 11:49, 76m
```

Note: `WAIT_ACTIVE_ONLY` is not the conservative "never wait" flag its name suggests — for an element that *is* the active one but has no AST cached yet, `CoreASTProvider.getAST`'s `canReturnNull` computation makes it behave identically to `WAIT_YES`. That's why `OrganizeImportsAction`, running on the main UI thread, ends up blocked too.

## Root cause

`getAST(WAIT_YES)` retries by recursing when the requested (active) element is still marked as reconciling. Each retry gets a **fresh** 30-second `fWaitLock.wait(...)`, with no overall deadline across retries. If nothing ever calls the matching `reconciled()` for that element again, `isReconciling` never clears, and the retry loop has no way to terminate — this reproduces the pattern already described in this comment in BugZilla eclipse.org bugzilla 366048 (comments 25–34): the 2013 fix that added the 30-second timeout turned an indefinite *wait* into an indefinite *retry loop* instead.

## Involved threads 

Here's an extract that shows the threads involved in the lock. You will notice that the blocked threads are in state `TIMED_WAITING` and they wait on `java.lang.Object@46d53006`, which will never be unlocked because the only thread that could actually call the `notifyAll()` on it is the `JavaReconciler`-thread, which is already idle in `AbstractReconciler.delay(...)`.

<details><summary>Thread dump (extract)</summary>
<p>

```
"main" prio=6 Id=3 TIMED_WAITING on java.lang.Object@46d53006
	at java.base@25.0.3/java.lang.Object.wait0(Native Method)
	-  waiting on java.lang.Object@46d53006
	at java.base@25.0.3/java.lang.Object.wait(Object.java:389)
	at org.eclipse.jdt.core.manipulation.CoreASTProvider.getAST(CoreASTProvider.java:176)
	at org.eclipse.jdt.core.manipulation.CoreASTProvider.getAST(CoreASTProvider.java:189)
	at org.eclipse.jdt.core.manipulation.CoreASTProvider.getAST(CoreASTProvider.java:189)
	at org.eclipse.jdt.core.manipulation.CoreASTProvider.getAST(CoreASTProvider.java:189)
	at org.eclipse.jdt.core.manipulation.CoreASTProvider.getAST(CoreASTProvider.java:189)
	at org.eclipse.jdt.core.manipulation.CoreASTProvider.getAST(CoreASTProvider.java:189)
	...

"Worker-344: Decoration Calculation" prio=5 Id=1077 TIMED_WAITING on java.lang.Object@46d53006
	at java.base@25.0.3/java.lang.Object.wait0(Native Method)
	-  waiting on java.lang.Object@46d53006
	at java.base@25.0.3/java.lang.Object.wait(Object.java:389)
	at org.eclipse.jdt.core.manipulation.CoreASTProvider.getAST(CoreASTProvider.java:176)
	at org.eclipse.jdt.core.manipulation.CoreASTProvider.getAST(CoreASTProvider.java:189)
	at org.eclipse.jdt.core.manipulation.CoreASTProvider.getAST(CoreASTProvider.java:189)
	at org.eclipse.jdt.core.manipulation.CoreASTProvider.getAST(CoreASTProvider.java:189)
	at org.eclipse.jdt.core.manipulation.CoreASTProvider.getAST(CoreASTProvider.java:189)
	at org.eclipse.jdt.core.manipulation.CoreASTProvider.getAST(CoreASTProvider.java:189)
	...

"Worker-348: Requesting Java AST from selection" prio=5 Id=1161 TIMED_WAITING on java.lang.Object@46d53006
	at java.base@25.0.3/java.lang.Object.wait0(Native Method)
	-  waiting on java.lang.Object@46d53006
	at java.base@25.0.3/java.lang.Object.wait(Object.java:389)
	at org.eclipse.jdt.core.manipulation.CoreASTProvider.getAST(CoreASTProvider.java:176)
	at org.eclipse.jdt.core.manipulation.CoreASTProvider.getAST(CoreASTProvider.java:189)
	at org.eclipse.jdt.core.manipulation.CoreASTProvider.getAST(CoreASTProvider.java:189)
	at org.eclipse.jdt.core.manipulation.CoreASTProvider.getAST(CoreASTProvider.java:189)
	at org.eclipse.jdt.core.manipulation.CoreASTProvider.getAST(CoreASTProvider.java:189)
	at org.eclipse.jdt.core.manipulation.CoreASTProvider.getAST(CoreASTProvider.java:189)
	...

"org.eclipse.jdt.internal.ui.text.JavaReconciler" daemon prio=1 Id=780 TIMED_WAITING on org.eclipse.jface.text.reconciler.DirtyRegionQueue@5d9fbda0
	at java.base@25.0.3/java.lang.Object.wait0(Native Method)
	-  waiting on org.eclipse.jface.text.reconciler.DirtyRegionQueue@5d9fbda0
	at java.base@25.0.3/java.lang.Object.wait(Object.java:389)
	at org.eclipse.jface.text.reconciler.AbstractReconciler.delay(AbstractReconciler.java:615)
	at org.eclipse.jface.text.reconciler.AbstractReconciler$BackgroundWorker.run(AbstractReconciler.java:176)
	at java.base@25.0.3/java.lang.Thread.runWith(Thread.java:1487)
	at java.base@25.0.3/java.lang.Thread.run(Thread.java:1474)

```

</p>
</details>

fedejeanne:
## What it does
Fixes #3207 

This PR contains 2 commits:

- The 1st commit is the "safety net": it makes sure that `CoreASTProvider` respects the overall 30s timeout in case some bad callers forget to balance the calls to `fWaitLock.wait(...)` and `fWaitLock.notifyAll()`. It also provides a test that simulates the "bad caller" and asserts that the timeout kicked in and the callers received a `null` AST back.

- The 2nd commit fixes some of the "bad callers" so that the safety net (the timeout) isn't even necessary.

Both commits fix the issue #3207 in their own way, but I'd like to have them both; especially the "safety net", just in case there are still some other "bad callers" out there and we haven't found them yet.

## How to test
Run the test.

## Author checklist


org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/text/JavaCompositeReconcilingStrategy.java
```diff
@@ -162,6 +169,19 @@ public void aboutToWork(JavaReconciler javaReconciler) {
...
+	 * Tells {@link CoreASTProvider} that a reconcile for this editor's input is starting right
+	 * now, immediately before the try block that unconditionally calls {@link #reconciled()} in
+	 * its finally clause - so the two calls can never become unbalanced the way the old
+	 * aboutToWork()-based registration could (it fired speculatively, before the reconciler
+	 * necessarily reached this method at all; see bugzilla 366048 comment 29).
+	 */
+	private void aboutToBeReconciledForASTProvider() {
+		ITypeRoot inputJavaElement= EditorUtility.getEditorInputJavaElement(fEditor, false);
+		if (inputJavaElement != null)
+			CoreASTProvider.getInstance().aboutToBeReconciled(inputJavaElement);
```
fedejeanne: 🟡 The old `CompilationUnitEditor.aboutToWork()` passed `javaReconciler::signalWaitForFinish` as the `finishReconciling` callback to `aboutToBeReconciled(ITypeRoot, Runnable)`. That callback is invoked by `CoreASTProvider.notifyReconciler()` while a caller is blocked in `getAST()`, to nudge the reconciler into actually doing the work instead of sitting idle until its own timer fires.

Here `aboutToBeReconciled(inputJavaElement)` uses the 1-arg overload, which passes `null` for that callback (`fFinishReconciling` stays `null`). That silently drops the "wake the reconciler up" nudge — waiters in `getAST()` will now just sit out the wait instead of prodding the reconciler to hurry up.

Is that intentional? If the goal is only to fix the aboutToWork()/reconciled() balance issue, could the `Runnable` still be threaded through here (e.g. via `JavaReconciler`) so we don't lose that liveness optimization along with the correctness fix?

*(Created by copilot)*

org.eclipse.jdt.ui.tests/ui/org/eclipse/jdt/ui/tests/core/CoreASTProviderConcurrencyTest.java
```diff
@@ -0,0 +1,167 @@
...
+ * concurrently and asserts none of them can be stuck forever.
+ */
+public class CoreASTProviderConcurrencyTest extends CoreTests {
+
+	@Rule
+	public ProjectTestSetup pts= new ProjectTestSetup();
+
+	/**
+	 * Upper bound for the whole test. Must be comfortably longer than
+	 * {@code CoreASTProvider.OVERALL_RECONCILE_TIMEOUT_SECONDS}
```
fedejeanne: 🟢 This javadoc refers to `CoreASTProvider.OVERALL_RECONCILE_TIMEOUT_SECONDS`, but the production code doesn't have a constant by that name — it has a private `getOverallReconcileTimeoutSeconds()` method reading the `"CoreASTProvider.reconcileTimeout"` system property. Please fix the reference so it doesn't look like a real, resolvable symbol.

*(Created by copilot)*

org.eclipse.jdt.ui.tests/ui/org/eclipse/jdt/ui/tests/core/CoreASTProviderConcurrencyTest.java
```diff
@@ -0,0 +1,167 @@
...
+
+	private IJavaProject fJProject1;
+
+	private IPackageFragmentRoot fSourceFolder;
+
+	private CoreASTProvider fProvider;
+
+	@Before
+	public void setUp() throws Exception {
+		System.setProperty("CoreASTProvider.reconcileTimeout", "1");
```
fedejeanne: 🟢 The system property key `"CoreASTProvider.reconcileTimeout"` is duplicated as a string literal here and in `CoreASTProvider#getOverallReconcileTimeoutSeconds()`. Consider exposing it as a small package/public constant on `CoreASTProvider` so the two can't drift apart under a future rename.

*(Created by copilot)*

fedejeanne (commented):
I reviewed with my own agent.

org.eclipse.jdt.ui.tests/ui/org/eclipse/jdt/ui/tests/core/CoreASTProviderConcurrencyTest.java
fedejeanne: Done, I removed the reference.

org.eclipse.jdt.ui.tests/ui/org/eclipse/jdt/ui/tests/core/CoreASTProviderConcurrencyTest.java
fedejeanne: Fixed, I introduced a public constant in `CoreASTProvider`

org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/text/JavaCompositeReconcilingStrategy.java
fedejeanne: Good point, I added the same nudge into `aboutToBeReconciledForASTProvider`:

```java
private void aboutToBeReconciledForASTProvider() {
    ITypeRoot inputJavaElement= EditorUtility.getEditorInputJavaElement(fEditor, false);
    if (inputJavaElement != null)
        CoreASTProvider.getInstance().aboutToBeReconciled(inputJavaElement,
                fJavaReconciler != null ? fJavaReconciler::signalWaitForFinish : null);
}
```

fedejeanne (commented):
Made the adjustments

org.eclipse.jdt.core.manipulation/common/org/eclipse/jdt/core/manipulation/CoreASTProvider.java
```diff
@@ -52,6 +53,34 @@
...
+	 * (currently 30_000 ms), e.g. by setting it to:
+	 *
+	 * <pre>
+	 * 	-DCoreASTProvider.reconcileTimeoutMilliseconds=15_000
+	 * </pre>
+	 *
+	 * @since 1.26
+	 */
+	public static final String CORE_AST_PROVIDER_RECONCILE_TIMEOUT_MILLSECONDS_PROPERTY= "CoreASTProvider.reconcileTimeoutMilliseconds"; //$NON-NLS-1$
+
```
fedejeanne: 🟡 Typo in this new **public** constant name: `CORE_AST_PROVIDER_RECONCILE_TIMEOUT_MILLSECONDS_PROPERTY` is missing the `I` in "MILLISECONDS" (same typo repeats in `getOverallReconcileTimeoutMillseconds()` below, and in the javadoc above it — `"CoreASTProvider.reconcileTimeoutMillseconds"` — which doesn't even match the correctly-spelled property value string `"CoreASTProvider.reconcileTimeoutMilliseconds"` used here).

Worth fixing now since this is a new public API surface (`@since 1.26`) — once released, the misspelled name is stuck for compatibility.

*(Created by copilot)*

org.eclipse.jdt.ui.tests/ui/org/eclipse/jdt/ui/tests/core/CoreASTProviderConcurrencyTest.java
```diff
@@ -0,0 +1,172 @@
...
+			CompletableFuture<CompilationUnit> organizeImports= CompletableFuture.supplyAsync(
+					() -> fProvider.getAST(cu, CoreASTProvider.WAIT_ACTIVE_ONLY, null), pool);
+			CompletableFuture<CompilationUnit> decorationCalculation= CompletableFuture.supplyAsync(
+					() -> fProvider.getAST(cu, CoreASTProvider.WAIT_YES, null), pool);
+			CompletableFuture<CompilationUnit> selectionAst= CompletableFuture.supplyAsync(
+					() -> fProvider.getAST(cu, CoreASTProvider.WAIT_YES, null), pool);
+
+			CompletableFuture<Void> all= CompletableFuture.allOf(organizeImports, decorationCalculation, selectionAst);
+			try {
+				all.get(OVERALL_RECONCILE_TIMEOUT_MILLISECONDS, TimeUnit.SECONDS);
```
fedejeanne: 🔴 `OVERALL_RECONCILE_TIMEOUT_MILLISECONDS` (200) is passed here with `TimeUnit.SECONDS`, not `TimeUnit.MILLISECONDS`. As written, `all.get(...)` waits up to 200 **seconds**, not 200 ms, before failing — the test still passes today because the real `getAST()` calls return well within that window, but it defeats the intent of a tight, fast-failing bound, and means a real regression back to the old "waits forever" behavior would hang this test for 200s instead of failing fast. The failure message text ("stuck after 200s", line 150) also becomes misleading once the unit is fixed. Should be `TimeUnit.MILLISECONDS`.

*(Created by copilot)*

org.eclipse.jdt.core.manipulation/common/org/eclipse/jdt/core/manipulation/CoreASTProvider.java
```diff
@@ -52,6 +53,34 @@
  */
 public final class CoreASTProvider {
 
+	/**
+	 * The default timeout for the reconcile operation. It can be overwritten via system property.
+	 *
+	 * @see #CORE_AST_PROVIDER_RECONCILE_TIMEOUT_MILLISECONDS_PROPERTY
+	 */
+	private static final int CORE_AST_PROVIDER_DEFAULT_RECONCILE_TIMEOUT_MILLSECONDS= 30_000;
```
fedejeanne: 🟢 The public constant `CORE_AST_PROVIDER_RECONCILE_TIMEOUT_MILLISECONDS_PROPERTY` got the spelling fix, but this private sibling (`CORE_AST_PROVIDER_DEFAULT_RECONCILE_TIMEOUT_MILLSECONDS`) and `getOverallReconcileTimeoutMillseconds()` at line 80 still say "MILLSECONDS"/"Millseconds". Not public API so not blocking, but worth making consistent with the now-correctly-spelled constant right next to them.

*(Created by copilot)*

org.eclipse.jdt.ui.tests/ui/org/eclipse/jdt/ui/tests/core/CoreASTProviderConcurrencyTest.java
```diff
@@ -0,0 +1,172 @@
...
+					() -> fProvider.getAST(cu, CoreASTProvider.WAIT_YES, null), pool);
+
+			CompletableFuture<Void> all= CompletableFuture.allOf(organizeImports, decorationCalculation, selectionAst);
+			try {
+				all.get(OVERALL_RECONCILE_TIMEOUT_MILLISECONDS, TimeUnit.MILLISECONDS);
+			} catch (TimeoutException e) {
+				fail((!organizeImports.isDone() ? 1 : 0) + (!decorationCalculation.isDone() ? 1 : 0) + (!selectionAst.isDone() ? 1 : 0)
+						+ " of 3 concurrent CoreASTProvider.getAST callers (OrganizeImportsAction/WAIT_ACTIVE_ONLY, "
+						+ "Decoration Calculation/WAIT_YES, Requesting Java AST from selection/WAIT_YES) were still "
+						+ "stuck after " + OVERALL_RECONCILE_TIMEOUT_MILLISECONDS + "s, waiting for a reconcile that "
```
fedejeanne: 🟢 Now that `all.get(...)` above correctly uses `TimeUnit.MILLISECONDS`, this failure message's `+ "s, waiting for a reconcile..."` suffix is misleading — it prints e.g. "stuck after 200s" when the value is actually 200 ms. Minor, but worth `+ "ms, waiting..."` for an accurate failure message.

*(Created by copilot)*

fedejeanne:
Drafted until I know that there are no more failing tests.

fedejeanne:
The failing check is odd:

```
[2026-09-21T13:51:52.330Z] [INFO] Reactor Summary:
[2026-09-21T13:51:52.330Z] [INFO] 
[2026-09-21T13:51:52.330Z] [INFO] eclipse.jdt.ui 4.42.0-SNAPSHOT ..................... SUCCESS [ 11.183 s]
[2026-09-21T13:51:52.330Z] [INFO] [bundle] Java Code Manipulation Functionality 1.26.0-SNAPSHOT SUCCESS [02:39 min]
[2026-09-21T13:51:52.330Z] [INFO] org.eclipse.jdt.ui 3.39.100-SNAPSHOT ............... SUCCESS [02:45 min]
[2026-09-21T13:51:52.330Z] [INFO] org.eclipse.jdt.bcoview 1.3.0-SNAPSHOT ............. SUCCESS [ 14.391 s]
[2026-09-21T13:51:52.330Z] [INFO] org.eclipse.jdt.junit.runtime 3.8.200-SNAPSHOT ..... SUCCESS [ 19.362 s]
[2026-09-21T13:51:52.330Z] [INFO] [bundle] AST View Plug-in 1.7.200-SNAPSHOT ......... SUCCESS [ 10.364 s]
[2026-09-21T13:51:52.330Z] [INFO] [bundle] Java Element View Plug-in 1.6.200-SNAPSHOT  SUCCESS [  8.025 s]
[2026-09-21T13:51:52.330Z] [INFO] org.eclipse.jdt.junit5.runtime 1.2.100-SNAPSHOT .... FAILURE [  4.128 s]
[2026-09-21T13:51:52.330Z] [INFO] [bundle] Java Development Tools JUnit core support 3.15.100-SNAPSHOT SUCCESS [ 20.522 s]
[2026-09-21T13:51:52.330Z] [INFO] [bundle] Java Development Tools JUnit support 3.17.700-SNAPSHOT SUCCESS [ 26.230 s]
[2026-09-21T13:51:52.330Z] [INFO] org.eclipse.jdt.junit4.runtime 1.4.100-SNAPSHOT .... SUCCESS [ 13.716 s]
[2026-09-21T13:51:52.330Z] [INFO] org.eclipse.jdt.junit6.runtime 1.0.200-SNAPSHOT .... FAILURE [  4.227 s]
[2026-09-21T13:51:52.330Z] [INFO] org.eclipse.jdt.ui.tests 3.16.500-SNAPSHOT ......... SKIPPED
[2026-09-21T13:51:52.330Z] [INFO] org.eclipse.jdt.text.tests 3.15.500-SNAPSHOT ....... SKIPPED
[2026-09-21T13:51:52.330Z] [INFO] org.eclipse.jdt.ui.tests.refactoring 3.16.500-SNAPSHOT SKIPPED
[2026-09-21T13:51:52.330Z] [INFO] org.eclipse.jdt.ui.junit.sampleproject 1.0.0-SNAPSHOT SUCCESS [ 22.299 s]
[2026-09-21T13:51:52.330Z] [INFO] org.eclipse.jdt.ui.examples.projects 3.3.0-SNAPSHOT  SUCCESS [  9.763 s]
[2026-09-21T13:51:52.330Z] [INFO] org.eclipse.jdt.ui.unittest.junit 1.4.0-SNAPSHOT ... SUCCESS [ 22.030 s]
[2026-09-21T13:51:52.330Z] [INFO] [feature] JUnit Test Runner Client for UnitTest View 1.2.300-SNAPSHOT SUCCESS [  3.248 s]
[2026-09-21T13:51:52.330Z] [INFO] [feature] Java AST View 1.2.700-SNAPSHOT ........... SUCCESS [  0.696 s]
[2026-09-21T13:51:52.330Z] [INFO] [feature] Java Element View 1.1.900-SNAPSHOT ....... SUCCESS [  0.611 s]
[2026-09-21T13:51:52.330Z] [INFO] [feature] Java Bytecode Outline View 1.2.800-SNAPSHOT SUCCESS [  0.743 s]
[2026-09-21T13:51:52.330Z] [INFO] ------------------------------------------------------------------------
[2026-09-21T13:51:52.330Z] [INFO] BUILD FAILURE
[2026-09-21T13:51:52.330Z] [INFO] ------------------------------------------------------------------------
[2026-09-21T13:51:52.330Z] [INFO] Total time:  09:06 min
[2026-09-21T13:51:52.330Z] [INFO] Finished at: 2026-09-21T13:51:52Z
[2026-09-21T13:51:52.330Z] [INFO] ------------------------------------------------------------------------
[2026-09-21T13:51:52.889Z] [ERROR] Failed to execute goal org.eclipse.tycho.extras:tycho-p2-extras-plugin:5.0.4:compare-version-with-baselines (compare-attached-artifacts-with-release) on project org.eclipse.jdt.junit5.runtime: Baseline and reactor have the same fully qualified version, but different content
[2026-09-21T13:51:52.889Z] [ERROR] different
[2026-09-21T13:51:52.889Z] [ERROR]    META-INF/ECLIPSE_.RSA: present in baseline only
[2026-09-21T13:51:52.889Z] [ERROR]    META-INF/ECLIPSE_.SF: present in baseline only
[2026-09-21T13:51:52.889Z] [ERROR] -> [Help 1]
[2026-09-21T13:51:52.889Z] org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.eclipse.tycho.extras:tycho-p2-extras-plugin:5.0.4:compare-version-with-baselines (compare-attached-artifacts-with-release) on project org.eclipse.jdt.junit5.runtime: Baseline and reactor have the same fully qualified version, but different content
[2026-09-21T13:51:52.889Z] different
[2026-09-21T13:51:52.889Z]    META-INF/ECLIPSE_.RSA: present in baseline only
[2026-09-21T13:51:52.889Z]    META-INF/ECLIPSE_.SF: present in baseline only
```

I rebased on `master` to force another build, let's see if it goes away and all tests are executed.

org.eclipse.jdt.ui.tests/ui/org/eclipse/jdt/ui/tests/core/CoreASTProviderConcurrencyTest.java
```diff
@@ -0,0 +1,172 @@
...
+		@SuppressWarnings("resource") // The pool is going to be "closed" in the finally-block by calling pool.shutdownNow()
+		ExecutorService pool= Executors.newFixedThreadPool(3);
+		try {
+			// The three real callers from the production freeze, same element, same flags.
+			CompletableFuture<CompilationUnit> organizeImports= CompletableFuture.supplyAsync(
+					() -> fProvider.getAST(cu, CoreASTProvider.WAIT_ACTIVE_ONLY, null), pool);
+			CompletableFuture<CompilationUnit> decorationCalculation= CompletableFuture.supplyAsync(
+					() -> fProvider.getAST(cu, CoreASTProvider.WAIT_YES, null), pool);
+			CompletableFuture<CompilationUnit> selectionAst= CompletableFuture.supplyAsync(
+					() -> fProvider.getAST(cu, CoreASTProvider.WAIT_YES, null), pool);
```
r-mennig, line 141: The variable names, assertion texts and javadoc for/inside this test class seem overly specific to the original bug. The root cause of the problem is a more general problem, so I think this test case should also use more generic naming / descriptions

org.eclipse.jdt.core.manipulation/common/org/eclipse/jdt/core/manipulation/CoreASTProvider.java
```diff
@@ -117,15 +146,31 @@ public String toString() {
...
+	}
+
+	/**
+	 * Same as {@link #getAST(ITypeRoot, WAIT_FLAG, IProgressMonitor)}, but reusing the given
+	 * overall deadline across recursive retries instead of giving each retry's
+	 * {@code fWaitLock.wait(...)} a fresh timeout. Without this, a reconcile that never completes
+	 * (or a missed {@code notifyAll()}) makes the retries continue forever. See
+	 * https://bugs.eclipse.org/bugs/show_bug.cgi?id=366048
+	 */
+	private CompilationUnit getAST(final ITypeRoot input, WAIT_FLAG waitFlag, IProgressMonitor progressMonitor, long waitDeadlineNanos) {
```
r-mennig, line 173: Nitpicking: Using the java.time.Instant abstraction instead of the nanoseconds makes it easier to see what this parameter is used for and avoids most of the unit conversion code

org.eclipse.jdt.core.manipulation/common/org/eclipse/jdt/core/manipulation/CoreASTProvider.java
```diff
@@ -258,6 +318,41 @@ public void aboutToBeReconciled(ITypeRoot javaElement, Runnable finishReconcilin
...
+	 * @param javaElement the Java element whose cached AST is no longer up to date, may be
+	 *            <code>null</code>
+	 * @since 1.26
+	 */
+	public void invalidateCachedAST(ITypeRoot javaElement) {
+		if (javaElement == null)
+			return;
+
+		if (JavaManipulationPlugin.DEBUG_AST_PROVIDER)
+			System.out.println(getThreadName() + " - " + DEBUG_PREFIX + "invalidating cached AST for: " + toString(javaElement)); //$NON-NLS-1$ //$NON-NLS-2$
```
r-mennig, line 351: This produces duplicate "invalidating cached AST ..."/ "disposing AST" debug logs (via internal call to `disposeAST()` ) when called while `fAST != null`. 

I also think this method is a little redundant and the null-check + `cache(null, javaElement)` could be inlined into `aboutToWork(JavaReconciler javaReconciler)`

org.eclipse.jdt.ui.tests/ui/org/eclipse/jdt/ui/tests/core/CoreASTProviderConcurrencyTest.java
fedejeanne: Fair point, but I decided to go the other way because I was aiming for a more concrete reproducer: I changed the reference so that it points to the new GH issue instead of the old bugzilla issue. The GH issue has a more detailed explanation and cites the old bugzilla issue anyway.

org.eclipse.jdt.core.manipulation/common/org/eclipse/jdt/core/manipulation/CoreASTProvider.java
fedejeanne: Done

org.eclipse.jdt.core.manipulation/common/org/eclipse/jdt/core/manipulation/CoreASTProvider.java
fedejeanne: Done, I moved the logic to the caller and removed the new method.

org.eclipse.jdt.core.manipulation/common/org/eclipse/jdt/core/manipulation/CoreASTProvider.java
```diff
@@ -52,6 +54,34 @@
...
+	 * @see #CORE_AST_PROVIDER_RECONCILE_TIMEOUT_MILLISECONDS_PROPERTY
+	 */
+	private static final int CORE_AST_PROVIDER_DEFAULT_RECONCILE_TIMEOUT_MILLISECONDS= 30_000;
+
+	/**
+	 * Use this system property to override the default timeout of the reconcile operation
+	 * (currently 30_000 ms), e.g. by setting it to:
+	 *
+	 * <pre>
+	 * 	-DCoreASTProvider.reconcileTimeoutMilliseconds=15_000
```
r-mennig: This comment is misleading. `Integer.decode(String)` (used internally by `Integer.getInteger(String nm, int val)`) does not allow underscores. Using this system property verbatim produces a `NumberFormatException`

org.eclipse.jdt.core.manipulation/common/org/eclipse/jdt/core/manipulation/CoreASTProvider.java
```diff
@@ -166,14 +212,29 @@ public CompilationUnit getAST(final ITypeRoot input, WAIT_FLAG waitFlag, IProgre
...
 				notifyReconciler();
-				// Wait for AST
+				// Wait for AST, bounded by the overall deadline above, not reset on every retry
 				synchronized (fWaitLock) {
 					if (isReconciling(input)) {
 						if (JavaManipulationPlugin.DEBUG_AST_PROVIDER)
 							System.out.println(getThreadName() + " - " + DEBUG_PREFIX + "waiting for AST for: " + input.getElementName()); //$NON-NLS-1$ //$NON-NLS-2$
-						fWaitLock.wait(30000); // XXX: The 30 seconds timeout is an attempt to at least avoid a deadlock. See https://bugs.eclipse.org/366048#c21
+						long remainingMillis= reconcilingDeadline.toEpochMilli() - System.currentTimeMillis();
+						fWaitLock.wait(remainingMillis); // See https://bugs.eclipse.org/366048#c21
```
r-mennig: `remainingMillis` can become negative here if the current time is now after the `reconcilingDeadline` which would cause an `IllegalArgumentException` when calling `fWaitLock.wait(remainingMillis)`. Also, if remainingMillis reaches 0, `fWaitLock.wait(remainingMillis)` waits indefinitely. 

 
```suggestion
				if (remainingMillis > 0) fWaitLock.wait(remainingMillis); // See https://bugs.eclipse.org/366048#c21
```

org.eclipse.jdt.ui.tests/ui/org/eclipse/jdt/ui/tests/core/CoreASTProviderConcurrencyTest.java
```diff
@@ -0,0 +1,172 @@
...
+public class CoreASTProviderConcurrencyTest extends CoreTests {
+
+	@Rule
+	public ProjectTestSetup pts= new ProjectTestSetup();
+
+	/**
+	 * Upper bound for the whole test.
+	 *
+	 */
+	private static final long OVERALL_RECONCILE_TIMEOUT_MILLISECONDS= 200;
```
r-mennig: This upper bound seems to be too strict (see the current failing tests), maybe this could be increased?

fedejeanne: Indeed. I separated the timeouts so that the reconciler timeout is low (should let the test pass very fast) and the overall test timeout is high (just in case the CI machine it runs on is old and slow). This effectively changes the shape of the test to "reconciling should have been aborted after 50ms but it was still running after 10s" i.e. the test doesn't strictly check that the timeout is honored but it still tests that there is a timeout and that the wait is not infinite (the bug is therefore fixed).

org.eclipse.jdt.core.manipulation/common/org/eclipse/jdt/core/manipulation/CoreASTProvider.java
fedejeanne: Good catch, changed it to `15000`.

org.eclipse.jdt.core.manipulation/common/org/eclipse/jdt/core/manipulation/CoreASTProvider.java
fedejeanne: Done, thank you!

